PR #1, opened 2023-02-21
[WIP] Add "real" `read_parquet` logic

rjzamora:
Updates `ReadParquet` to use metadata-parsing and IO logic from `dask.dataframe.io.parquet`.

Requires https://github.com/dask/dask/pull/9637 (only because my environment was using that PR when I put this together).


dask_match/core.py
```diff
@@ -106,15 +106,15 @@ def __repr__(self):
...
     def __getattr__(self, key):
-        if key in type(self)._parameters:
-            idx = type(self)._parameters.index(key)
-            return self.operands[idx]
-        elif key in dir(type(self)):
-            return object.__getattribute__(self, key)
-        elif key in self.columns:
-            return self[key]
-        else:
-            return object.__getattribute__(self, key)
```
mrocklin, line 117: I'm curious, how did this come about?

Also, this seems unnecessarily nested.  I suspect that if this is necessary then this would be better with something like the following:

```python
if key == "__name__":
else:
    ...
```

I'll probably be pretty allergic to anything that increases nesting

dask_match/core.py
```diff
@@ -553,110 +553,10 @@ class IO(API):
...
-        if self.columns is not None:
-            df = df[self.columns]
-        return df.head(0)
-
-    @classmethod
-    def _replacement_rules(cls):
-        _ = Wildcard.dot()
-        a, b, c, d, e, f = map(Wildcard.dot, "abcdef")
+def ReadParquet(*args, **kwargs):
+    from dask_match.io.parquet import ReadParquet
```
mrocklin, line 557: If we're going with a normal function then let's call it `read_parquet` as dask.dataframe does.  I only used CamelCase previously to respect Python class conventions.

dask_match/io/parquet.py
```diff
@@ -0,0 +1,330 @@
...
+            calculate_divisions,
+            ignore_metadata_file,
+            metadata_task_size,
+            split_row_groups,
+            blocksize,
+            aggregate_files,
+            parquet_file_extension,
+            filesystem,
+            kwargs,
+        ), {}
```
mrocklin, line 121: I'm thinking of removing this method and instead relying on the `def read_parquet` and similar functions.  The classes won't do any normalization, and we'll depend on the user-facing API to handle it.  Thoughts?

dask_match/tests/test_core.py
```diff
@@ -105,7 +107,10 @@ def test_dask():
         M.min,
         M.sum,
         M.count,
-        pytest.mark.skip(M.mean, reason="scalars don't work yet"),
+        pytest.param(
+            M.mean,
+            marks=pytest.mark.skip(reason="scalars don't work yet"),
+        ),
```
mrocklin, line 113: Oh thanks!

dask_match/tests/test_core.py
```diff
@@ -114,15 +119,15 @@ def test_reductions(func):
     df["y"] = df.y * 10.0
     ddf = from_pandas(df, npartitions=10)
 
-    assert_eq(func(ddf), func(df))
-    assert_eq(func(ddf.x), func(df.x))
+    assert_eq(func(ddf).compute(), func(df))
+    assert_eq(func(ddf.x).compute(), func(df.x))
```
mrocklin, line 123: Ah, you probably needed this diff applied to dask/dask.  Sorry

```diff
diff --git a/dask/dataframe/utils.py b/dask/dataframe/utils.py
index a52a53619..43b06b41b 100644
--- a/dask/dataframe/utils.py
+++ b/dask/dataframe/utils.py
@@ -507,9 +507,6 @@ def _check_dask(dsk, check_names=True, check_dtypes=True, result=None, scheduler
             )
             if check_dtypes:
                 assert_dask_dtypes(dsk, result)
-        else:
-            msg = f"Unsupported dask instance {type(dsk)} found"
-            raise AssertionError(msg)
         return result
     return dsk
```

dask_match/tests/test_core.py
```diff
@@ -144,14 +149,26 @@ def test_conditionals(func):
     df["y"] = df.y * 2.0
     ddf = from_pandas(df, npartitions=10)
 
-    assert_eq(func(df), func(ddf))
+    assert_eq(func(df), func(ddf).compute())
+
+
+@pytest.mark.xfail(reason="TODO: Debug this")
+def test_predicate_pushdown(tmpdir):
+    from dask_match.io.parquet import ReadParquet as ReadPq
```
mrocklin, line 157: Why this renaming?  It seems like it might make more sense to just use the `ReadParquet` class in `io.parquet`.  Having two class names like this seems like it complicates things to me.

mrocklin: Oh, maybe this is coming from the difference of having a function and class with the same name.  I suspect that this will become simpler if we name the function `read_parquet` like in dask.dataframe.

mrocklin (commented):
First pull request!  Fun!

I added a few comments.  It's interesting to see someone else interact with this library.  

Any other thoughts/concerns you have about the feel of the library would be welcome.  What could do to improve the ease and understandability of things?

dask_match/core.py
rjzamora: Like much of this PR, this change is just a temporary hack to get the existing tests running :)

The problem I ran into was that calling `self.columns` often requires us to actually parse parquet metadata to calculate `self._meta`, which doesn't work when `"myfile.parquet"` is not a real parquet dataset. Although this shouldn't "break" when the dataset is real, it also feels unnecessary to parse metadata when you are retrieving an attribute that doesn't require you to do so.

dask_match/core.py
rjzamora: Totally agree - Just a temporary redirect to avoid changing some of the existing tests while I played around.

dask_match/io/parquet.py
rjzamora: I think that would be fine. We definitely need a space to "normalize" user inputs, but don't need to do this within the `API` subclasses themselves.

dask_match/tests/test_core.py
rjzamora: yes, exactly. This is another temporary thing - You were too quick :)

dask_match/tests/test_core.py
```diff
@@ -147,8 +152,20 @@ def test_conditionals(func):
...
+    fn = os.path.join(str(tmpdir), "myfile.parquet")
+    pd.DataFrame(
+        {
+            "a": [1, 2, 3, 4, 5] * 10,
+            "b": [0] * 50,
+            "c": range(50),
+        }
+    ).to_parquet(fn)
+
+    df = read_parquet(fn, columns=("a", "b", "c"))
```
mrocklin, line 168: Can I ask for the following?

```suggestion
    original = pd.DataFrame(
        {
            "a": [1, 2, 3, 4, 5] * 10,
            "b": [0] * 50,
            "c": range(50),
        }
    )
    original.to_parquet(fn)

    df = read_parquet(fn)
    assert_eq(df, original)
```

My goal here is to verify that the graph generation works well

mrocklin:
Looks clean to me.  I'm hitting the "Ready for review" button.  I had one final request, mostly to make sure that we cover the actual task graph code.

dask_match/tests/test_core.py
rjzamora: Just a note that the graph logic works but this test is still marked as xfail, since I didn't get a chance to debug the predicate-pushdown behavior just yet.

mrocklin: Ah, thanks for calling this out.  I see what's happening.  This class has a `.columns` attribute, and so that's taking precedence over `._meta.columns`.

The current system puts all of the parameters directly onto the object, which will inevitably conflict with other names, like column names or key attributes (as it does here).

The long-term solution is to have a wrapper class that handles user interaction and then hide all of the expression attributes inside an expr class.  This is also necessary for things like mutation.  This is good motivation for me to think about that class.  I think that you can safely leave this problem with me if you want.

mrocklin: Or, for a short term fix, we would rename the `columns=` parameter to something else

mrocklin: OK, things work on main now

dask_match/io/parquet.py
fjetter: coming a bit late to the party but strong +1 for doing normalization/validation of inputs far up in the stack

mrocklin: This is done in main now.  I've removed the normalize function.  So far I'm still comfortable with Exprs not having `__init__` methods.

dask_match/tests/test_core.py
rjzamora: Cool - Thanks. It looks like we need a few more changes to make the computed result correct. About to submit something now.

PR #2, opened 2023-02-22
Fix predicate-pushdown compute

rjzamora:
Updates `ReadParquet` and `test_predicate_pushdown` to get the correct result after `compute`.

Future **TODO**: Currently, the `read_parquet` engine is not required to perform row-wise filtering to satisfy the `filters` argument. For the pyarrow engine, you **will** get row-wise filtering, but the fastparquet and cudf engines don't do this yet. Since the replacement rules result in the removal of the explicit `Filter` operation, we need to make sure the the partition-wise IO function ensures that row-wise filtering has been applied.

### DIFF
--- a/dask_match/core.py
+++ b/dask_match/core.py
@@ -100,9 +100,6 @@ class Expr(Operation, DaskMethodsMixin, metaclass=_ExprMeta):
     def __getattr__(self, key):
         if key == "__name__":
             return object.__getattribute__(self, key)
-        elif key in type(self)._parameters:
-            idx = type(self)._parameters.index(key)
-            return self.operands[idx]
         elif key in dir(type(self)):
             return object.__getattribute__(self, key)
         elif is_dataframe_like(self._meta) and key in self._meta.columns:
@@ -110,16 +107,15 @@ class Expr(Operation, DaskMethodsMixin, metaclass=_ExprMeta):
         else:
             return object.__getattribute__(self, key)
 
+    def operand(self, key):
+        return self.operands[type(self)._parameters.index(key)]
+
     @property
     def index(self):
         return Index(self)
 
     def __setattr__(self, key, value):
-        if key in type(self)._parameters:
-            idx = type(self)._parameters.index(key)
-            self.operands[idx] = value
-        else:
-            object.__setattr__(self, key, value)
+        object.__setattr__(self, key, value)
 
     def __getitem__(self, other):
         if isinstance(other, Expr):
@@ -374,15 +370,15 @@ class Apply(Elemwise):
 
     @property
     def _meta(self):
-        return self.frame._meta.apply(self.function, *self.args, **self.kwargs)
+        return self.operand("frame")._meta.apply(self.operand("function"), *self.operand("args"), **self.operand("kwargs"))
 
     def _layer(self):
         return {
             (self._name, i): (
                 apply,
                 M.apply,
-                [(self.frame._name, i), self.function] + list(self.args),
-                self.kwargs,
+                [(self.operand("frame")._name, i), self.operand("function")] + list(self.operand("args")),
+                self.operand("kwargs"),
             )
             for i in range(self.npartitions)
         }
@@ -413,20 +409,20 @@ class Projection(Elemwise):
     operation = operator.getitem
 
     def _divisions(self):
-        return self.frame.divisions
+        return self.operand("frame").divisions
 
     @property
     def _meta(self):
-        return self.frame._meta[self.columns]
+        return self.operand("frame")._meta[self.columns]
 
     def _layer(self):
         return {
-            (self._name, i): (operator.getitem, (self.frame._name, i), self.columns)
+            (self._name, i): (operator.getitem, (self.operand("frame")._name, i), self.columns)
             for i in range(self.npartitions)
         }
 
     def __str__(self):
-        base = str(self.frame)
+        base = str(self.operand("frame"))
         if " " in base:
             base = "(" + base + ")"
         return f"{base}[{repr(self.columns)}]"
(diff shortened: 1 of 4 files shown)


mrocklin:
Cool.  Things here seem generally fine to me. 

As a heads-up, I'll be mostly unresponsive at least tomorrow and maybe Friday as well.

mrocklin:
> but the fastparquet and cudf engines don't do this yet

FWIW if this actually becomes a real library then my plan was to use it as an opportunity to reset a lot of things like using fastparquet, and depending on pandas < 2.  If cudf could grow support for filtering that would be nice.

I actually removed the `engine` keyword from your code with this in mind.  I hadn't though of cudf though.

> Since the replacement rules result in the removal of the explicit Filter operation

If you wanted to play with replacement rules you could add back in the engine keyword, and then add in new replacement rules that matched against `ReadParquet(filename, columns=a, filters=b, engine="pyarrow")` and `ReadParquet(filename, columns=a, filters=b, engine="cudf")` or just leave off the cudf replacement rules for now and just force `engine="pyarrow"`.  That way you'll know that these rules won't apply when engine is anything else.

Note that if you do do this then you'll need to leave `self.engine` as a string, and not as an Engine object.


I mention this not because I necessarily want you to do this work (there are probably more important things to sort out) but because I think it might give you a feel for replacement rules, which could be interesting / fun / educational.

rjzamora:
>FWIW if this actually becomes a real library then my plan was to use it as an opportunity to reset a lot of things like using fastparquet, and depending on pandas < 2. If cudf could grow support for filtering that would be nice.

>I actually removed the engine keyword from your code with this in mind. I hadn't though of cudf though.

I mostly agree with this, but will obviously insist on making it as easy as possible to use a cudf backend as early as possible.

>Note that if you do do this then you'll need to leave self.engine as a string, and not as an Engine object.

Ah, good point

>I mention this not because I necessarily want you to do this work (there are probably more important things to sort out) but because I think it might give you a feel for replacement rules, which could be interesting / fun / educational.

Thanks for sharing these thoughts. It is likely that I will end up playing with rules like this. However, the more-immediate issue is probably that tests unrelated to parquet/predicate_pushdown are still failing in `assert_eq`. So we probably want to iron out the best way to validate and compare an expression to an expected pandas result.

mrocklin:
Fair point.  Maybe we should pull over assert_eq into this library and modify it to our own needs?

rjzamora:
>Fair point. Maybe we should pull over assert_eq into this library and modify it to our own needs?

I'm open to this

rjzamora:
>My eyes are back in a decent state. Is this ready for review?

I came down with something yesterday, so haven't pushed on this since last week - You are welcome to review and/or make changes. If I remember correctly, this PR is probably trying to do too many things at once (and pushdown tests are not passing).

dask_match/core.py
```diff
@@ -110,6 +110,10 @@ def __getattr__(self, key):
         else:
             return object.__getattribute__(self, key)
 
+    @property
+    def index(self):
+        return Index(self)
```
rjzamora, line 115: I'm wondering if we want various base classes, like `FrameExpr`, `DataFrameExpr`, `SeriesExpr`, and `ScalarExpr` to make sure we only expose attributes like `index` when it makes sense.

dask_match/io/parquet.py
```diff
@@ -62,6 +71,23 @@ class ReadParquet(IO):
...
+        return self.operands[self._parameters.index("columns")]
+
+    @property
+    def columns(self):
+        if self.input_columns is None:
+            return self._meta.columns
+        else:
+            import pandas as pd
+
+            return pd.Index(_list_columns(self.input_columns))
```
rjzamora, line 85: Note that overlapping parameters and properties can be a bit confusing, and I don't really like the solution used here.

We may want to require each class to define a set of reserved names, and forbid any parameter name from intersecting with that set. This means the user-facing `read_parquet` API, for example, would need to translate arguments like `columns` into a *different* name (like `column_projection`), so that the `ReadParquet` implementation wouldn't need to worry about implementing anything like the `input_columns` workarounds used here.

dask_match/core.py
```diff
@@ -100,22 +100,22 @@ def __repr__(self):
...
-            return self.operands[idx]
         elif key in dir(type(self)):
             return object.__getattribute__(self, key)
         elif is_dataframe_like(self._meta) and key in self._meta.columns:
             return self[key]
         else:
             return object.__getattribute__(self, key)
 
+    def operand(self, key):
+        return self.operands[type(self)._parameters.index(key)]
```
rjzamora, line 111: Note that commit [6f191e4](https://github.com/mrocklin/dask-match/pull/2/commits/6f191e46381f37d5c1b0c9ee20bc581cce4c46c1) demonstrates one possible approach to the proposal in #4 (adding a distinct `operand` API for accessing parameters used to create the expression). Although this prohibits us from using `self.<operand>` syntax, I feel that sacrificing this short-hand makes it much easier to avoid recursion traps and unexpected attribute/column-name collisions.

dask_match/core.py
```diff
@@ -409,25 +409,45 @@ class Projection(Elemwise):
...
+    def _divisions(self):
+        return self.operand("frame").divisions
+
+    @property
+    def _meta(self):
+        return self.operand("frame")._meta.index
+
+    def _layer(self):
+        return {
+            (self._name, i): (getattr, (self.operand("frame")._name, i), "index")
```
mrocklin, line 446: Why was `.operand` necessary here?  This seems unpleasant to do for all operands.  I'm optimizing pretty hard here for "pleasant to work with"

rjzamora: >Why was .operand necessary here? This seems unpleasant to do for all operands. I'm optimizing pretty hard here for "pleasant to work with"

I'm totally on board with the "pleasant to work with" goal, and so the `operand` proposal may not be the best solution. Overall, I found the current practice of accessing parameters/operands as attributes (e.g. `self.<operand>`) to be surprisingly problematic, and eventually rationalized the idea to forbid the practice altogether.  I realize that `self.<operand>` pattern *can* be supported just fine, but I'm not quite convinced that it is worth the potential for pain (though my mind is open).

mrocklin:
Anything further to do here @rjzamora ?

rjzamora:
These changes were addressed in #6

Issue #5, opened 2023-03-08
Create separate Expr and API classes


mrocklin:
There are a few issues that motivate the desire to have a separate encapulsating class

```python
class DataFrame:
    expr: Expr

class Series:
    expr: Expr

...
```

Those issues are as follows:

## Mutation

For cases like the following:

```python
df["z"] = df.x + df.y
```

We need to be able to do something like the following:

```python
def __setitem__(self, column, value):
    expr = self.expr.assign(column=value)
    self.expr = expr  # we need to be able to do this for mutation
```

This requires that the object the user has needs to be able to have all of its metadata rewritten, and the underlying operation needs to be able to be changed.  Relying on types for the operation means that this is a no-go with the current single class architecture.

## Class Explosion

DataFrame, Series, and Index have different APIs that we'll need to present to the user.  The easiest way to do this is with new classes.  However if we're also using classes for the operation (Sum, Add, Join) then we enter `n by k` subclass hell.  Composition handles this fine.

## Poluting the API

Expressions have various attributes like `.frame`, `.left`, `.right`, which we probably don't want to present to the user because it'll end up confusing things with column attributes and so on.  We probably want to hide these a bit.  With the proposed architecture these become `df.expr.left` which is appropriately hidden.  We'll always have the parameters take precedence over anything else, and this will probably be fine because users won't touch this layer by default.

## Concerns

I haven't done this yet just because it adds a bit of development inertia (have to add two methods for everything, or else drop syntactic sugar for Exprs).

mrocklin:
> Therefore, I don't see any great reason not to adopt the composition pattern from the start

Mostly my thinking was to keep things nimble while we reinvent a bunch.  Inertia is bad at this stage.

However, if you're enthused then do you want to give it a shot and see how it feels?

PR #6, opened 2023-03-23
Separate the collection-API from the `Expr` classes

rjzamora:
Rough implementation of the plan discussed in #5 - This PR corresponds to a refactoring of `dask_match` so that users interact with an API based on the `FrameBase` class, instead of interacting with `Expr` classes directly.

**NOTES**:
- The current implementation does not go "all in" on enforcing a completely distinct API class yet.  More specifically, the `FrameBase.__getattr__` method will still "fall back" to a wrapped version of the underlying `Expr` attribute when an `AttributeError` is encountered. The advantage of this approach is that we can avoid having two methods for everything, and can thus limit unnecessary developer inertia for now (which I think is a good selling point).
- There are probably many loose ends that I haven't bothered to find/fix yet, because I don't want to waste time cleaning up something that won't be adopted :)
- The current state of this PR includes the primary "fixes" from https://github.com/mrocklin/dask-match/pull/2
- This PR still includes the `operand` API suggested in https://github.com/mrocklin/dask-match/issues/4 and used in https://github.com/mrocklin/dask-match/pull/2, though it is now only used when the name of an operand collides with a reserved `Expr` property name (e.g. `_meta`, `divisions`, `index`, etc.). After some experimentation, and lots of recursion debugging, I decided that it is probably a bad idea to rely on special logic in `_meta`/`divisions` to check if the object happens to have an operand with the same name (because sometimes we actually **don't** want the operand!). Instead, we should only make it possible to access operands as attributes if there is no name collision, otherwise the `operand` API should be used to access the operand unambiguously.


mrocklin:
Oh cool.  I look forward to reviewing this this afternoon.

mrocklin:
> Oh cool.  I look forward to reviewing this this afternoon.

This didn't happen, but I'll now repeat the same mistake and promise to review again this afternoon 🙂

mrocklin:
> More specifically, the FrameBase.__getattr__ method will still "fall back" to a wrapped version of the underlying Expr attribute when an AttributeError is encountered

I'm not seeing this in the code currently.  Is this all pushed up?

If it's easy, would you be up for implementing Assign, and then implementing `__setitem__` or `__delitem__` to verify that mutation works?

```python
# example
# df = df.assign(a=df.x + df.y, b=df.x - df.y)

def __setitem__(self, key, value):
        out = self.assign(**{key: value})
        self._expr = out._expr
        return self

def __delitem__(self, key):
    columns = [c for c in self.columns if c != key]
    out = self[columns]
    self._expr = out._expr
```

dask_match/core.py
```diff
@@ -1,628 +1,279 @@
...
-class Elemwise(Blockwise):
-    """
-    This doesn't really do anything, but we anticipate that future
-    optimizations, like `len` will care about which operations preserve length
-    """
-
-    pass
-
+    def size(self):
+        return new_collection(self.expr.size)
```
mrocklin, line 133: These can probably move up to the super class I think

dask_match/core.py
```diff
@@ -1,628 +1,279 @@
...
+            except AttributeError:
+                # Fall back to `BaseFrame.__getattr__`
+                return super().__getattr__(key)
 
-    _parameters = ["frame", "dtypes"]
-    operation = M.astype
+    def __getitem__(self, other):
+        if isinstance(other, FrameBase):
+            return new_collection(self.expr.__getitem__(other.expr))
+        return new_collection(self.expr.__getitem__(other))
```
mrocklin, line 153: This can move up too?

dask_match/core.py
```diff
@@ -1,628 +1,279 @@
...
+    return new_collection(FromGraph(*args, **kwargs))
+
+
+def read_csv(*args, **kwargs):
+    from dask_match.expr.io.csv import ReadCSV
+
+    return new_collection(ReadCSV(*args, **kwargs))
+
+
+def read_parquet(
```
mrocklin, line 234: Maybe the functions above can move into some namespace module, like `__init__.py` without being in core.py

dask_match/expr/core.py
```diff
@@ -0,0 +1,578 @@
+import functools
```
mrocklin, line 1: Maybe de-nest the `expr/` directory?  It seems like this is most of the real code, while the current `core.py` is actually more of the non-core element, wrapping API around.  I would probably flatten everything, and rename core.py to something else.  I'm not sure what exactly.  `api.py`?  Something better would be welcome.  

In general though, keeping things more flat would make me happy.

dask_match/expr/core.py
```diff
@@ -0,0 +1,578 @@
...
+            (self._name, i): (
+                self.operation,
+                (self.left._name, i) if isinstance(self.left, Expr) else self.left,
+                (self.right._name, i) if isinstance(self.right, Expr) else self.right,
+            )
+            for i in range(self.npartitions)
+        }
+
+    def __str__(self):
+        return f"{self.operand('left')} {self._operator_repr} {self.operand('right')}"
```
mrocklin, line 445: My preference would be to leave these as `self.left` and `self.right`.  This is subjective though and I'm open to a larger discussion on this topic.  I suspect that others may agree with you.

dask_match/expr/core.py
rjzamora: Yeah, I'm okay with de-nesting now - I first expected there to be more "api' code, and thought we *may* want to leave space for "array" expressions. However, I now agree that it would be simpler to use something like `api.py`, `frame.py`, or `collection.py` for collection code.

dask_match/core.py
```diff
@@ -1,628 +1,279 @@
...
+            try:
+                # Fall back to `expr` API
+                # (Making sure to convert to/from Expr)
+                val = getattr(self.expr, key)
+                if callable(val):
+                    return functools.partial(_wrap_expr_api, wrap_api=val)
+                return val
+            except AttributeError:
+                # Raise original error
+                raise err
```
rjzamora, line 97: @mrocklin - This is where `FrameBase.__getattr__` falls back to the wrapped version of the `Expr` method. Is this what you were asking about?

dask_match/expr/core.py
rjzamora: Called it `collection.py` for now, but this can be changed pretty easily if we decide on something else.

dask_match/tests/test_collection.py
```diff
@@ -5,64 +5,101 @@
...
+    assert "b" in x.columns
+
+    # Check __delitem__
+    del x["a"]
+    assert "a" not in x.columns
+    assert "b" in x.columns
+
+    # Check __setitem__
+    x.a = x.b + 1
+    assert_eq(x.a.compute(), (x.b + 1).compute(), check_names=False)
```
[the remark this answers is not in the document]
mrocklin: It looks like this was hiding some dtype sadness.

dask_match/tests/test_collection.py
```diff
@@ -5,64 +5,101 @@
...
+
+    # Check assign
+    y = x.assign(d=x.a + x.b, e=x.b + x.c)
+    assert_eq(y.d.compute(), (x.a + x.b).compute(), check_names=False)
+    assert_eq(y.e.compute(), (x.b + x.c).compute(), check_names=False)
+
+
+def df(fn):
+    return read_parquet(fn, columns=["a", "b", "c"])
```
mrocklin: We should probably not rely on these in the basic test suite. Maybe it's time to separate.

mrocklin (commented):
Thanks for this work @rjzamora in principle the approach here seems sensible to me.  

There are a few places where I think things could be cleaner.  I've gone ahead and pushed some some changes to address these.  I hope that you don't mind.  I'm adding these comments to highlight the changes, in hopes that it builds alignment

mrocklin:
@rjzamora when you have a chance can I ask you to review my recent commits.   If you're good then I'm good to merge.

rjzamora:
Thanks for cleaning up @mrocklin - Agree with your comments (was just holding off on cleanup until I was a bit more confident you were on board with the general class organization).

Your recent commits seem fine to me, but tests are now failing for me locally.

dask_match/tests/test_collection.py
```diff
@@ -0,0 +1,162 @@
...
+
+    cc = 2 * aa - 3 * bb
+    assert set(cc.columns) == {"x", "y", "z"}
+
+
+def test_dask(df, ddf):
+    assert (ddf.x + ddf.y).npartitions == 10
+    z = (ddf.x + ddf.y).sum()
+
+    assert assert_eq(z, (df.x + df.y).sum())
```
rjzamora, line 71: We cant use `assert_eq` on `Scalar` objects using `dask:main`. We can do `assert z == (df.x + df.y).sum()` here, modify `dask/dataframe/utils/assert_eq`, or implement our own version of `assert_eq` in dask-match.

mrocklin: Ah indeed.  I had that change locally.  

Well, rather than work around this, lets fix the problem: https://github.com/dask/dask/pull/10125

In general it seems like we're pushing a bit more in this direction.  It's probably time to make this effort a bit cleaner in general, and avoid workarounds where we can.

mrocklin:
Good to go @rjzamora ?

PR #11, opened 2023-03-31
Enable task fusion

rjzamora:
Uses `SubgraphCallable` to enable task fusion between expressions satisfying the `Fusable` protocol.

### DIFF
--- a/dask_match/core.py
+++ b/dask_match/core.py
@@ -3,11 +3,13 @@ import numbers
 import operator
 from collections import defaultdict
 from collections.abc import Iterator
+from typing import Protocol, runtime_checkable
 
 import pandas as pd
 import toolz
 from dask.base import normalize_token, tokenize
 from dask.dataframe import methods
+from dask.optimization import SubgraphCallable
 from dask.utils import M, apply, funcname
 from matchpy import (
     Arity,
@@ -69,7 +71,6 @@ class Expr(Operation, metaclass=_ExprMeta):
 
     commutative = False
     associative = False
-    fusable = False
     _parameters = []
     _defaults = {}
 
@@ -115,9 +116,12 @@ class Expr(Operation, metaclass=_ExprMeta):
     @property
     def dependencies(self):
         # Non-uniform dependencies.
-        # Usually corresponds to `Expr` operands,
-        # but may include BlockwiseDep objects.
-        return [operand for operand in self.operands if isinstance(operand, Expr)]
+        # Corresponds to `Expr` and `MappedArg` operands
+        return [
+            operand
+            for operand in self.operands
+            if isinstance(operand, (Expr, MappedArg))
+        ]
 
     @property
     def index(self):
@@ -256,6 +260,17 @@ class Expr(Operation, metaclass=_ExprMeta):
     def _meta(self):
         raise NotImplementedError()
 
+    def _layer(self):
+        if isinstance(self, Fusable):
+            return _subgraph_callable_layer(
+                self._name,
+                self._subgraph_callable(),
+                self.dependencies,
+                self.npartitions,
+            )
+        else:
+            raise NotImplementedError
+
     def __dask_graph__(self):
         """Traverse expression tree, collect layers"""
 
@@ -264,7 +279,7 @@ class Expr(Operation, metaclass=_ExprMeta):
         layers = []
         fusable = _fusable_ops(self)
         while stack:
-            expr = _fuse_blockwise_deps(stack.pop(), fusable)
+            expr = _fuse_expr_groups(stack.pop(), fusable)
 
             if expr._name in seen:
                 continue
@@ -290,7 +305,6 @@ class Blockwise(Expr):
     """
 
     operation = None
-    fusable = True
 
     @property
     def _meta(self):
@@ -319,7 +333,7 @@ class Blockwise(Expr):
     def _name(self):
         return funcname(self.operation) + "-" + tokenize(*self.operands)
 
-    def _block(self):
+    def _subgraph_callable(self):
         return {
             self._name: (
                 apply,
@@ -331,14 +345,6 @@ class Blockwise(Expr):
                 self._kwargs,
             )
         }
-
-    def _layer(self):
-        return _blockwise_layer(
-            self._name,
-            self._block(),
-            self.dependencies,
-            self.npartitions,
-        )
 
 
 class Elemwise(Blockwise):
@@ -368,7 +374,7 @@ class Apply(Elemwise):
     def _meta(self):
         return self.frame._meta.apply(self.function, *self.args, **self.kwargs)
 
-    def _block(self):
+    def _subgraph_callable(self):
         return {
             self._name: (
                 apply,
@@ -389,7 +395,7 @@ class Assign(Elemwise):
     def _meta(self):
         return self.frame._meta.assign(**{self.key: self.value._meta})
 
-    def _block(self):
+    def _subgraph_callable(self):
         return {
             self._name: (
                 methods.assign,
@@ -438,7 +444,7 @@ class Projection(Elemwise):
     def _meta(self):
         return self.frame._meta[self.columns]
 
-    def _block(self):
+    def _subgraph_callable(self):
         return {self._name: (operator.getitem, self.frame._name, self.columns)}
 
     def __str__(self):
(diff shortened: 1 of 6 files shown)


dask_match/core.py
```diff
@@ -247,13 +250,26 @@ def dtypes(self):
...
     def __dask_graph__(self):
         """Traverse expression tree, collect layers"""
+
         stack = [self]
         seen = set()
         layers = []
+        fusable = _fusable_ops(self)
         while stack:
-            expr = stack.pop()
+            expr = _fuse_expr_groups(stack.pop(), fusable)
```
mrocklin, line 272: I don't think that this concept should be embedded directly into the Expr class.  This feels like a specific abstraction finding its way into what should be very general (the US Tax code is a good example of this).

Instead, I would expect us to have a pass that crawled over the expression tree, found groups that could be fused into a single Blockwise object, and then did that, returning a new expression tree with fewer nodes in it.

mrocklin:
Thanks for starting this @rjzamora .  I'm glad to see how something like this can be done, as you say, pretty easily.  I'm curious how simple/separable we can make this.

If I were to approach this problem, I would probably avoid the `IO` case at first, and just focus on blockwise operations.  I would probably make a new Expr class `BlockwiseFused` or something, and add a new optimization function that swaps out lots of Blockwise/Fusable operations with fewer `BlockwiseFused` operations.  I would then break open `optimize_expr` to have both this, and the term-rewrite-system stuff from matchpy as different passes. 

I probably wouldn't worry yet about things like config.

Then, once we got that very clean, I would then consider complications to that, like IO, and see if different things made sense.

dask_match/core.py
rjzamora: >Instead, I would expect us to have a pass that crawled over the expression tree, found groups that could be fused into a single Blockwise object, and then did that, returning a new expression tree with fewer nodes in it.

Yes. My original plan was to actually fuse `Blockwise` expressions into new `Blockwise` expressions like we do in for `Blockwise` layers in dask/dask. However, I soon realized that I actually **dislike** that we need to do this in dask/dask.  The opaque composite layers we end up crating tend to loose most of their value/meaning after fusion (besides their ability to ensure that the appropriate "low-level" tasks get fused).

My mind is still very open on the subject (and is very likely to change), because I do appreciate the logical simplicity of treating fusion as a distinct `Expr` optimization pass. With that in mind, I currently like the idea of only applying fusion at graph-materialization time.

rjzamora:
>If I were to approach this problem, I would probably avoid the IO case at first, and just focus on blockwise operations.

This probably makes sense. The critical importance of IO fusion in the past affected my goals here. I'm sure task queuing makes this less critical now.

>I would probably make a new Expr class BlockwiseFused or something, and add a new optimization function that swaps out lots of Blockwise/Fusable operations with fewer BlockwiseFused operations. I would then break open optimize_expr to have both this, and the term-rewrite-system stuff from matchpy as different passes.

As I mentioned [in a comment above](https://github.com/mrocklin/dask-match/pull/11#discussion_r1154749180), this plan makes a lot of sense and is exactly what I originally wanted to do. There is a good chance that I'll come back around to wanting to do something like this. However, at the moment I'm a bit worried about rewriting the expression tree for graph-related purposes like fusion and/or culling. Instead, I'd like these these consideration to only come into play at graph-materialization time, instead of at `Expr`-optimization time.

mrocklin:
> With that in mind, I currently like the idea of only applying fusion at graph-materialization time.

At the moment I'm not convinced.  I'm open to talk about this live if you like.

> However, at the moment I'm a bit worried about rewriting the expression tree for graph-related purposes like fusion and/or culling.

I don't think that these matter any longer at this level.

rjzamora:
>At the moment I'm not convinced. I'm open to talk about this live if you like.

I understand the feeling. I'm trying to convince myself that it's a good idea to use a second round of replacement for fusion, and that we can leave out IO. My gut tells me that we are walking into a trap in both cases, but my lack in confidence tells me I should keep pondering things.

I suppose we are probably on the same page about wanting "simplicity."  For me, it just seems pretty complex to add a specialized replacement phase for the sole purpose of `Blockwise` fusion when the only thing one "should" be able to do after such an optimization is performed is materialize the graph. However, this design is consistent with the current behavior of HLG layers, and users may like to visualize fusion at the Expr-tree level.  Therefore, I can imagine myself getting on board with this.

At the moment, I'm actually a bit more worried about leaving out IO in the original design, unless I know for sure that other mechanisms can guarantee that an IO task will effectively be performed within the same compute unit as a succeeding `Blockwise` task, without there being any chance for spilling or work-stealing in-between. I would hate to design a "complicated but elegant" solution that formalizes Expr-level fusion while leaving out IO, only to need to add back an IO complication later on. If I discover that my worries about IO existing as a distinct task are unfounded, then I can get on board on this front as well.

Overall, I do want the **final** design to be as simple as possible.

>I don't think that these matter any longer at this level.

Can you expand on this a bit? Right now the `Expr._layer` implementations all assume that we are generating tasks for every partition in the collection. If we want to support "efficient" `head`/`tail`, or leverage `divisions`/partition-statistics,  we will probably want to avoid materializing the full layer for each expression.

mrocklin:
Do you have some time to chat through things live?  I think that having the
two of us sync up on this would be helpful.  My calendar is available at
https://calendly.com/mrocklin if any of those times work for you.  If you
have a calendar I'm also happy to work from that.

mrocklin:
> For me, it just seems pretty complex to add a specialized replacement phase for the sole purpose of Blockwise fusion when the only thing one "should" be able to do after such an optimization is performed is materialize the graph

I think that when you say "replacement phase" you mean the term I know of as "optimization pass".  If so, then I don't think that this is complex because I think that we're going to have many different optimization passes.  If I can write something as an optimization pass then I'm going to do so because that's an easy place to expand broadly.  

> I'm actually a bit more worried about leaving out IO in the original design,

Ideally we would design something that wouldn't "paint ourselves into a corner".  We'd be smart enough to know that we coud go down this path easily in the future, without explicitly going down that path.  Writing simple things that don't result in traps is, I think, the key to simple and iterative design.

> Right now the Expr._layer implementations all assume that we are generating tasks for every partition in the collection. If we want to support "efficient" head/tail, or leverage divisions/partition-statistics, we will probably want to avoid materializing the full layer for each expression

I think that high level optimization will give us what we want here.  For example I would expect optimizations to push `Head` down pretty far in the expression tree.  Same with any other single-partition operation.

mrocklin:
@rjzamora and I spoke.  I'm more open to the IO fusion, especially if we include the fusion of multiple inputs.  The conversation helped to remind me of the following case:

```python
a = df.read_parquet(..., columns="a")
b = df.read_parquet(..., columns="b")

(a + b).sum().compute()
```

We want to make sure that the partitions of `a` and the partitions of `b` are loaded onto the same machine.

mrocklin:
In symbolic expression tree parlance I think that I'm hearing two things:

1. We need to get better at tree traversals and replacement strategies
2. We need to support expressions with *variadic* dependencies

I agree with both of these statements.  I don't know if matchpy has a

rjzamora:
>For the first point, I wouldn't think of this as replacement, but rather rebuilding. This should be relatively easy because type(obj)(*obj.operands) should just work in all cases (this is what we get by not implementing __init__ anywhere.

Yes, agree with this. The tree rebuilding logic is not difficult.

>For the second point matchpy certainly supports variadic expressions. We'll have to make sure that we don't break this anywhere in our code, like ExprMeta

Makes sense. If matchpy already supports it, then I'll try to work out a clean way to handle variadic dependencies in Expr.

mrocklin:
> Makes sense. If matchpy already supports it, then I'll try to work out a clean way to handle variadic dependencies in Expr

I'm not sure that we have to do anything here.  Here is an example that would maybe work?

```python
class AddMany(Expr):
    @property
    def _meta(self):
        out = self.operands[0]
        for op in self.operands:
            if hasattr(op, "_meta"):
                op = op._meta()
            out = out + op
        return out

    def _divisions(self):
        return self.operands[0].divisions
```

rjzamora:
You are correct - I didn't realize I was allowed to pass in operands without naming them. My mistake!

rjzamora:
Closing this in favor of #15

PR #15, opened 2023-04-05
Add dedicated optimization pass for fusion

### LINKED ISSUE #9: [Question] What is the best plan for task fusion
Dask-match certainly makes most high-level optimizations simpler than they would be with Dask's current HLG design. However, it is not yet clear to me how task fusion should come into play. There seem to be three general paths forward:

1. We do nothing about task fusion on the expression side.  Instead, we rely on the scheduler to either perform low-level fusion, or to use something like speculative task assignment.
2. We implement a new mechanism within `dask-match` to enable the fusion of "blockwise" parent and child expressions.
3. We modify the "blockwise" `_layer` methods to output `Blockwise` HLG layers instead of low-level dictionaries.

I like that (1) allows use to keep the logic in `dask-match` a bit simpler, but I'm not very confident that this approach will deliver the same benefits as (2) or (3). (3) is clearly the "fastest" way to support task fusion, but we will eventually need to migrate the necessary subgraph-callable logic into the `Expr` version of `Blockwise` if the long-term plan is to scrap HLGs altogether.

rjzamora:
This PR repurposes most of the code in #11 to enable `Blockwise`/`IO` fusion as a distinct optimization pass. Note that this iteration is designed to compose and fuse arbitrary groups of `Fusable` `Expr`s where no members of the group have external "dependents" (e.g. diamond-shaped sub-graphs).

Supersedes #11
Closes #9 

**TODO**:
- [x] Clean up and organize new utilities

### DIFF
--- a/dask_match/core.py
+++ b/dask_match/core.py
@@ -374,6 +374,7 @@ class Blockwise(Expr):
     def _name(self):
         return funcname(self.operation) + "-" + tokenize(*self.operands)
 
+    @functools.lru_cache
     def _blockwise_subgraph(self):
         args = tuple(
             operand._name if isinstance(operand, Expr) else operand
@@ -384,6 +385,7 @@ class Blockwise(Expr):
         else:
             return {self._name: (self.operation,) + args}
 
+    @functools.lru_cache
     def _layer(self):
         # Use BlockwiseArg to broadcast dependencies (if necessary)
         dependencies = [
@@ -878,6 +880,7 @@ class FusedExpr(Blockwise):
     def dependencies(self):
         return self.operands[1:]
 
+    @functools.lru_cache
     def _blockwise_subgraph(self):
         block = {self._name: self.exprs[0]._name}
         for _expr in self.exprs:
--- a/dask_match/core.py
+++ b/dask_match/core.py
@@ -3,7 +3,6 @@ import numbers
 import operator
 from collections import defaultdict
 from collections.abc import Iterator
-from typing import Protocol, runtime_checkable
 
 import pandas as pd
 import toolz
@@ -318,7 +317,7 @@ class Blockwise(Expr):
         # This property is required to enable fusion
         return [operand for operand in self.operands if isinstance(operand, Expr)]
 
-    def _block_subgraph(self):
+    def _blockwise_subgraph(self):
         return {
             self._name: (
                 apply,
@@ -332,12 +331,26 @@ class Blockwise(Expr):
         }
 
     def _layer(self):
-        return _subgraph_callable_layer(
+        # Create SubgraphCallable
+        dependencies = self._subgraph_dependencies()
+        func = SubgraphCallable(
+            self._blockwise_subgraph(),
             self._name,
-            self._block_subgraph(),
-            self._subgraph_dependencies(),
-            self.npartitions,
+            [dep._name for dep in dependencies],
+            self._name,
         )
+
+        # Tasks depend on external-dependency keys only
+        return {
+            (self._name, i): (
+                func,
+                *[
+                    dep[i] if isinstance(dep, MappedArg) else (dep._name, i)
+                    for dep in dependencies
+                ],
+            )
+            for i in range(self.npartitions)
+        }
 
 
 class Elemwise(Blockwise):
@@ -367,7 +380,7 @@ class Apply(Elemwise):
     def _meta(self):
         return self.frame._meta.apply(self.function, *self.args, **self.kwargs)
 
-    def _block_subgraph(self):
+    def _blockwise_subgraph(self):
         return {
             self._name: (
                 apply,
@@ -388,7 +401,7 @@ class Assign(Elemwise):
     def _meta(self):
         return self.frame._meta.assign(**{self.key: self.value._meta})
 
-    def _block_subgraph(self):
+    def _blockwise_subgraph(self):
         return {
             self._name: (
                 methods.assign,
@@ -437,7 +450,7 @@ class Projection(Elemwise):
     def _meta(self):
         return self.frame._meta[self.columns]
 
-    def _block_subgraph(self):
+    def _blockwise_subgraph(self):
         return {self._name: (operator.getitem, self.frame._name, self.columns)}
 
     def __str__(self):
@@ -460,7 +473,7 @@ class ProjectIndex(Elemwise):
     def _meta(self):
         return self.frame._meta.index
 
-    def _block_subgraph(self):
+    def _blockwise_subgraph(self):
         return {self._name: (getattr, self.frame._name, "index")}
 
 
@@ -485,7 +498,7 @@ class Binop(Elemwise):
     _parameters = ["left", "right"]
     arity = Arity.binary
 
-    def _block_subgraph(self):
+    def _blockwise_subgraph(self):
         return {
             self._name: (
                 self.operation,
@@ -653,23 +666,23 @@ def replace_nodes(expr, nodes_to_replace: dict):
         return nodes_to_replace[expr]
 
     new_operands = []
-    new_count = 0
+    update = False
     for operand in expr.operands:
         if isinstance(operand, Expr) and operand in nodes_to_replace:
             # Replacing this operand
             val = nodes_to_replace[operand]
-            new_count += 1
+            update = True
         elif isinstance(operand, Expr):
             # Non-Expr operand - Recursive call
             val = replace_nodes(operand, nodes_to_replace)
             if operand._name != val._name:
-                new_count += 1
+                update = True
         else:
             # Non-Expr operand
             val = operand
         new_operands.append(val)
 
-    if new_count:
+    if update:
         # Only return new object if something changed
         return type(expr)(*new_operands)
     return expr
@@ -691,7 +704,7 @@ def _blockwise_fusion(expr):
                 continue
             seen.add(next._name)
 
-            if isinstance(next, Fusable):
+            if isinstance(next, Blockwise):
                 dependencies[next] = set()
                 if next not in dependents:
                     dependents[next] = set()
@@ -699,18 +712,18 @@ def _blockwise_fusion(expr):
             for operand in next.operands:
                 if isinstance(operand, Expr):
                     stack.append(operand)
-                    if isinstance(operand, Fusable):
+                    if isinstance(operand, Blockwise):
                         if next in dependencies:
                             dependencies[next].add(operand)
                         dependents[operand].add(next)
 
         # Traverse each "root" until we find a fusable sub-group.
-        # Here we use root to refer to a Fusable Expr node that
-        # has no Fusable dependents
+        # Here we use root to refer to a Blockwise Expr node that
+        # has no Blockwise dependents
         roots = [
             k
             for k, v in dependents.items()
-            if v == set() or all(not isinstance(_expr, Fusable) for _expr in v)
+            if v == set() or all(not isinstance(_expr, Blockwise) for _expr in v)
         ]
         while roots:
             root = roots.pop()
@@ -762,29 +775,7 @@ def _blockwise_fusion(expr):
     return expr
 
 
-@runtime_checkable
-class Fusable(Protocol):
-    """Fusable Expr Protocol
-    An `Expr` with these methods will be treated as "fusable"
-    """
-
-    def _subgraph_dependencies(self):
-        """List of `Expr` operands or `MappedArg`
-        dependencies. Since `_block_subgraph`
-        cannot include mapped arguments explicitly,
-        IO expressions must represent these task
-        arguments as `MappedArg` dependencies.
-        """
-        raise NotImplementedError
-
-    def _block_subgraph(self):
-        """Return the subgraph for an abstract 'block'.
-        Used to initialize a `SubgraphCallabe` object.
-        """
-        raise NotImplementedError
-
-
-class FusedExpr(Expr):
+class FusedExpr(Blockwise):
     @property
     def exprs(self):
         return self.operands[0]
@@ -807,20 +798,12 @@ class FusedExpr(Expr):
     def _subgraph_dependencies(self):
         return self.operands[1:]
 
-    def _block_subgraph(self):
+    def _blockwise_subgraph(self):
         block = {self._name: self.exprs[0]._name}
         for _expr in self.exprs:
-            for k, v in _expr._block_subgraph().items():
+            for k, v in _expr._blockwise_subgraph().items():
                 block[k] = v
         return block
-
-    def _layer(self):
-        return _subgraph_callable_layer(
-            self._name,
-            self._block_subgraph(),
-            self._subgraph_dependencies(),
-            self.npartitions,
-        )
 
 
 class MappedArg:
@@ -831,39 +814,16 @@ class MappedArg:
     in a fusion-compatible way.
     """
 
-    def __init__(self, lookup):
+    def __init__(self, lookup, token=None):
         assert callable(lookup) or isinstance(lookup, (list, dict))
         self.lookup = lookup
+        self.token = token or tokenize(self.lookup)
 
     @property
     def _name(self):
-        return f"mapped-arg-{tokenize(self.lookup)}"
+        return f"mapped-arg-{self.token}"
 
     def __getitem__(self, index):
         if callable(self.lookup):
             return self.lookup(index)
         return self.lookup[index]
-
-
-def _subgraph_callable_layer(name, block, dependencies, npartitions, funcname=None):
-    """Construct a low-level blockwise layer"""
-
-    # Convert `_block` logic to SubgraphCallable
-    func = SubgraphCallable(
-        block,
-        name,
-        [dep._name for dep in dependencies],
-        funcname or name,
-    )
-
-    # Tasks depend on operand keys only
-    return {
-        (name, i): (
-            func,
-            *[
-                dep[i] if isinstance(dep, MappedArg) else (dep._name, i)
-                for dep in dependencies
-            ],
-        )
-        for i in range(npartitions)
-    }
--- a/dask_match/core.py
+++ b/dask_match/core.py
@@ -769,7 +769,7 @@ def _blockwise_fusion(expr):
 
 class FusedExpr(Blockwise):
     """Fused ``Blockwise`` expression
-    
+
     A ``FusedExpr`` corresponds to the fusion of multiple
     ``Blockwise`` expressions into a single ``Expr`` object.
     Before graph-materialization time, the behavior of this
@@ -791,6 +791,7 @@ class FusedExpr(Blockwise):
         operands, because those arguments should already be
         captured in the fused subgraphs.
     """
+
     _parameters = ["exprs"]
 
     @property
--- a/dask_match/core.py
+++ b/dask_match/core.py
@@ -1,12 +1,15 @@
 import functools
 import numbers
 import operator
+from collections import defaultdict
 from collections.abc import Iterator
+from typing import Protocol, runtime_checkable
 
 import pandas as pd
 import toolz
 from dask.base import normalize_token, tokenize
 from dask.dataframe import methods
+from dask.optimization import SubgraphCallable
 from dask.utils import M, apply, funcname
 from matchpy import (
     Arity,
@@ -91,6 +94,9 @@ class Expr(Operation, metaclass=_ExprMeta):
 
     def __repr__(self):
         return str(self)
+
+    def __hash__(self):
+        return hash(self._name)
 
     def __getattr__(self, key):
         try:
@@ -307,19 +313,31 @@ class Blockwise(Expr):
     def _name(self):
         return funcname(self.operation) + "-" + tokenize(*self.operands)
 
-    def _layer(self):
+    def _subgraph_dependencies(self):
+        # List `Expr` operands only.
+        # This property is required to enable fusion
+        return [operand for operand in self.operands if isinstance(operand, Expr)]
+
+    def _block_subgraph(self):
         return {
-            (self._name, i): (
+            self._name: (
                 apply,
                 self.operation,
                 [
-                    (operand._name, i) if isinstance(operand, Expr) else operand
+                    operand._name if isinstance(operand, Expr) else operand
                     for operand in self.operands
                 ],
                 self._kwargs,
             )
-            for i in range(self.npartitions)
         }
+
+    def _layer(self):
+        return _subgraph_callable_layer(
+            self._name,
+            self._block_subgraph(),
+            self._subgraph_dependencies(),
+            self.npartitions,
+        )
 
 
 class Elemwise(Blockwise):
@@ -349,15 +367,14 @@ class Apply(Elemwise):
     def _meta(self):
         return self.frame._meta.apply(self.function, *self.args, **self.kwargs)
 
-    def _layer(self):
+    def _block_subgraph(self):
         return {
-            (self._name, i): (
+            self._name: (
                 apply,
                 M.apply,
-                [(self.frame._name, i), self.function] + list(self.args),
+                [self.frame._name, self.function] + list(self.args),
                 self.kwargs,
             )
-            for i in range(self.npartitions)
         }
 
 
@@ -371,15 +388,14 @@ class Assign(Elemwise):
     def _meta(self):
         return self.frame._meta.assign(**{self.key: self.value._meta})
 
-    def _layer(self):
+    def _block_subgraph(self):
         return {
-            (self._name, i): (
+            self._name: (
                 methods.assign,
-                (self.frame._name, i),
+                self.frame._name,
                 self.key,
-                (self.value._name, i),
+                self.value._name,
             )
-            for i in range(self.npartitions)
         }
 
 
@@ -421,11 +437,8 @@ class Projection(Elemwise):
     def _meta(self):
         return self.frame._meta[self.columns]
 
-    def _layer(self):
-        return {
-            (self._name, i): (operator.getitem, (self.frame._name, i), self.columns)
-            for i in range(self.npartitions)
-        }
+    def _block_subgraph(self):
+        return {self._name: (operator.getitem, self.frame._name, self.columns)}
 
     def __str__(self):
         base = str(self.frame)
@@ -447,11 +460,8 @@ class ProjectIndex(Elemwise):
     def _meta(self):
         return self.frame._meta.index
 
-    def _layer(self):
-        return {
-            (self._name, i): (getattr, (self.frame._name, i), "index")
-            for i in range(self.npartitions)
-        }
+    def _block_subgraph(self):
+        return {self._name: (getattr, self.frame._name, "index")}
 
 
 class Head(Expr):
@@ -475,14 +485,13 @@ class Binop(Elemwise):
     _parameters = ["left", "right"]
     arity = Arity.binary
 
-    def _layer(self):
+    def _block_subgraph(self):
         return {
-            (self._name, i): (
+            self._name: (
                 self.operation,
-                (self.left._name, i) if isinstance(self.left, Expr) else self.left,
-                (self.right._name, i) if isinstance(self.right, Expr) else self.right,
+                self.left._name if isinstance(self.left, Expr) else self.left,
+                self.right._name if isinstance(self.right, Expr) else self.right,
             )
-            for i in range(self.npartitions)
         }
 
     def __str__(self):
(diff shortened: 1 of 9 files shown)


mrocklin:
Oh fun.  I look forward to reviewing this.  My afternoon is somewhat packed, but hopefully tomorrow morning (or this afternoon if I find some time)

dask_match/tests/test_collection.py
```diff
@@ -139,12 +139,34 @@ def test_repr(ddf):
...
+    pdf = pd.DataFrame({"x": range(10), "y": range(10)})
+    ddf = from_pandas(pdf, npartitions=1)
+
+    ddf2 = (ddf["x"] + ddf["y"]) - 1
+    unfused = optimize(ddf2, fuse=False)
+    fused = optimize(ddf2, fuse=True)
+
+    # Should only get one task per partition
+    assert len(fused.dask) == ddf.npartitions
+    assert_eq(fused, unfused)
```
rjzamora, line 158: @mrocklin - This test checks that we can handle the case you described here: https://github.com/mrocklin/dask-match/pull/11#issuecomment-1497413475

dask_match/tests/test_collection.py
```diff
@@ -139,12 +139,34 @@ def test_repr(ddf):
...
-    result = optimize(ddf[ddf.x > 5].y)
+    result = optimize(ddf[ddf.x > 5].y, fuse=False)
     expected = ddf.y[ddf.x > 5]
 
     assert str(result) == str(expected)
 
 
+def test_optimize_fusion():
+    pdf = pd.DataFrame({"x": range(10), "y": range(10)})
+    ddf = from_pandas(pdf, npartitions=1)
```
mrocklin, line 150: These have fixtures now btw I think

dask_match/tests/test_collection.py
mrocklin, line 158: Nice.  That gives good confidence.

dask_match/tests/test_collection.py
```diff
@@ -139,12 +139,34 @@ def test_repr(ddf):
...
+    assert_eq(fused, unfused)
+
+    # Check that we still get fusion when
+    # non-blockwise operations are added
+    ddf3 = ddf2.sum()
+    unfused = optimize(ddf3, fuse=False)
+    fused = optimize(ddf3, fuse=True)
+
+    assert len(fused.dask) < len(unfused.dask)
+    assert_eq(fused, unfused)
```
mrocklin, line 167: Bonus points if we can demonstrate fusing both before and after non-fusable sections.  Maybe we don't have many operations like this though.

dask_match/core.py
```diff
@@ -616,7 +625,245 @@ def optimize_expr(expr):
...
+        new_operands.append(val)
+
+    if new_count:
+        # Only return new object if something changed
+        return type(expr)(*new_operands)
+    return expr
+
+
+def _blockwise_fusion(expr):
+    """Traverse the expression graph and apply fusion"""
```
mrocklin, line 679: Is this doing diamond-y things?

mrocklin (commented):
Just some questions for now

dask_match/tests/test_collection.py
mrocklin: I see!  🎉

dask_match/tests/test_collection.py
rjzamora: Yeah, sorry. This is leftover from me debugging with one partition.

dask_match/core.py
rjzamora: Yes (or at least that is my intention). I'd like to clean things up a bit, but the basic pseudo-code is:

- Start with an un-optimized `expr`
- (**Outer loop**) While not done:
  - Find the "dependents" of every `Fusable` node in `expr` (`dependents`)
  - Use `dependents` to construct a list of possible fusion roots (`roots`). Note that a "root" is a `Fusable` node in `expr` that has no `Fusable` dependents.
  - (**Inner loop**) For each `root` in `roots`:
    - Pop `root` from `roots` list 
    - Starting at `root`, traverse the tree of `Fusable` dependencies until we come to a node that is not eligible for fusion into the local group. We only allow a node to be fused into the group if it has no dependents outside of the known group. Since we traverse the graph in a breadth-first manner, we should be accounting for typical "diamond" dependency patterns.
    - Once we have found a node that cannot be fused into the current group, we have finished constructing the current `group`. If the ineligible node has remaining `Fusable` dependencies, we append it to the list of possible `roots`.
    - If `group` only contains `root`, we ignore the group and continue the loop at the next `root`
    - If `group` contains multiple nodes, we replace the `root` node of `expr` with a `FusedExpr` object (wrapping the original group)
    - If `roots` is empty, we are done fusing, and can break from the inner **and** outer loop
    - If `roots` is not empty, we need to run through the outer loop again, and only break from the **inner** loop
- Return `expr`, which is now fused.

dask_match/tests/test_collection.py
```diff
@@ -139,12 +139,38 @@ def test_repr(ddf):
...
+
+    # Check that we still get fusion
+    # after a non-blockwise operation as well
+    fused_2 = optimize(ddf3 + 10 - 5, fuse=True)
```
wence-, line 169: maybe explicitly parenthesise `(ddf3 + 10) - 5` so people don't have to remember associativity rules.

dask_match/core.py
```diff
@@ -307,20 +313,32 @@ def _divisions(self):
     def _name(self):
         return funcname(self.operation) + "-" + tokenize(*self.operands)
 
-    def _layer(self):
+    def _subgraph_dependencies(self):
+        # List `Expr` operands only.
+        # This property is required to enable fusion
+        return [operand for operand in self.operands if isinstance(operand, Expr)]
```
mrocklin, line 319: Maybe we should rename this `depencencies` and move it up to the `Expr`level?

dask_match/core.py
```diff
@@ -307,20 +313,32 @@ def _divisions(self):
...
+
+    def _block_subgraph(self):
```
mrocklin, line 321: Slight preference to name this either `_blockwise_subgraph` or `_subgraph` to either reuse the existing term (blockwise) or keep things short.

dask_match/core.py
```diff
@@ -616,7 +625,245 @@ def optimize_expr(expr):
...
+    """
+    if not nodes_to_replace:
+        # Nothing to replace
+        return expr
+    elif expr in nodes_to_replace:
+        # Already at targetted expr
+        return nodes_to_replace[expr]
+
+    new_operands = []
+    new_count = 0
```
mrocklin, line 656: Nit, but maybe make this a bool?

dask_match/core.py
```diff
@@ -616,7 +625,245 @@ def optimize_expr(expr):
...
     return expr
 
 
 from dask_match.reductions import Count, Max, Min, Mode, Size, Sum
+
+
+## Utilites for Expr fusion
+
+
+def replace_nodes(expr, nodes_to_replace: dict):
```
mrocklin, line 642: I propose making this a method on expr, and naming it `substitute`. 

You might find it interesting to look at sympy's API for the same operation: https://docs.sympy.org/latest/tutorials/intro-tutorial/basic_operations.html#substitution

dask_match/core.py
```diff
@@ -616,7 +625,245 @@ def optimize_expr(expr):
...
+        # Nothing to replace
+        return expr
+    elif expr in nodes_to_replace:
+        # Already at targetted expr
+        return nodes_to_replace[expr]
+
+    new_operands = []
+    new_count = 0
+    for operand in expr.operands:
+        if isinstance(operand, Expr) and operand in nodes_to_replace:
```
mrocklin, line 658: I suggest that we support substituting non-exprs as well

dask_match/core.py
```diff
@@ -616,7 +625,245 @@ def optimize_expr(expr):
...
+        # Return original expr if no fusable sub-groups were found
+        return expr, True
+
+    while True:
+        original_name = expr._name
+        expr, done = _fusion_pass(expr)
+        if done or expr._name == original_name:
+            break
+
+    return expr
```
mrocklin, line 762: I'm not reviewing this logic, but I think that it would be good to think of and write down some more complex test cases.

dask_match/core.py
```diff
@@ -616,7 +625,245 @@ def optimize_expr(expr):
...
+        """Return the subgraph for an abstract 'block'.
+        Used to initialize a `SubgraphCallabe` object.
+        """
+        raise NotImplementedError
+
+
+class FusedExpr(Expr):
+    @property
+    def exprs(self):
+        return self.operands[0]
```
mrocklin, line 790: You're doing this because it's hard to define `_parameters` here I'm guessing?

dask_match/core.py
```diff
@@ -307,20 +313,32 @@ def _divisions(self):
...
+                    operand._name if isinstance(operand, Expr) else operand
                     for operand in self.operands
                 ],
                 self._kwargs,
             )
-            for i in range(self.npartitions)
         }
 
+    def _layer(self):
+        return _subgraph_callable_layer(
```
mrocklin, line 335: Maybe this method could be inlined here?  I think that we only need to use this logic in this one place.

dask_match/core.py
```diff
@@ -616,7 +625,245 @@ def optimize_expr(expr):
...
+                block[k] = v
+        return block
+
+    def _layer(self):
+        return _subgraph_callable_layer(
+            self._name,
+            self._block_subgraph(),
+            self._subgraph_dependencies(),
+            self.npartitions,
+        )
```
mrocklin, line 823: I'm surprised that this class doesn't inherit from `Blockwise`.  If it did then this method would inherit nicely from the `Blockwise` superclass.

dask_match/core.py
```diff
@@ -616,7 +625,245 @@ def optimize_expr(expr):
...
+
+
+class MappedArg:
```
mrocklin, line 826: The term `MappedArg` here is strange to me.  Maybe `Indexable` or something else?   Thoughts on what feels good here?

dask_match/tests/test_collection.py
```diff
@@ -139,12 +139,38 @@ def test_repr(ddf):
...
 
 def test_columns_traverse_filters(df, ddf):
-    result = optimize(ddf[ddf.x > 5].y)
+    result = optimize(ddf[ddf.x > 5].y, fuse=False)
     expected = ddf.y[ddf.x > 5]
 
     assert str(result) == str(expected)
 
 
+def test_optimize_fusion(ddf):
```
mrocklin, line 148: Given the heavy logic in the fusion code my guess is that we'll want to think defensibly here a bit, and come up with some cases that might trip up a lesser algorithm.  I wouldn't be surprised, if in doing so, we uncover some failure cases of the code as it is written today.

mrocklin (commented):
Thanks @rjzamora !  Some more comments after having had a chance to look at the code more deeply.

dask_match/core.py
wence-: The existance of non-Expr objects in the operands is an anti-pattern since it means the language isn't closed. So I think it would rather make sense to avoid the case that non-Expr things can show up here, unless that is really crucial. It makes writing generic type-safe traversals and rewrites very hard.

mrocklin: In general I agree with you, but in the specific case of lightweight languages embedded in other languages (like Python) I think that being impure ends up working better.   To make this concrete I'll ask the question:

*What would you recommend for operands like filenames and numbers?*

If you suggest that we make a `String` type and an `Int` type then I'll suggest that that is likely more heavyweight than what I personally would like to see.

dask_match/core.py
rjzamora: I was hesitant to in-line this because we are also using it for `Fusable` expressions that don't inherit from `Blockwise` (i.e. `ReadParquet` and `FromPandas`). I realize we can make `IO` a `Blockwise` subclass, but that will require a bit more work in` ReadCSV`.  Also, I wasn't sure if we actually **want** to require all "fuseable" expressions to be `Blockwise` subclasses, though it is certainly fine with me if we do.

dask_match/core.py
rjzamora: Sorry, this was a debugging leftover. This can be replaced with `_parameters = ["exprs"]`

dask_match/core.py
```diff
@@ -616,7 +625,245 @@ def optimize_expr(expr):
...
+    def __str__(self):
+        descr = "-".join([expr._name.split("-")[0] for expr in self.exprs])
+        return f"Fused-{descr}"
+
+    @property
+    def _name(self):
+        return f"{str(self)}-{tokenize(self.exprs)}"
+
+    def _divisions(self):
+        return self.exprs[0]._divisions()
```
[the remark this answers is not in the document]
rjzamora: Yes exactly. After fusion, the output of this Expr should be the same as the "root" Expr in the fused group (i.e. the top-most expression).

dask_match/core.py
rjzamora: Decided it was worthwhile to make everything "fusable" be a `Blockwise` subclass, so this is now inlined.

dask_match/core.py
rjzamora: Moved this to `Blockwise` now.

dask_match/core.py
rjzamora: Agree that this feels off. I changed the name to `IndexableArg` for now, but I'm still a bit unhappy with the new name. I'd like to communicate that this object is something that will be eagerly indexed at graph-creation time.

dask_match/tests/test_collection.py
rjzamora: Absolutely agree. Once we are happy with the general organization of things, we will want very robust tests to help iterate on this algorithm.

dask_match/core.py
rjzamora: Agree that this logic belongs in `Expr` - I moved over the existing API, but will probably try to make things look more like the sympy API.

mrocklin:
I hope you don't mind I pushed a minor change (hopefully good)

dask_match/core.py
```diff
@@ -616,7 +667,156 @@ def optimize_expr(expr):
...
+        original_name = expr._name
+        expr, done = _fusion_pass(expr)
+        if done or expr._name == original_name:
+            break
+
+    return expr
+
+
+class FusedExpr(Blockwise):
+    _parameters = ["exprs"]
```
mrocklin, line 771: Can I ask for more information here on what this class contains?  I'm curious about what `exprs` looks like.  I'm also curious about the other operands here.  

I suspect that I could learn this by looking more deeply at the fusion logic, but I'm probably not going to end up doing that.

dask_match/tests/test_collection.py
mrocklin: My hope is that the tests will be more robust than the code itself.  I would probably do this pretty early.  (There are some who would do this before writing any code really).  Hopefully these tests are just a bunch of situations that we think might get tricky.  They wouldn't depend on the actual implementation that we end up coming up with.

dask_match/core.py
rjzamora: I added a docstring to (hopefully) clarify things a bit.

rjzamora: The idea is that `exprs` is just a list of `Blockwise` objects that the `_blockwise_fusion` algorithm has decided we can "fuse" together into a single `Expr` node. In order for the `FusedExpr` object to behave as a proper `Expr` node, we also need to make sure that its external dependencies are still reflected in its operands (which is why `exprs` is not the **only** operand).

dask_match/tests/test_collection.py
mrocklin: > Turns out that it is pretty easy to break things outside of fusion while trying to "stress" fusion.

Can you say more here?  In what kinds of situations does this break?

rjzamora: One of the things that "broke" was just the fact that `Series.name` was missing (trivial to fix).  However, I think the main challenge I'm running into is the fact that there are not many non-`Blockwise` operations to work with, and I'd like to be able to do something like `ddf1["a"] + ddf2["b"].sum()` to introduce a non-`Blockwise` `Expr` dependency to a `Blockwise` expression. Without being able to break up `Blockwise` chains like this, it is a bit challenging to write robust tests.

rjzamora: Also, I believe if `optimize` was called within `compute` by default (even without fusion), some of the existing tests would be failing. I will probably try to find a way to test this hypothesis in a systematic way.

rjzamora: Yes, that seems reasonable to me.

Regarding my earlier point about scalar-reduction results: It may be necessary for an `Expr` to "know" if it corresponds to a "scalar" result. Although we want most DataFrame/Series/Index-specific logic to live at the collection level, it seems important for the `Expr` to know when `divisions` are irrelevant. For example, if we had a `Expr.scalar` attribute that only returned `True` for expressions that we know will produce scalar values (using `Reduction._meta` for example), it would be relatively easy to broadcast the key for the scalar result to all partitions of a dependent `Blockwise` operation.

mrocklin: > It may be necessary for an Expr to "know" if it corresponds to a "scalar" result

If we're looking for `is_scalar` then maybe we can infer this from `_meta`?

```python
def is_scalar(self):
    return is_scalar(self._meta)
```

Alternatively, maybe the thing we're looking for here isn't necessarily `is_scalar` but rather `is_broadcastable`.  For example in this case:

```python
df - df.mean()
```

The `df.mean()` value is likely a pandas.Series.  If that's what we're looking for then my suggestion above probably doesn't work.  In that case I'd agree that we probably want something like the following (which maybe you were suggesting above):

```python
class Expr:
    broadcastable = False

class Reduction(Expr):
    broadcastable = True
```

Are there other situations where we'll have broadcastable objects, or where Reductions aren't broadcastable?  I guess maybe there's the more general ACA case.  Maybe some of them will or won't be?

rjzamora: I *think* you are right that I mostly care about whether or not the result is *broadcastable*. However, my understanding is that the only case that a result is "always" broadcastable is if it is also scalar, otherwise it sometimes depends on the specific operation (I may have this wrong).

>If we're looking for is_scalar then maybe we can infer this from _meta?

Yes, we can infer this from the meta, but my earlier profiling suggest that we should avoid materializing the meta when we don't **need** it. Therefore, I'd probably want to return `False` be default, and only add meta-inspection logic to the `Expr` classes that **might** be returning something that is broadcastable. Therefore, I'm pretty much proposing what yourr last code example, but with `Reduction.broadcastable` implemented as a property with simple logic to inspect `meta`.

>Are there other situations where we'll have broadcastable objects

Not sure if there are any other high-profile DataFrame operations that are broadcastable - good question.

>or where Reductions aren't broadcastable?

It depends on the collection you are broadcasting over, right?  I understand that you can broadcast a series over a DataFrame collection, but can you broadcast a series over a series collection? If we support split_out>1, there will certainly be general ACA results that can't be broadcasted.

dask_match/core.py
wence-: I think it makes sense to have a `Literal(value)` type (or if you will `Opaque(...)`) that doesn't have any operands. If doing things like replacement you still need to match on the object and dispatch based on the contents of `value`, but you no longer have to special-case it if you don't care.

IME, while it is tempting to avoid recreating types that already exist, this is a decision that typically becomes painful down the line, since every traversal and rewrite algorithm that is subsequently written has to remember that not every operand of every expr is an expr. This can then only be handled by isinstance checks which is fragile because you have an open untagged union of potential types (so it's easy to write `isinstance(expr, (str, int))` and then later for someone to shove a float in, or whatever). At least with the `Literal` object you control the surface area of this kind of checking.

That's not to say that one can't add sugar to the constructors such that literal python types are not wrapped up, but I think the internal representation should have an Expr object for them.

An alternative approach is to store python literals as "non-child" data, but that comes with drawbacks too if you want to rewrite there.

dask_match/tests/test_collection.py
mrocklin: > but my earlier profiling suggest that we should avoid materializing the meta when we don't need it. 

I'm curious to learn more about this. We used to eagerly compute meta and it wasn't a problem.  My guess is that there is a way to make this work.  Probably this means tracking which expressions have expensive meta calculations (my guess is that this is only read_parquet) and then making sure that we're caching things appropriately.

 > Therefore, I'm pretty much proposing your last code example, but with Reduction.broadcastable implemented as a property with simple logic to inspect meta.

If the protocol generally is to have a property then I'm fine with that.  My guess is that Reduction is always true though.   
 If we're doing something like dask.Series + pandas.Series then that should fail some other check, other than broadcastable is my guess.  Maybe ACA is sometimes true and needs a proper property due to, for example, split_out.

mrocklin:
@rjzamora how best can I help on this PR?  Are you waiting for me on any particular decision or do you have a clear path to proceed?

mrocklin:
I hope you don't mind, but I cleaned up substitute a little, added tests for it, and pushed.

dask_match/tests/test_collection.py
```diff
@@ -139,12 +139,85 @@ def test_repr(ddf):
...
+
+    # Check that we still get fusion
+    # after a non-blockwise operation as well
+    fused_2 = optimize((ddf3 + 10) - 5, fuse=True)
+    # The "+10 and -5" ops should get fused
+    assert len(fused_2.dask) == len(fused.dask) + 1
+
+
+@pytest.mark.parametrize("persist", [True, False])
+def test_optimize_fusion_many(ddf, persist):
```
mrocklin, line 174: These tests seem to take far longer than others. While we're this early in the project I'd love to keep tests snappy. I think that that will keep development fast.  Do you have thoughts on why these are slow?

mrocklin:
Type checks are pretty cheap.  I'm curious what the motivation is to avoid type checks.

mrocklin:
Going a different direction, here are the things that are stopping me from pressing the green button.  This is what I would probably prioritize in order to get this in:

1. Make sure that we have good tests for this functionality
2. Make sure that tests are fast
3. Consider finding a new name for `_blockwise_subgraph` (but maybe keep it if we think it's the best we can find)
4. Make sure that we have docstrings for things (example IndexableArg doesn't describe its parameters)

Do you have other thoughts for larger concerns?

mrocklin:
> I'm referring to code complexity

Or maybe I'm missing something.  Is there serious code complexity currently due to this?  Types aren't currently hitting  warning bells for me in the same way that slow tests, test coverage, and docstrings are.

rjzamora:
>This is the case today I think?

It is at the `Expr` level, but the IO methods then include `IndexableArg` objects in their `dependencies` output. I'm just saying that there is no real reason that `IndexableArg` couldn't be a lightweight `Expr` subclass.

>Make sure that we have good tests for this functionality

Do you think we can do this without having "broadcastable" operations to break fusion into multiple groups?

>Make sure that tests are fast

If the problem is just in `assert_eq`, do we just use smaller collections and fewer expressions?

mrocklin:
> These are dataframes with ~100 rows. It's hard for me to believe that computation is what's taking up time. It's got to be something in our code. Maybe I'm wrong though? Have you profiled here? Saying that assert_eq is slow could mean, for example, that graph generation is slow or meta generation is slow or many other things that are still in this code base

I'm profiling now. It's definitely our fault.  One sec, I'll see if I can push up a change.  

For context on how I identified this, I like using snakeviz, so I removed parameters from the test

```diff
diff --git a/dask_match/tests/test_collection.py b/dask_match/tests/test_collection.py
index b663cf4..0095766 100644
--- a/dask_match/tests/test_collection.py
+++ b/dask_match/tests/test_collection.py
@@ -170,8 +170,11 @@ def test_optimize_fusion(ddf):
     assert len(fused_2.dask) == len(fused.dask) + 1


-@pytest.mark.parametrize("persist", [True, False])
-def test_optimize_fusion_many(ddf, persist):
+def test_optimize_fusion_many():
+    persist = False
+    df = pd.DataFrame({"x": range(100)})
+    df["y"] = df.x * 10.0
+    ddf = from_pandas(df, 10)
```

Then I install snakeviz

```
mamba install snakeviz
```

Then I did the following in ipython

```python
from dask_match.tests.test_collection import test_optimize_fusion_many
test_optimize_fusion_many()
%load_ext snakeviz
snakeviz test_optimize_fusion_many()
```

That gave me this view:

Screen Shot 2023-04-11 at 12 16 17 PM

I'll go throw cached_property around on a few methods and see what happens.

rjzamora:
Yeah, my local branch has a few more `cached_property` decorators - Things are certainly faster if you cache the _layer result.

mrocklin:
With my recent commit we cache only small things (`_meta`, `_name`) and I've got tests down to about 1.5s

mrocklin:
profiling shows that the rest of the time is spent in the dask local scheduler, and matchpy (about half-half)

mrocklin:
Snakeviz screenshot

Screen Shot 2023-04-11 at 12 28 02 PM

There's still room for improvement, but only about 20-30% of total I think.

rjzamora:
Okay - I added in basic broadcasting support for now. I will shift focus on making the necessary changes for this to be merged:

- Improve tests
- "Make sure that tests are fast" (probably need to re-profile)
- "Consider finding a new name for _blockwise_subgraph" - I don't personally care much about this name. Feel free to suggest any name you like.
- Add/improve missing docstrings

mrocklin:
Pushed a couple more:

1.  I noticed `None` names in the graph, I think from from_pandas
2. I added a basic broadcast test.   It seems like this is important to us, is potentially tricky, and is an easy thing to test in isolation

dask_match/tests/test_collection.py
```diff
@@ -139,12 +139,97 @@ def test_repr(ddf):
...
+    )
+    assert_eq(
+        ddf.x + ddf.x.sum(),
+        df.x + df.x.sum(),
+    )
+
+
+def test_optimize_fusion_broadcast(ddf):
+    # Check fusion with broadcated reduction
+    ser = ((ddf["x"] + 1) + ddf["y"].sum()) + 1
```
mrocklin, line 227: I'm curious, what does `ser` stand for?

rjzamora: Sorry - should be called `result` - Comes from an earlier iteration with both dataframe and series results.

mrocklin: I can do that later though too

rjzamora: Thanks for the active engagement! You are welcome to go crazy - Need to pause for child pickup anyway :)

dask_match/tests/test_collection.py
```diff
@@ -139,12 +139,97 @@ def test_repr(ddf):
...
 
 
+def test_optimize_fusion(ddf):
+    ddf2 = (ddf["x"] + ddf["y"]) - 1
+    unfused = optimize(ddf2, fuse=False)
+    fused = optimize(ddf2, fuse=True)
+
+    # Should only get one task per partition
+    assert len(fused.dask) == ddf.npartitions
+    assert_eq(fused, unfused)
```
mrocklin, line 155: calling `assert_eq` will call optimize with the default `fuse=True`, correct?  If so, are these `assert_eq(fused, unfused)` doing what we expect?

rjzamora: I don’t think we configured compute to call our optimize function yet (I think that’s what you’re asking)

mrocklin:
I've cleaned them up (at least according to my aesthetics) and pushed to a new file

dask_match/tests/test_collection.py
mrocklin: Ah, good point.  Thanks.

mrocklin:
I've renamed `_blockwise_subgraph` to `_blockwise_layer`.  I don't love this name either, but it at least reuses the same term "layer" that is used elsewhere in the codebase.  Hopefully that helps newcomers quickly identify the structure of what that method does.

dask_match/core.py
```diff
@@ -290,38 +345,114 @@ def _meta(self):
...
         return {}
 
+    def _broadcast_dep(self, dep: Expr):
+        # Checks if a dependency should be broadcasted to
+        # all partitions of this `Blockwise` operation
+        return (
+            not isinstance(dep, BlockwiseArg)
+            and dep.npartitions == 1
+            and dep.ndim < self.ndim
+        )
```
mrocklin, line 355: I'm curious, why this relationship rather than `dep.is_broadcastable`?

What would happen in the case where self.ndim is also equal to one?  Would it even matter?

mrocklin: My intuition is that making an `is_broadcastable` property will end up being smoother.  I don't have great evidence for that though, just intuition.

mrocklin: Oh, this is the DataFrame + Series thing?

rjzamora: This could be wrong? I assumed that you cannot “broadcast” a series over a series or a dataframe over a dataframe. If both sides have the same dimension you would need to align divisions rather than broadcast.

mrocklin: Yeah, ok, nm, that makes sense.

mrocklin:
Merging later today if there are no further comments

rjzamora:
>I've renamed _blockwise_subgraph to _blockwise_layer

Seems like a reasonable name to me.

>Merging later today if there are no further comments

Sounds good. Thanks for cleaning things up!

PR #16, opened 2023-04-06
Use `ProjectPartitions` to enable high-level culling

rjzamora:
This PR implements `ProjectPartitions` and uses it within `head` to avoid the need for low-level culling. This was originally an experiment meant to explore @mrocklin's thoughts in [this comment](https://github.com/mrocklin/dask-match/pull/11#issuecomment-1494884001).

I do think this work demonstrates that low-level culling is probably unnecessary.  That is, It seems relatively easy to push down `ProjectPartitions` through `Blockwise` expressions, and then only non-`Blockwise` expressions need to support something like `output_partitions_rule` to avoid the materialization of unnecessary tasks.

### DIFF
--- a/dask_match/core.py
+++ b/dask_match/core.py
@@ -280,6 +280,11 @@ class Blockwise(Expr):
 
     operation = None
 
+    @classmethod
+    def _replacement_rules(cls):
+        # Push down partition projection
+        yield partition_pushdown_rule(cls)
+
     @property
     def _meta(self):
         return self.operation(
@@ -388,7 +393,7 @@ class Filter(Blockwise):
     operation = operator.getitem
 
     @classmethod
-    def _replacement_rules(self):
+    def _replacement_rules(cls):
         df = Wildcard.dot("df")
         condition = Wildcard.dot("condition")
         columns = Wildcard.dot("columns")
@@ -399,6 +404,9 @@ class Filter(Blockwise):
             Pattern(Filter(df, condition)[columns]),
             lambda df, condition, columns: df[columns][condition],
         )
+
+        # Push down partition projection
+        yield partition_pushdown_rule(cls)
 
 
 class Projection(Elemwise):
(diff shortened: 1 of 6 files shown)


dask_match/core.py
```diff
@@ -454,6 +462,61 @@ def _layer(self):
...
+
+    @property
+    def _meta(self):
+        return self.frame._meta
+
+    @property
+    def partitions(self):
+        if isinstance(self.projection, (list, tuple)):
+            return list(self.projection)
+        raise TypeError
```
mrocklin, line 478: I think in general we should assume types in Expr code, and ask for user-facing API code to coerce/normalize things.  Thoughts?  

(this comment goes beyond this PR, and is more about general best practices in this repo)

dask_match/core.py
```diff
@@ -454,6 +462,61 @@ def _layer(self):
         }
 
 
+class ProjectPartitions(Expr):
```
mrocklin, line 465: The term `Project` may be misused here.  

My understanding is that `Projection` is a term in relational algebra that is a fancy way of saying "select some columns".  I don't think that it gets applied to rows in general. 

Maybe `FilterPartitions` would be better.

mrocklin: Or at least more in line with the nomenclature of that field.

dask_match/core.py
```diff
@@ -524,6 +590,9 @@ def _replacement_rules(cls):
 
         yield from super()._replacement_rules()
 
+        # Push down partition projection
+        yield partition_pushdown_rule(cls)
```
mrocklin, line 594: Could this be done in a superclass somehow, like `Blockwise`?

dask_match/io/io.py
```diff
@@ -10,23 +11,36 @@ class IO(Expr):
...
     def _meta(self):
         return self.frame.head(0)
 
     def _divisions(self):
-        return [None] * (self.npartitions + 1)
+        return [None] * (len(self._chunks) + 1)
 
-    def _layer(self):
+    @cached_property
+    def _chunks(self):
```
mrocklin, line 30: This name clashes a bit with Dask Array.  Maybe `._partitions` instead?

mrocklin (commented):
A few comments.  I haven't yet given this topic a ton of thought.  

I think that it definitely makes sense to push down head/tail/... operations through blockwises (and whatever else).  

I don't yet have strong confidence on what's best when it comes to having IO/Shuffle operations know about desired outputs, or if there should be a head/tail operation just afterwards.  Still thinking.

mrocklin:
Another thought is that we don't always need to use replacement rules.  We can also traverse the tree ourselves at any point if we think that that's easier.

dask_match/core.py
mrocklin: Also, with the current rule as written, does the head/tail operation get pushed down into each side of the `Add/Sub/...`

PR #19, opened 2023-04-11
Support Pandas 2.0

mrocklin:


mrocklin:
@rjzamora can you help me with this?  It looks like the `use_nullable_dtypes` code has gone stale upstream and we have some drift.

mrocklin:
You should have write access now I hope

rjzamora:
>It looks like the use_nullable_dtypes code has gone stale upstream and we have some drift.

Ah - That makes sense. I can probably update the parquet code pretty easily.

PR #20, opened 2023-04-12
Rename fixtures from ddf->df and df->pdf

mrocklin:
We're going to use the dask variante more.  Let's keep it small.


rjzamora (approved):
Yeah - I'm a fan of using `df` over `ddf`

mrocklin:
Want to do the honors and press the green button?

mrocklin:
Sorry, got impatient :)

PR #21, opened 2023-04-12
Add basic simplification traversal

mrocklin:
This adds a basic class-based traversal for optimization.  This has some pros/cons over rewrite rules

**Pro:** 

-  it's probably simpler for new folks to understand
- it understands inheritance

**Con:**

-  it may become hard when we have different kinds of classes interacting with each other.  It becomes hard to think of all-pairs interactions

For comparison's sake I added the same optimization twice in each form, `df[a][b] -> df[]b]`

This is also in the theme of "let's have a few different kinds of optimization going on" 

@rjzamora any thoughts are welcome

### DIFF
--- a/dask_match/expr.py
+++ b/dask_match/expr.py
@@ -128,6 +128,9 @@ class Expr(Operation, metaclass=_ExprMeta):
         # Dependencies are `Expr` operands only
         return [operand for operand in self.operands if isinstance(operand, Expr)]
 
+    def simplify(self):
+        return self
+
     @property
     def index(self):
         return ProjectIndex(self)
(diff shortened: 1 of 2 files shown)


dask_match/expr.py
```diff
@@ -590,6 +630,16 @@ def _layer(self):
             (self._name, 0): (M.head, (self.frame._name, 0), self.n),
         }
 
+    def simplify(self):
+        if isinstance(self.frame, Elemwise):
+            operands = [
+                Head(op, self.n) if isinstance(op, Expr) else op
+                for op in self.frame.operands
+            ]
+            return type(self.frame)(*operands)
```
mrocklin, line 639: @rjzamora I was thinking of your culling PR when writing this.  Maybe something like this would end up being easier?  I'm not sure.

rjzamora: I think I agree that the `simplify` approach is more intuitive than a general replacement rule when you know that you will always want to push a particular expression down the tree as far as possible. I'll think a bit more about this, but I think I like the general idea.

PR #22, opened 2023-04-12
Rename ProjectIndex to Index

mrocklin:


mrocklin:
I get that this was likely named as it was to avoid a name collision with the eventual user-facing class.  Now that we're namespaced maybe we don't need this.

rjzamora (approved):
No objections from me.

PR #23, opened 2023-04-12
Add shuffle algorithms

rjzamora:
Adds `shuffle` API and very-simple shuffle algorithm to make it work. ~Also adds a `partitions` API to make it easier to test the shuffled result.~

### DIFF
--- a/dask_match/expr.py
+++ b/dask_match/expr.py
@@ -205,20 +205,20 @@ class Expr(Operation, metaclass=_ExprMeta):
         else:
             return NE(other, self)
 
-    def sum(self, skipna=True, level=None, numeric_only=None, min_count=0):
-        return Sum(self, skipna, level, numeric_only, min_count)
+    def sum(self, skipna=True, numeric_only=None, min_count=0):
+        return Sum(self, skipna, numeric_only, min_count)
 
-    def mean(self, skipna=True, level=None, numeric_only=None, min_count=0):
+    def mean(self, skipna=True, numeric_only=None, min_count=0):
         return self.sum(skipna=skipna) / self.count()
 
-    def max(self, skipna=True, level=None, numeric_only=None, min_count=0):
-        return Max(self, skipna, level, numeric_only, min_count)
+    def max(self, skipna=True, numeric_only=None, min_count=0):
+        return Max(self, skipna, numeric_only, min_count)
 
     def mode(self, dropna=True):
         return Mode(self, dropna=dropna)
 
-    def min(self, skipna=True, level=None, numeric_only=None, min_count=0):
-        return Min(self, skipna, level, numeric_only, min_count)
+    def min(self, skipna=True, numeric_only=None, min_count=0):
+        return Min(self, skipna, numeric_only, min_count)
 
     def count(self, numeric_only=None):
         return Count(self, numeric_only)
--- a/dask_match/expr.py
+++ b/dask_match/expr.py
@@ -10,6 +10,7 @@ import toolz
 from dask.base import normalize_token, tokenize
 from dask.core import ishashable
 from dask.dataframe import methods
+from dask.dataframe.core import is_dataframe_like
 from dask.utils import M, apply, funcname
 from matchpy import (
     Arity,
@@ -117,6 +118,8 @@ class Expr(Operation, metaclass=_ExprMeta):
             if key in _parameters:
                 idx = _parameters.index(key)
                 return self.operands[idx]
+            if is_dataframe_like(self._meta) and key in self._meta.columns:
+                return self[key]
             raise err
 
     def operand(self, key):
--- a/dask_match/expr.py
+++ b/dask_match/expr.py
@@ -11,7 +11,12 @@ import toolz
 from dask.base import normalize_token, tokenize
 from dask.core import ishashable
 from dask.dataframe import methods
-from dask.dataframe.core import is_dataframe_like
+from dask.dataframe.core import (
+    is_dataframe_like,
+    apply_and_enforce,
+    _get_meta_map_partitions,
+    _get_divisions_map_partitions,
+)
 from dask.utils import M, apply, funcname
 from matchpy import (
     Arity,
@@ -25,6 +30,8 @@ from matchpy import (
 from matchpy.expressions.expressions import _OperationMeta
 
 replacement_rules = []
+
+no_default = "__no_default__"
 
 
 class _ExprMeta(_OperationMeta):
@@ -498,6 +505,61 @@ class Blockwise(Expr):
             return (apply, self.operation, args, self._kwargs)
         else:
             return (self.operation,) + args
+
+
+class MapPartitions(Blockwise):
+    _parameters = [
+        "frame",
+        "func",
+        "meta",
+        "enforce_metadata",
+        "transform_divisions",
+        "kwargs",
+    ]
+
+    def _broadcast_dep(self, dep: Expr):
+        # Always broadcast single-partition dependencies in MapPartitions
+        return dep.npartitions == 1
+
+    @property
+    def args(self):
+        return [self.frame] + self.operands[len(self._parameters) :]
+
+    @functools.cached_property
+    def _meta(self):
+        meta = self.operand("meta")
+        args = [arg._meta if isinstance(arg, Expr) else arg for arg in self.args]
+        return _get_meta_map_partitions(args, [], self.func, self.kwargs, meta, None)
+
+    def _divisions(self):
+        dfs = [arg for arg in self.args if isinstance(arg, Expr)]
+        return _get_divisions_map_partitions(
+            False,  # Partitions must already be "aligned"
+            self.transform_divisions,
+            dfs,
+            self.func,
+            self.args,
+            self.kwargs,
+        )
+
+    def _task(self, index: int):
+        args = [self._blockwise_arg(op, index) for op in self.args]
+        if self.enforce_metadata:
+            kwargs = self.kwargs.copy()
+            kwargs.update(
+                {
+                    "_func": self.func,
+                    "_meta": self._meta,
+                }
+            )
+            return (apply, apply_and_enforce, args, kwargs)
+        else:
+            return (
+                apply,
+                self.func,
+                args,
+                self.kwargs,
+            )
 
 
 class Elemwise(Blockwise):
--- a/dask_match/expr.py
+++ b/dask_match/expr.py
@@ -397,7 +397,9 @@ class Expr(Operation, metaclass=_ExprMeta):
         Parameters
         ----------
         substitutions:
-            mapping old terms to new terms
+            mapping old terms to new terms. Note that using
+            non-`Expr` keys may produce unexpected results,
+            and substituting boolean values is not allowed.
 
         Examples
         --------
@@ -413,7 +415,11 @@ class Expr(Operation, metaclass=_ExprMeta):
         new = []
         update = False
         for operand in self.operands:
-            if ishashable(operand) and operand in substitutions:
+            if (
+                not isinstance(operand, bool)
+                and ishashable(operand)
+                and operand in substitutions
+            ):
                 new.append(substitutions[operand])
                 update = True
             elif isinstance(operand, Expr):
--- a/dask_match/expr.py
+++ b/dask_match/expr.py
@@ -130,7 +130,7 @@ class Expr(Operation, metaclass=_ExprMeta):
 
     def simplify(self):
         return None
-    
+
     def compose(self):
         return None
 
--- a/dask_match/expr.py
+++ b/dask_match/expr.py
@@ -128,9 +128,12 @@ class Expr(Operation, metaclass=_ExprMeta):
         # Dependencies are `Expr` operands only
         return [operand for operand in self.operands if isinstance(operand, Expr)]
 
+    def simplify(self):
+        return self
+
     @property
     def index(self):
-        return ProjectIndex(self)
+        return Index(self)
 
     @property
     def size(self):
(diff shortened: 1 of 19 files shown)


dask_match/tests/test_collection.py
```diff
@@ -222,3 +222,35 @@ def test_copy(pdf, df):
...
+        pd.DataFrame({"x": list(range(20)) * 5, "y": range(100)}),
+        npartitions=10,
+    )
+    df2 = df.shuffle("x", npartitions=npartitions, ignore_index=ignore_index)
+
+    # Check that the output partition count is correct
+    assert df2.npartitions == (npartitions or df.npartitions)
+
+    # Check the computed (re-ordered) result
+    assert_eq(df, df2.compute(), check_index=not ignore_index)
```
mrocklin, line 248: I'm curious, why the compute call here?

dask_match/expr.py
```diff
@@ -696,6 +696,29 @@ class NE(Binop):
...
+        for part in self.partitions:
+            divisions.append(self.frame.divisions[part])
+        divisions.append(self.frame.divisions[part + 1])
+        return tuple(divisions)
+
+    def _layer(self):
+        return {
+            (self._name, i): (self.frame._name, part)
+            for i, part in enumerate(self.partitions)
+        }
```
mrocklin, line 719: Can I ask for a small test for this?

mrocklin (commented):
Seems cool.  

Any thoughts on how this feels while writing it?  

Any thoughts on what would be hard about moving over the full algorithms?

dask_match/expr.py
rjzamora: Sure - Do you have specific functionality in mind that isn’t covered by test_partitions?

mrocklin: Oh I'm sorry, I did a cursory look and didn't see anything.  Whoops.

dask_match/tests/test_collection.py
rjzamora: I’m just checking that the data is unchanged. assert_eq will sort the result, but will not repartition. Other ideas are very welcome.

mrocklin: > assert_eq will sort the result, but will not repartition

These things seem ok?  Things seem to pass without the compute call.  In general I always try to give assert_eq the full collection so that it can run it's other checks.  I may not fully understand what's going on here though.

dask_match/tests/test_collection.py
```diff
@@ -222,3 +222,35 @@ def test_copy(pdf, df):
...
+
+    # Check that df was really partitioned by "x"
+    unique = []
+    for part in df2.partitions:
+        unique.extend(part["x"].compute().unique().tolist())
```
mrocklin, line 253: This chunk here is fairly expensive.  It's about 500ms.   I'm probably being overly pedantic about test durations, but I'd love to keep things fast for as long as we can.  My guess is that a small change like making one `dask.compute` call rather than several would help avoid this.  

My apologies for what I suspect is perceived to be absurd pickiness here.

dask_match/tests/test_collection.py
rjzamora: Ah fair point - This was originally failing for me at one point for me, but probably for a different reason than I thought.

dask_match/tests/test_collection.py
rjzamora: No worries. The slight delay at `test_simple_shuffle` was already on my radar :)

rjzamora:
>Any thoughts on how this feels while writing it?

Writing an `Expr` is relatively simple (especially before you consider rewrite or pattern-matching optimizations). I think the key task is working out an intuitive API, and deciding how to interface it with a user-friendly collection API. The graph-construction logic itself is almost copy-paste from dask.

>Any thoughts on what would be hard about moving over the full algorithms?

Nothing scary jumps out at me (yet).  I suppose I'm not entirely sure if the alternative algorithms should be dispatched at the collection level or as `BaseShuffle` subclasses (as this PR currently assumes). For example, a collection-level algorithm may wish to compose multiple `Expr` steps (including something like a `Repartition`).

dask_match/tests/test_collection.py
mrocklin: Thank you for humoring my obsession here 🙂

mrocklin:
> For example, a collection-level algorithm may wish to compose multiple Expr steps (including something like a Repartition)

That's an interesting question.  I suspect that some of this is affected by what kinds of optimizations we'll want to support.  Getting fancy, I can also imagine us having expressions that don't actually have `_layer` implementations, but strictly lower themselves down in optimization passes to other expressions that do have `_layer` expressions.  Some `AbstractShuffle` might decide to become a `Repartition` followed by a `P2PShuffle` in the right circumstances, but only after various abstract filter/join optimizations had taken place.

rjzamora:
>Some AbstractShuffle might decide to become a Repartition followed by a P2PShuffle in the right circumstances

Good point. Maybe `shuffle` should always call the same abstract `Shuffle` expression, and that class can just have an optional `engine="auto"` operand that can be chosen at optimization/compute time.

mrocklin:
> and that class can just have an optional engine="auto" operand that can be chosen at optimization/compute time

To be clear, my suggestion above is that rather than have a keyword, it would be optimized into a different class.  Functionally the same thing, but maybe a different way of thinking about different algorithms for different operations.

rjzamora:
>To be clear, my suggestion above is that rather than have a keyword, it would be optimized into a different class.

Right, I do get the general idea. I'm just trying to check if we are starting with a reasonable class organization here (I *think* we are). We can certainly change things later, but I'd like to have a rough idea of what "default" behavior is, and what would need to be done to add a new shuffle algorithm.

If I were to stick to the current design in this PR, the user would be able to specify a specific `Shuffle` subclass with `algorithm=foo`. I suppose `foo` could also be a `callable` object that actually produces something like `P2PShuffle(Repartition(Expr))`, so we don't really need the extra "engine" attribute I floated in my previous comment. This design makes it easy to add/select different shuffle algorithms. However, I also like the idea of using an `AbstractShuffle` default, where the algorithm is chosen automatically at optimization time.

mrocklin:
> I'm just trying to check if we are starting with a reasonable class organization here

Yeah, good point to raise.  Going even more abstract, maybe we should eventually have operations like Join/SetIndex/Sort/GroupbyApply that then get lowered down to Shuffle.  

> the user would be able to specify a specific Shuffle subclass with algorithm=foo.

Sure, that makes sense to me.

mrocklin:
I'm curious to learn more about your thoughts on `compose`.  This is another lowering operation?  

I'm also around most of this afternoon if you want to chat.  It seems like this might be a fun open question to talk through.

rjzamora:
>Going even more abstract, maybe we should eventually have operations like Join/SetIndex/Sort/GroupbyApply that then get lowered down to Shuffle.

Okay - I decided to experiment with this idea a bit here, and I'm totally expecting us to poke holes in the design I have so far :)

I introduced the idea of requiring a `compose` pass (which we may choose to rename to `lower`, or throw out altogether) before a dask graph can be generated.  This way, we can use abstract `Shuffle`/`Join`/etc expressions that can later be "lowered"/"composed" into multiple concrete expressions.  Still thinking this through.

mrocklin:
Or rather, I'll propose that we maybe don't, and that we should do something like the following:

```python
class Shuffle(Expr):
    def simplify(self):
        return SimpleShuffle(...)
```

Then, as we get more algorithms we can choose to make that method a bit smarter by looking at, for example, `self.frame` or `get_client()`

rjzamora:
>Or rather, I'll propose that we maybe don't, and that we should do something like the following:

I'm totally okay with re-using `simplify` to achieve what `compose` is currently trying to do. However, this means we would need to require `simplify` to be called any time we materialized a graph, because I want an abstract `Shuffle` operation to actually map to something like `SimpleShuffle(Assign(Expr, ...,PartitioningIndex(Expr, ...)))` (and I don't want it to be possible for `Shuffle` to materialize a graph. Does that make sense?

rjzamora:
>I notice that you're renaming things to shuffle_or_compose

Yeah, I wasn't expecting to keep things this way. I was just trying to experiment with minimal changes. I was hoping to get your thoughts on **requiring** simplify (or some other form of optimization). 

>This is forcing us to think more deeply about how we require optimize to be called as part of the computation process.

Right - exactly

mrocklin:
I agree that we'll have to iterate between different optimization passes.  `def optimize` will get more complex.  

Right now we have three ways to optimize:

1.  Method based with simplify
2. Replacement rules with matchpy
3. blockwise fusion (an example of a completely external pass)

We're already making arbitraty choices to run them in this order.  That's ok.  We'll end up making other choices in the future.  

I don't want to introduce more abstractions than we need to.  Right now I think that each of these is good, because they show us different ways to write optimizations, and we can figure out what works best in which situations.  I'd be happy if one of these eventually went away.

I don't currently see a need to add another method-based optimization pass.  I think that simplify can handle things today.  It could totally be though that in the future we want to separate which optimizations run at which times.  In that case I could imagine spliting these out into different families.  I'd like to avoid that for now though if we can and deal with it later if it's not getting in the way.

rjzamora:
>I don't currently see a need to add another method-based optimization pass. I think that simplify can handle things today. It could totally be though that in the future we want to separate which optimizations run at which times. In that case I could imagine spliting these out into different families. I'd like to avoid that for now though if we can and deal with it later if it's not getting in the way.

I'm good with this as long as we can start requiring `simplify` to run before a graph can be generated by `Expr`. In the future, I would like to be able to optionally disable "lowering" within `simplify`, so that it becomes possible to prepare an optimized "abstract" graph. In my latest commit, I added a `lower` argument to `optimize`/`simplify` to demonstrate what I want. However, I'll be happy to pull that out for now.

mrocklin:
> I'm good with this as long as we can start requiring simplify to run before a graph can be generated by Expr.

If you haven't already, I recommend looking at https://github.com/mrocklin/dask-match/pull/26/files

> In the future, I would like to be able to optionally disable "lowering" within simplify, so that it becomes possible to prepare an optimized "abstract" graph

I agree in that general we'll want to make optimization fairly configurable.  Many similar systems have to do similar things.  I plan to address this lazily rather than eagerly.

>  my latest commit, I added a lower argument to optimize/simplify to demonstrate what I want. However, I'll be happy to pull that out for now

Yeah, I'd like to avoid using a bunch of keyword arguments for this.  I think that that approach will paint us into a corner.  My guess now is that we'll end up using Dask config until we need something more sophisticated.

rjzamora:
Thanks @mrocklin - Sounds good. Let's get #26 in and that should indeed unblock me here.

mrocklin:
Well, that's good motivation to work on that I guess :)

dask_match/shuffle.py
```diff
@@ -0,0 +1,444 @@
...
+        if isinstance(backend, ShuffleBackend):
+            lower = backend.from_abstract_shuffle
+        elif backend == "simple":
+            lower = SimpleShuffle.from_abstract_shuffle
+        elif backend == "tasks":
+            lower = TaskShuffle.from_abstract_shuffle
+        else:
+            # Only support task-based shuffling for now
+            raise ValueError(f"{backend} not supported")
+        return lower(self)
```
mrocklin, line 62:
```suggestion
        if isinstance(backend, ShuffleBackend):
            return backend.from_abstract_shuffle(self)
        elif backend == "simple":
            return SimpleShuffle.from_abstract_shuffle(self)
        elif backend == "tasks":
            return TaskShuffle.from_abstract_shuffle(self)            
        else:
            # Only support task-based shuffling for now
            raise ValueError(f"{backend} not supported")
```

dask_match/shuffle.py
```diff
@@ -0,0 +1,444 @@
...
+                        self.npartitions,
+                        self.ignore_index,
+                        self.npartitions,
+                    )
+
+        return dsk
+
+
+class TaskShuffle(SimpleShuffle):
+    """Staged task-based shuffle implementation"""
```
mrocklin, line 206: This is the same code roughly from dask.dataframe?

dask_match/repartition.py
```diff
@@ -0,0 +1,53 @@
+import functools
+
+from dask.dataframe import methods
+
+from dask_match.expr import Expr
+
+
+class Repartition(Expr):
+    # TODO: Make this a proper abstract expression
+    pass
```
mrocklin, line 10: I'm glad to see some repartitioning.  This should be, I hope, useful to test out some structures like `Partitions` and culling generally.

If you don't have them already (forgive me if I've missed them) can I ask for a few tests around this functionality?

dask_match/shuffle.py
rjzamora: Yes. Though the subtle differences made it surprisingly painful to copy over :)

mrocklin: I'm curious about this pain if you have a moment to share.  (I'm curious generally about the cost of migration)

rjzamora: It was more-so tedious than (necessarily) painful. I just decided to combine all stages into the same expression (rather than the multiple layers used in HLGs). I made some translation errors in the porting process and struggled to find a small typo for more than an hour :)

In general, I would expect the most challenging part of copying things over to be that it could sometimes be tedious. Sometimes because you are copying from `Layer` logic that is slightly different from what you need, and sometimes because you are copying logic that mixes collection- and graph-level logic. For the latter case, I'm finding the concept of an abstract expression very useful.

dask_match/repartition.py
rjzamora: Yes, I agree with this. I will probably break this into a separate PR today and add test coverage.

rjzamora:
**Update**: Added disk shuffle (pretty trivial). So, only p2p is left. I'll plan to introduce that in a follow-up.

dask_match/shuffle.py
```diff
@@ -0,0 +1,486 @@
...
+
+    _parameters = ["frame", "index", "npartitions_out"]
+    operation = make_partitioning_index
+
+    @property
+    def _meta(self):
+        index = self.operand("index")
+        if isinstance(index, Expr):
+            index = index._meta
+        return make_partitioning_index(self.frame._meta, index, self.npartitions_out)
```
mrocklin, line 474: Was this not handled by `Blockwise._meta`?

mrocklin:
In general I'm fine with what's here.  It seems sensible and well isolated.

dask_match/shuffle.py
rjzamora: Hmm, maybe not (given CI failures)

mrocklin: Not a big deal.  I'm fine with it as it was.  (also fine with anything else here).

rjzamora:
>Just to make my last comment more concrete. Feel free to merge at will.

Thanks - I plan to merge as soon as I have CI passing (unless doing so requires "questionable" changes for some reason).

PR #26, opened 2023-04-13
Always optimize

mrocklin:
This is a work in progress.  

We pull out replacement rules into their own function.  We then add those to `__dask_graph__` (we should move them elsewhere to some external protocol eventually).  

Tests aren't passing though.  We're running into weird things.  Mostly I'm starting to be concerned about the complexity of graphs produced by Blockwise.  For one it looks like we're always using SubgraphCallable, even when it's a very simple graph.  I'm going to play with this for a bit.  cc'ing @rjzamora in case he has ideas here.

### DIFF
--- a/dask_match/expr.py
+++ b/dask_match/expr.py
@@ -802,21 +802,27 @@ def optimize(expr: Expr, fuse: bool = True) -> Expr:
     matchpy
     optimize_blockwise_fusion
     """
+
+    expr, _ = simplify(expr)
+    expr = optimize_matchpy(expr)
+
+    if fuse:
+        expr = optimize_blockwise_fusion(expr)
+
+    return expr
+
+
+def optimize_matchpy(expr: Expr) -> Expr:
     last = None
     global _defer_to_matchpy
 
-    expr, _ = simplify(expr)
-
     _defer_to_matchpy = True  # take over ==/!= when optimizing
     try:
-        while str(expr) != str(last):
+        while last is None or expr._name != last._name:
             last = expr
             expr = replace_all(expr, replacement_rules)
     finally:
         _defer_to_matchpy = False
-
-    if fuse:
-        expr = optimize_blockwise_fusion(expr)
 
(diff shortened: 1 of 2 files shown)


rjzamora:
We should be able to avoid converting `self._blockwise_layer()` to a `SubgraphCallable` when the subgraph only contains a single task. Let me see how difficult that is.

mrocklin:
Here's a more clear example of the thing that I want: https://github.com/mrocklin/dask-match/pull/27

It came up in this effort because I ran into a graph issue and found that I could no longer debug things, and so became sad.

mrocklin:
OK, this doesn't work because calling `simplify` `__dask_graph__` changes `_name` (as it should) and so keys and the graph don't match up.  This is probably best handled upstream in a change to the Dask base protocol.

mrocklin:
Actually, we can just call simplify on keys too and things work fine.  This should still be handled upstream (this is not the right way to handle this I think) but in order to not bug upstream with this experimental codebase I think it makes sense to do this somewhat suboptimal approach for now.

I tried adding in the replacement rules, but it made tests 3x slower.

mrocklin:
@rjzamora hopefully this unblocks you when it comes to lowering abstract shuffle operations to more concrete ones.

dask_match/expr.py
```diff
@@ -766,7 +766,9 @@ def _task(self, index: int):
         return (self.frame._name, self.partitions[index])
 
     def simplify(self):
-        if isinstance(self.frame, Blockwise):
+        if isinstance(self.frame, Blockwise) and not isinstance(
+            self.frame, BlockwiseIO
```
rjzamora, line 770: This makes sense, because I *think* `BlockwiseIO` is the only case where we will end up getting the same expression back. Can also be more general and just verify that we are not just returning `self.frame` below.

rjzamora (approved):
Looks reasonable to me

PR #27, opened 2023-04-13
Simple graphs

mrocklin:
We ensure that our graphs are simple and easy to understand

@rjzamora is it possible to make this test true?  Sorry, I should have written this a while ago.

PR #28, opened 2023-04-13
Make unfused `Blockwise` graphs simpler

rjzamora:
Possible solution to the question raised in #27

~Avoids the use of `SubgraphCallable` outside of `Fused`. Requires all `Blockwise` expressions to implement a proper `_blockwise_task` method (which produces a partition-agnostic task when `index=None`).  Very open to suggestions here - Especially regarding naming.~

As introduced in https://github.com/rjzamora/dask-match/pull/1, abandons `SubgraphCallable` altogether.


mrocklin:
Interesting approach!

I'm curious about `_blockwise_arg`.  I *think* it's necessary because `BlockwiseArg`s exist.  I'm curious though if, now that we replace `_blockwise_layer` (a static thing) with `_blockwise_task` if maybe `BlockwiseArg` is no longer necessary.  Is that possible?

rjzamora:
>I'm curious about _blockwise_arg. I think it's necessary because BlockwiseArgs exist. I'm curious though if, now that we replace _blockwise_layer (a static thing) with _blockwise_task if maybe BlockwiseArg is no longer necessary. Is that possible?

It will probably take further re-org, but may be possible. Looking into this now.

rjzamora:
**Update**: I spent some time running in circles and ultimately decided that the code gets unnecessarily messy without having a `BlockwiseArg` class (now renamed to `BlockwiseInput`).

>I'm curious about _blockwise_arg. I think it's necessary because BlockwiseArgs exist

The `_blockwise_arg` does provide a single place to handle special logic for `BlockwiseInput` arguments. However, it also allows us to centralize the logic to choose between `expr._name` and `(expr._name, index)`-formatted keys (and so I'd probably want to have the method whether or not `BlockwiseInput` existed).

>with _blockwise_task if maybe BlockwiseArg is no longer necessary

After several experiments, I decided that it is very convenient to be able to include literal input arguments as operands to `Fused`, and the cleanest way to do this is to wrap those inputs in a special type of `Expr`. If we don't do this, we end up needing to add extra logic in multiple places, and it becomes difficult to track the order of interleaved `Expr` and literal-input dependencies.

mrocklin:
What about this approach?

```python
class FromPandas(BlockwiseIO):
    """The only way today to get a real dataframe"""

    _parameters = ["frame", "npartitions"]
    _defaults = {"npartitions": 1}

    @functools.cached_property
    def _name(self):
        return "from-pandas-" + tokenize(self.frame, self.npartitions)

    @property
    def _meta(self):
        return self.frame.head(0)

    def _divisions(self):
        return [None] * (self.npartitions + 1)

    def _blockwise_task(self, index: int | None = None):
        partsize = int(math.ceil(len(self.frame) / self.npartitions))
        start = partsize * index
        stop = partsize * (index + 1)
        return self.frame.iloc[start:stop]

    def __str__(self):
        return "df"

    __repr__ = __str__
```

We don't have the separate dependency, we take in an index and generate the corresponding task (or in this case literal value).  We might not even need SubGraphCallable, we can just generate the entire graph each time.

mrocklin:
I don't have any confidence that that's the right approach.  I don't know this space well enough.  I would hope that something like that would be doable though.  It *feels* right and it's certainly simple.

rjzamora:
>What about this approach?

Yeah, in the case without fusion we can do something like that, but we would still need to support the `index=None` case to allow fusion.

mrocklin:
> but we would still need to support the index=None case to allow fusion.

I wonder if code would generally be simpler if we removed the `index=None` case, and fed the `SubGraphCallable` a graph with `index=0`.  You'd get a bunch of `(..., 0)` wrappers that weren't entirely necessary, but I suspect it would simplify things generally.

rjzamora:
>I wonder if code would generally be simpler if we removed the index=None case, and fed the SubGraphCallable a graph with index=0. You'd get a bunch of (..., 0) wrappers that weren't entirely necessary, but I suspect it would simplify things generally.

Sure - We can probably do this. I'll try it out

mrocklin:
Screwing around a little here: https://github.com/mrocklin/dask-match/pull/29/commits/a79d62a14d6a55694fc0bae13fa2cf6a5124f8c4

rjzamora:
>Screwing around a little here: a79d62a

Thanks for experimenting. I'm having a hard time imagining how `Fused` can work without `BlockwiseIO` expressions adding indexable objects to their dependencies. Without something like `BlockwiseInput`, we just end up repeating the first partition.

mrocklin:
I'm getting pretty close.  I'm pushing to that branch still.  I had to remove SubGraphCallable.

rjzamora:
>I had to remove SubGraphCallable.

Right - This is probably the only way to do what you want. I welcome this experiment, as I don't love using `SubgraphCallable` :)

mrocklin:
Better PR is up here: https://github.com/rjzamora/dask-match/pull/1

rjzamora:
@mrocklin - I reviewed your PR and decided it was the right way to go. Thanks for your help!

PR #29, opened 2023-04-14
No blockwise input

mrocklin:
@rjzamora I'm screwing around here.  Things don't work. Probably I'm going down a path that you've already explored.  If so, please let me know what to expect :)

### DIFF
--- a/dask_match/expr.py
+++ b/dask_match/expr.py
@@ -334,6 +334,9 @@ class Blockwise(Expr):
     This is fairly generic, and includes definitions for `_meta`, `divisions`,
     `_layer` that are often (but not always) correct.  Mostly this helps us
     avoid duplication in the future.
+
+    Note that `Fused` expressions rely on every `Blockwise`
+    expression having a proper `_blockwise_task` method.
     """
 
     operation = None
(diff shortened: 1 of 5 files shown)


mrocklin:
Whoops, wrong root of the PR.  Hopefully the last commit is useful

mrocklin:
Ah, I see now, by sending the `0` index down we don't give the places where there were `BlockwiseInput`s enough information to give the right answer.

🤔

rjzamora:
>Probably I'm going down a path that you've already explored.

Probably - Especially a couple years ago with HLGs.  However, you may be able to make things work without `SubgraphCallable` (a path I didn't explore much yet).

dask_match/expr.py
```diff
@@ -1003,15 +962,17 @@ def _name(self):
     def _divisions(self):
         return self.exprs[0]._divisions()
 
-    def dependencies(self):
-        return self.operands[1:]
-
-    def _blockwise_layer(self):
-        block = {self._name: self.exprs[0]._name}
+    def _task(self, index):
+        graph = {self._name: (self.exprs[0]._name, index)}
```
rjzamora, line 966: Are you worried about re-constructing this graph in a loop for every task? One of the advantages of `SubgraphCallable` is that we only generate it once and then *use* it in every task.

mrocklin: I am mildly concerned about that cost.  I'm far more concerned about the complexity of abstractions that was building up though.  It was getting hard for me to follow, which probably means that it'll be hard for other developers to work on as well.

At the end of the day we're going to be linear in the size of that graph anyway.  For example we're going to need to serialize each subgraphcallable separately to send to the workers.  So we're going to be bound close to some cost similar to this no matter what.

mrocklin:
This is set up more cleanly here: https://github.com/rjzamora/dask-match/pull/1/

PR #30, opened 2023-04-14
Move _task up into Expr

mrocklin:
Previously expressions defined _layer methods.  Most of them were of the form

```python
return {(self._name, i): f(i) for i in range(self.npartitions)}
```

We took advantage of this structure for Blockwise Fused expressions, but maybe it's a good general default to have more often.  This moves the generic version of this up from Blockwise to Expr, and reuses the structure in Head and Partitions, leaving only FromGraph and ApplyConcatApply with custom _layer definitions.

I'm not sure that this is a good idea, but I thought I'd put it up.

### DIFF
--- a/dask_match/expr.py
+++ b/dask_match/expr.py
@@ -128,7 +128,57 @@ class Expr(Operation, metaclass=_ExprMeta):
         # Dependencies are `Expr` operands only
         return [operand for operand in self.operands if isinstance(operand, Expr)]
 
-    def _layer(self):
+    def _task(self, index: int):
+        """The task for the i'th partition
+
+        Parameters
+        ----------
+        index:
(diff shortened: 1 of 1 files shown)


dask_match/expr.py
```diff
@@ -128,6 +128,9 @@ def dependencies(self):
         # Dependencies are `Expr` operands only
         return [operand for operand in self.operands if isinstance(operand, Expr)]
 
+    def _layer(self):
+        return {(self._name, i): self._task(i) for i in range(self.npartitions)}
```
rjzamora, line 132: I may be on board with this if `Expr._task` is defined with a clear `NotImplementedError` message.

rjzamora (approved):
Thanks for adding the docstrings!

PR #31, opened 2023-04-14
Support column access in `Expr.__getattr__`

mrocklin:


dask_match/expr.py
```diff
@@ -117,6 +117,8 @@ def __getattr__(self, key):
             if key in _parameters:
                 idx = _parameters.index(key)
                 return self.operands[idx]
+            if isinstance(self._meta, pd.DataFrame) and key in self._meta.columns:
```
rjzamora, line 120: I'd like to avoid pandas-specific checks, as one of my next highest priorities is to make sure this all works smoothly with a cudf backend.

rjzamora: i.e. can we use `is_dataframe_like` here?

mrocklin: Done.

rjzamora (approved):
Seems fine to me - Thanks!

PR #40, opened 2023-04-16
Add basic statistics function

mrocklin:
This does very little, but it's an example of what we could do.

### DIFF
--- a/dask_match/expr.py
+++ b/dask_match/expr.py
@@ -198,6 +198,30 @@ class Expr(Operation, metaclass=_ExprMeta):
     def size(self):
         return Size(self)
 
+    def _statistics(self):
+        return {}
+
+    def statistics(self) -> dict:
+        """Known quantities of an expression, like length or min/max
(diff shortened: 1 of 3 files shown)


rjzamora:
Cool!

I experimented with this in `dask.dataframe` in the past and ended up wanting partition-wise statistics to do things more complicated than optimizing `len` (i.e. row-count, and column min/max/null_count).

mrocklin:
> and ended up wanting partition-wise statistics to do things more complicated than optimizing

Yup.  I agree.  I added length just because that was trivial to do in a few minutes.  I hope that we start to track more sophisticated metadata.  

My hope is that you give this some thought and it becomes something that we keep in mind.

rjzamora:
>Yup. I agree. I added length just because that was trivial to do in a few minutes... My hope is that you give this some thought and it becomes something that we keep in mind.

Got it. This was my assumption - Thanks for confirming.

PR #41, opened 2023-04-17
Revise `timeseries` to enable deterministic result after optimize

rjzamora:


dask_match/datasets.py
```diff
@@ -224,5 +262,8 @@ def timeseries(
     if seed is None:
         seed = np.random.randint(2e9)
 
-    expr = Timeseries(start, end, dtypes, freq, partition_freq, seed, kwargs)
+    projection = kwargs.pop("_projection", None)  # Not intended for public use
```
rjzamora, line 265: Don't really *need* to expose this in `timeseries` at all, but it does make testing a bit simpler.

mrocklin:
Can you say a bit more about the approach here?  I'm not sure I understand it at first glance.

(also, fwiw, I think that this is probably lower priority than other work.  Happy to engage, but also happy to defer this conversation to later)

rjzamora:
make_timeseries_part already makes the assumption that column projection will not change dtypes, but will only select a subset of columns in dtypes. The _projection operant just saves the column subset to use (if any).

dask_match/datasets.py
```diff
@@ -12,22 +14,43 @@
 
 
 class Timeseries(BlockwiseIO):
-    _parameters = ["start", "end", "dtypes", "freq", "partition_freq", "seed", "kwargs"]
+    _parameters = [
+        "start",
+        "end",
+        "dtypes",
+        "_projection",
```
mrocklin, line 21: This seems wrong to me.  I would hope to just change `dtypes` to reflect the new columns.

I think that the way I would normally solve this would be to create a separate set of seeds for each column.  Then when we drop a column all of the seeds for the other columns stay the same.  This means more seeds, but hoepfully those are cheap.

rjzamora: Okay - Updated the PR to use a random state of length `num-partitions * num-columns`. Sorry, I overlooked the fact that we have control of `make_timeseries_part`. My mistake.

mrocklin:
This looks good to me.  Thanks @rjzamora !

Issue #12, opened 2023-04-04
Culling


mrocklin:
There has been pain around graph culling in the past.  We want to be efficient in cases where people ask for only a small subset of the output result, and that requires only a small subset of the graph to be generated.  This occurs in situations like `.head` and array slicing.  

My _hope_ is that a lot of this pain goes away with high level optimization.  For example we should have optimizations that push `.head` through many operations (everything except IO, shuffling, ACA, etc..).  Even then there are a couple of options:

1.  Have operations like read_parquet and shuffle know the output partitions that are requested as part of their own metadata (so the head operation disappears entirely once optimized)
2. Keep the head optimization around, and accept some wasted generation, but feel glad because it's only happening on one layer, rather than on many layers.

Regardless, a good first step here, I think, is to build some optimizations around Head

rjzamora:
Related: https://github.com/mrocklin/dask-match/pull/16

rjzamora:
Now that we have `Partitions` and `PartitionsFiltered`, I think we have a reasonable plan to handle culling. Therefore, I'm going to close this issue.

If we run into culling-related pain in the future, we can reference this issue in a separate discussion.

PR #42, opened 2023-04-18
Improve culling

rjzamora:
While playing around with `pprint` and `dask.visualize`, I discovered a lot of fragile behavior when combining various optimizations with `Head`/`Partitions` (i.e. fusion and column projection).  After a lot of experimentation, I found that the simplest way to allow all three optimizations to work smoothly together is to do the following:

1.  Add special logic in `Expr` for the case that an optional `_take_partitions` parameter is defined.
2. Change the `head` logic to use `Head(Partitions(...))`
3. Change `Head` to inherit from `Blockwise`. The purpose of this expression is now to return the first `n` rows of every partition.

#### Meaning of `_take_partitions`:

When an `Expr` subclass exposes "_take_partitions" in its `_parameters` list, that subclass (`cls`) is ensuring that the behavior of `cls(*operands, _take_partitions=[0])` is equivalent to that of `Partitions(cls(*operands), [0])`.

#### Why we want `Partitions` to "simplify away:

Optimizations like fusion, column-projection and predicate pushdown are all "blocked" when a `Partitions`/`Head` operation follows a `BlockwiseIO` expression. The only way to "unblock" these optimizations is if the necessary expressions can "absorb" the `Partitions` logic.

#### why we probably want `Expr` to be aware of `_take_partitions`:

This PR only adds `_take_partitions` to a few IO layers. However, we will eventually want to add the same logic for many other expressions that are likely to precede `Partitions` (e.g. joins and shuffles).

### Example of current behavior

```python
from dask_match import from_pandas, optimize
from dask.base import visualize
import pandas as pd

ddf = from_pandas(
    pd.DataFrame({"x": range(30), "id": [99, 1000, 1001] * 10}),
    npartitions=3,
)
s = (ddf["x"][ddf["id"] > 1000] + 1).head(compute=False)
optimize(s, fuse=False).pprint()
```
```
Add:1
  Head:
    Filter:
      Projection:'x'
        FromPandas: npartitions=3
      GT:1000
        Projection:'id'
          FromPandas: npartitions=3
```
```python
visualize(s, filename=None, optimize_graph=False, orientation="left", size="8")
```
Screen Shot 2023-04-18 at 1 41 36 PM

```python
visualize(optimize(s), filename=None, optimize_graph=False, orientation="left", size="8")
```

Screen Shot 2023-04-18 at 2 00 17 PM

### Example of NEW behavior

```python
optimize(s, fuse=False).pprint()
```
```
Add:1
  Head:
    Filter:
      Projection:'x'
        FromPandas: npartitions=3 _take_partitions=[0]
      GT:1000
        Projection:'id'
          FromPandas: npartitions=3 _take_partitions=[0]
```
```python
visualize(s, filename=None, optimize_graph=False, orientation="left", size="8")
```
Screen Shot 2023-04-18 at 1 57 31 PM

```python
visualize(optimize(s), filename=None, optimize_graph=False, orientation="left", size="8")
```
Screen Shot 2023-04-18 at 1 58 11 PM

### DIFF
--- a/dask_match/datasets.py
+++ b/dask_match/datasets.py
@@ -12,7 +12,16 @@ __all__ = ["timeseries"]
 
 
 class Timeseries(BlockwiseIO):
-    _parameters = ["start", "end", "dtypes", "freq", "partition_freq", "seed", "kwargs"]
+    _parameters = [
+        "start",
+        "end",
+        "dtypes",
+        "freq",
+        "partition_freq",
+        "seed",
+        "kwargs",
+        "_take_partitions",
+    ]
     _defaults = {
         "start": "2000-01-01",
         "end": "2000-01-31",
@@ -21,6 +30,7 @@ class Timeseries(BlockwiseIO):
         "partition_freq": "1d",
         "seed": None,
         "kwargs": {},
+        "_take_partitions": None,
     }
 
     @property
@@ -37,9 +47,17 @@ class Timeseries(BlockwiseIO):
 
     @functools.cached_property
     def random_state(self):
-        return np.random.randint(2e9, size=self.npartitions)
+        size = (
+            len(pd.date_range(start=self.start, end=self.end, freq=self.partition_freq))
+            + 1
+        )
+        return np.random.randint(2e9, size=size)
 
     def _task(self, index):
+        original_index = index
+        if self._take_partitions is not None:
+            # Need original index for random_state
+            original_index = self._take_partitions[index]
         return (
             make_timeseries_part,
             self.divisions[index],
@@ -47,7 +65,7 @@ class Timeseries(BlockwiseIO):
             self.operand("dtypes"),
             self.columns,
             self.freq,
-            self.random_state[index],
+            self.random_state[original_index],
             self.kwargs,
         )
 
--- a/dask_match/expr.py
+++ b/dask_match/expr.py
@@ -328,7 +328,17 @@ class Expr(Operation, metaclass=_ExprMeta):
 
     @functools.cached_property
     def divisions(self):
-        return tuple(self._divisions())
+        # Common case: Use self._divisions()
+        full_divisions = tuple(self._divisions())
+        if "_take_partitions" not in self._parameters or self._take_partitions is None:
+            return full_divisions
+
+        # Specific case: Specific partitions were selected
+        new_divisions = []
+        for part in self._take_partitions:
+            new_divisions.append(full_divisions[part])
+        new_divisions.append(full_divisions[part + 1])
+        return tuple(new_divisions)
 
     def _divisions(self):
         raise NotImplementedError()
@@ -340,7 +350,10 @@ class Expr(Operation, metaclass=_ExprMeta):
 
     @property
     def npartitions(self):
-        if "npartitions" in self._parameters:
+        if "_take_partitions" in self._parameters and self._take_partitions is not None:
+            # Special case: Specific partitions were selected
+            return len(self._take_partitions)
+        elif "npartitions" in self._parameters:
             idx = self._parameters.index("npartitions")
             return self.operands[idx]
         else:
@@ -662,7 +675,13 @@ class Index(Elemwise):
         )
 
 
-class Head(Expr):
+class Head(Blockwise):
+    """Take the first `n` rows of every partition
+
+    Typically used after `Partition(..., [0])` to take
+    the first `n` rows of an entire collection.
+    """
+
     _parameters = ["frame", "n"]
     _defaults = {"n": 5}
 
@@ -817,6 +836,19 @@ class Partitions(Expr):
             operands = [
                 Partitions(op, self.partitions) if isinstance(op, Expr) else op
                 for op in self.frame.operands
+            ]
+            return type(self.frame)(*operands)
+        elif (
+            "_take_partitions" in self.frame._parameters
+            and self.frame._take_partitions is None
+        ):
+            # We assume that expressions defining a special "_take_partitions"
+            # parameter can internally capture the same logic as `Partitions`
+            operands = [
+                self.partitions
+                if self.frame._parameters[i] == "_take_partitions"
+                else op
+                for i, op in enumerate(self.frame.operands)
             ]
             return type(self.frame)(*operands)
 
--- a/dask_match/io/tests/test_io.py
+++ b/dask_match/io/tests/test_io.py
@@ -2,9 +2,11 @@ import os
 
 import pandas as pd
 import pytest
+import dask.dataframe as dd
 from dask.dataframe.utils import assert_eq
 
-from dask_match import optimize, read_parquet, read_csv
+from dask_match import from_pandas, optimize, read_parquet, read_csv
+from dask_match.utils import _check_take_partitions
 
 
 def _make_file(dir, format="parquet", df=None):
@@ -93,7 +95,7 @@ def test_predicate_pushdown(tmpdir):
     fn = _make_file(tmpdir, format="parquet", df=original)
     df = read_parquet(fn)
     assert_eq(df, original)
-    x = df[df.a == 5][df.c > 20]["b"]
+    x = df[df.a == 5][df.c > 20][["b"]]
     y = optimize(x)
     assert isinstance(y.expr, ReadParquet)
     assert ("a", "==", 5) in y.expr.operand("filters") or (
@@ -109,3 +111,29 @@ def test_predicate_pushdown(tmpdir):
     assert list(y_result.columns) == ["b"]
     assert len(y_result["b"]) == 6
     assert all(y_result["b"] == 4)
+
+
+@pytest.mark.parametrize("fmt", ["parquet", "csv", "pandas"])
+def test_io_culling(tmpdir, fmt):
+    pdf = pd.DataFrame({c: range(10) for c in "abcde"})
+    if fmt == "parquet":
+        dd.from_pandas(pdf, 2).to_parquet(tmpdir)
+        df = read_parquet(tmpdir)
+    elif fmt == "parquet":
+        dd.from_pandas(pdf, 2).to_csv(tmpdir)
+        df = read_csv(tmpdir + "/*")
+    else:
+        df = from_pandas(pdf, 2)
+    df = (df[["a", "b"]] + 1).partitions[1]
+    df2 = optimize(df)
+
+    # All tasks should be fused for the single output partition
+    assert df2.npartitions == 1
+    assert len(df2.dask) == df2.npartitions
+    expected = pdf.iloc[5:][["a", "b"]] + 1
+    assert_eq(df2, expected, check_index=False)
+
+    # Check that we still get culling without fusion
+    df3 = optimize(df, fuse=False)
+    _check_take_partitions(df3.expr, [1])
+    assert_eq(df3, expected, check_index=False)
--- a/dask_match/collection.py
+++ b/dask_match/collection.py
@@ -118,8 +118,7 @@ class FrameBase(DaskMethodsMixin):
 
     def head(self, n=5, compute=True):
         # We special-case head because matchpy uses 'head' as a special term
-
-        out = new_collection(expr.Head(self.expr, n=n))
+        out = new_collection(expr.Head(expr.Partitions(self.expr, [0]), n=n))
         if compute:
             out = out.compute()
         return out
--- a/dask_match/collection.py
+++ b/dask_match/collection.py
@@ -264,6 +264,10 @@ def optimize(collection, fuse=True):
     return new_collection(expr.optimize(collection.expr, fuse=fuse))
 
 
+def simplify(collection):
+    return new_collection(expr.simplify(collection.expr)[0])
+
+
 def from_pandas(*args, **kwargs):
     from dask_match.io.io import FromPandas
 
--- a/dask_match/io/csv.py
+++ b/dask_match/io/csv.py
@@ -4,8 +4,8 @@ from dask_match.io.io import BlockwiseIO
 
 
 class ReadCSV(BlockwiseIO):
-    _parameters = ["filename", "usecols", "header"]
-    _defaults = {"usecols": None, "header": "infer"}
+    _parameters = ["filename", "usecols", "header", "_take_partitions"]
+    _defaults = {"usecols": None, "header": "infer", "_take_partitions": None}
 
     @functools.cached_property
     def _ddf(self):
@@ -35,4 +35,6 @@ class ReadCSV(BlockwiseIO):
         return next(iter(dsk.values()))[0]
 
     def _task(self, index: int):
+        if self._take_partitions is not None:
+            index = self._take_partitions[index]
         return (self._io_func, self._tasks[index][1])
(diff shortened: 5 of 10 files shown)


rjzamora:
@mrocklin - I'm expecting critical comments on this (so don't worry if you have them). That said, establishing a plan for culling is a relatively strong priority for me.

rjzamora:
Related to #12

dask_match/datasets.py
```diff
@@ -12,7 +12,16 @@
...
-    _parameters = ["start", "end", "dtypes", "freq", "partition_freq", "seed", "kwargs"]
+    _parameters = [
+        "start",
+        "end",
+        "dtypes",
+        "freq",
+        "partition_freq",
+        "seed",
+        "kwargs",
+        "_take_partitions",
```
mrocklin, line 23: Thoughts on other names?  Maybe just `_partitions`?

dask_match/datasets.py
```diff
@@ -21,6 +30,7 @@ class Timeseries(BlockwiseIO):
         "partition_freq": "1d",
         "seed": None,
         "kwargs": {},
+        "_take_partitions": None,
```
mrocklin, line 33: I'm curious if it makes sense to use None here, or to always use `range(self.npartitions)` instead

dask_match/datasets.py
```diff
@@ -37,17 +47,25 @@ def _divisions(self):
 
     @functools.cached_property
     def random_state(self):
-        return np.random.randint(2e9, size=self.npartitions)
+        size = (
+            len(pd.date_range(start=self.start, end=self.end, freq=self.partition_freq))
+            + 1
+        )
```
mrocklin, line 53: I'm curious, why this change?

dask_match/datasets.py
```diff
@@ -37,17 +47,25 @@ def _divisions(self):
...
+        return np.random.randint(2e9, size=size)
 
     def _task(self, index):
+        original_index = index
+        if self._take_partitions is not None:
+            # Need original index for random_state
+            original_index = self._take_partitions[index]
```
mrocklin, line 60: I propose that we handle this in `Expr._layer` instead

mrocklin: Maybe something like this?

```python
class Expr:
    @property
    def _partitions(self):
        return range(self.npartitions)

    def _layer(self):
        return {(self._name, i): self._task(k) for i, k in enumerate(self._partitions)}
```

mrocklin: Or this:

```class Expr:
    def _layer(self):
        return {(self._name, i): self._task(k) for i, k in enumerate(getattr(self, "_partitions", range(self.npartitions)))}
```

mrocklin (commented):
I'm not sure yet how I feel about this, but for the moment I'm suspending suspicion and reviewing this as though it will go in.  Maybe after a while of iterating here we both think "yeah, this seems pretty clean".

I don't think that we're there just yet, but it doesn't seem unreasonable to get there.

What are your thoughts?

dask_match/datasets.py
rjzamora: I used `_partitions` at first be went with something more unique to avoid confusion with the collection-level `_partitions`. Though, I'll admit I like the simplicity of `_partitions` better. So, very open to changing it :)

dask_match/datasets.py
rjzamora: I thought a bit more about this and couldn't come up with a way to avoid using `None` as the default. The first reason is that, unless the expression includes an explicit `npartitions` parameter, we don't typically know how many partitions will be produced until after some amount of logic within the expression itself gets executed. It is also useful to be able to check if `self.operand("_partitions") is None`, because this tells us that the expression does not correspond to a culled/sliced operation.

dask_match/datasets.py
rjzamora: Okay - I was able to get this to work after I realized that we just need extra logic in `Fused._task` to deal with the fact that the index for one of the wrapped expressions may be different from the initial `index` argument.

mrocklin:
> Optimizations like fusion, column-projection and predicate pushdown are all "blocked" when a Partitions/Head operation follows a BlockwiseIO expression

Question: why is this?   In principle it should be easy to have a `Projection` pass through a `Partitions` with simplify or replacement rules.  Is this because they keep swapping with each other or something?

If we could make it so that projection/filter operations went through Projections would you still want the `_partitions` code written here?

rjzamora:
Sorry for not clarifying. Projection and predicate-pushdown can be unblocked without this. However, we cannot fuse the IO expression if there is a non-Blockwise Partitions expression right after it.

mrocklin:
> However, we cannot fuse the IO expression if there is a non-Blockwise Partitions expression right after it

What I'm hearing is that we have two options:

Have expressions like this:

    ReadParquet: ... partitions=...

or like this:

    Partitions:
        ReadParquet: ...


The benefit of the first I think is that we don't have any dangling tasks.  The benefit of the second I think is that it's a bit more composable.  Complex classes like ReadParquet/Shuffle/Repartition have one less thing to think about.  

I think that we have option 2 today and option 1 is what is implemented in this PR.  Are there advantages to option 1 other than what I describe above?  If not, I was initially inclined to optimize for lower complexity over  performance, at least for now.  

However, I then started to think about counter-examples to see if I could convince myself otherwise.  The best counter-argument I found were workloads like image processing with arrays and map overlap and slicing.

```python
x = da.from_array(...)
x = x.map_overlap(...)
x[:20, :20]
```

We get to skip the map_overlap cost, but still have to deal with `from_array`.  🤔 maybe that convinces me.  I'm not sure yet.  

I'm mildly nervous but not against this work.  Mostly I'm writing this to share my thinking and ask for any thoughts to improve that thinking.

dask_match/datasets.py
```diff
@@ -41,25 +43,29 @@ def _meta(self):
             "2000", "2000", dtypes, list(dtypes.keys()), "1H", states, self.kwargs
         )
 
+    @functools.lru_cache
```
mrocklin, line 46: This is already cached on `divisions`.  Maybe use that instead?

rjzamora:
We could rewrite the io expressions to support option 2, but I’m not seeing how option 2 helps us for shuffles/joins/groupby (unless the only concern is blocked fusion).

mrocklin:
> but I’m not seeing how option 2 helps us for shuffles/joins/groupby

Ah, right, what I'm hearing you say is that we're definitely going to need something like this for shuffling because this isn't just graph-generation costs, but computation costs.  Given that we're going to do it for shuffle operations, we might as well generalize it and do it across the codebase.  

Am I understanding that correctly?

mrocklin:
OK, I think I'm convinced.  Thank you for your patience with me on this topic.

dask_match/datasets.py
rjzamora: We cannot use `self.divisions` within `self.random_state` or `self._task`, because we need access to the original divisions (before `_partitions` was taken into account).  `self.divisions` will have already selected the `_partitions` indices.

dask_match/expr.py
```diff
@@ -1124,12 +1185,13 @@ def _divisions(self):
     def _task(self, index):
         graph = {self._name: (self.exprs[0]._name, index)}
         for _expr in self.exprs:
+            task_index = _expr._partitions[index]
```
mrocklin, line 1188: Just calling out here that it feels off that `_partitions` has to affect code in `Fused`.  This seems like an abstraction that leaks around to others, and leads to a hard-to-understand codebase.  I don't see a better way currently, but I want to call this out as concerning so that we don't get too comfortable doing things like this.

mrocklin: Could we handle this in the IO subclasses instead with something like the following?

```python
index 21bc986..258cfa2 100644
--- a/dask_match/expr.py
+++ b/dask_match/expr.py
@@ -1185,13 +1185,12 @@ class Fused(Blockwise):
     def _task(self, index):
         graph = {self._name: (self.exprs[0]._name, index)}
         for _expr in self.exprs:
-            task_index = _expr._partitions[index]
             if isinstance(_expr, Fused):
-                (_, subgraph, name) = _expr._task(task_index)
+                (_, subgraph, name) = _expr._task(index)
                 graph.update(subgraph)
                 graph[(name, index)] = name
             else:
-                graph[(_expr._name, index)] = _expr._task(task_index)
+                graph[(_expr._name, index)] = _expr._task(index)

         for i, dep in enumerate(self.dependencies()):
             graph[self._blockwise_arg(dep, index)] = "_" + str(i)
diff --git a/dask_match/io/csv.py b/dask_match/io/csv.py
index 2f139ff..b7cd441 100644
--- a/dask_match/io/csv.py
+++ b/dask_match/io/csv.py
@@ -35,4 +35,6 @@ class ReadCSV(BlockwiseIO):
         return next(iter(dsk.values()))[0]

     def _task(self, index: int):
+        if self._partitions is not None:
+            index = self._partitions[index]
         return (self._io_func, self._tasks[index][1])
diff --git a/dask_match/io/io.py b/dask_match/io/io.py
index fdd9ba5..c4e8e2b 100644
--- a/dask_match/io/io.py
+++ b/dask_match/io/io.py
@@ -75,6 +75,8 @@ class FromPandas(BlockwiseIO):
         return self._divisions_and_locations[1]

     def _task(self, index: int):
+        if self._partitions is not None:
+            index = self._partitions[index]
         start, stop = self._locations()[index : index + 2]
         return self.frame.iloc[start:stop]

diff --git a/dask_match/io/parquet.py b/dask_match/io/parquet.py
index aac3bd2..989f670 100644
--- a/dask_match/io/parquet.py
+++ b/dask_match/io/parquet.py
@@ -291,4 +291,6 @@ class ReadParquet(BlockwiseIO):
         return self._plan["divisions"]

     def _task(self, index: int):
+        if self._partitions is not None:
+            index = self._partitions[index]
         return (self._plan["func"], self._plan["parts"][index])
```

This doesn't work, but I haven't investigated deeply yet.

dask_match/expr.py
```diff
@@ -169,6 +169,22 @@ def dependencies(self):
         # Dependencies are `Expr` operands only
         return [operand for operand in self.operands if isinstance(operand, Expr)]
 
+    @property
+    def _culled(self) -> bool:
```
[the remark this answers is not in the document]
mrocklin: A thought, rather than add this to all Exprs, maybe we have a `PartitionsFiltered` mixin class that looks like the following:

```python
class PartitionsFiltered(Expr):
    def _partitions(self):
        ...
    def _divisions(self):
        ...
    def _layer(self):
        ...
```

And whatever other methods need modification.  Then we add this mixin to the IO classes

```python
class BlockwiseIO(Blockwise, PartitionsFiltered):
    ...
```

This is probably a bit more code, but my guess is that it isolates the oddity of `_partitions` to just those classes that are affected by it.

dask_match/io/tests/test_io.py
```diff
@@ -111,6 +112,39 @@ def test_predicate_pushdown(tmpdir):
...
+    assert len(df2.dask) == df2.npartitions
+    expected = pdf.iloc[5:][["a", "b"]] + 1
+    assert_eq(df2, expected, check_index=False)
+
+    def _check_culling(expr, partitions):
+        """CHeck that _partitions is set to the expected value"""
+        for dep in expr.dependencies():
+            _check_culling(dep, partitions)
+        if "_partitions" in expr._parameters:
+            assert expr._partitions == partitions
```
mrocklin, line 140: I moved this out of `utils.py`.  I understand that this means that we can't test datasets.timeseries.  I think that in the future we should pull out `assert_eq` and make our own.  This would then go there and be more universal.  Until then I'd like to have this isolated-ish if you're cool with it.

dask_match/datasets.py
mrocklin: Makes sense.  Thanks!

mrocklin (commented):
I was playing with this locally.  I made a few changes (hopefully ok) and have some ideas on how to isolate this away from `Expr`.  See comments below.  I'm curious about if that approach is feasible.

dask_match/expr.py
rjzamora: Right - This PR originally proposed that any `Expr` supporting `_partitions` would need to account for this operand in its `_task` implementation. When `_layer` was agnostic to `_partitions`, there were no problems with `Fused`.  I only attempted to move the logic into `_layer` in response to [your comment](https://github.com/mrocklin/dask-match/pull/42#discussion_r1170612271), so I'd be happy to go in a different direction.

dask_match/io/tests/test_io.py
rjzamora: Works for me

dask_match/expr.py
rjzamora: Took a quick stab at the `PartitionsFiltered`, and moved the logic out of `_layer`. Let me know if you think that direction makes sense.

dask_match/expr.py
```diff
@@ -887,6 +907,81 @@ def simplify(self):
...
+    In order to leverage the default `Expr._layer`
+    method, subclasses should define `_filtered_task`
+    instead of `_task`.
+    """
+
+    @property
+    def _filtered(self) -> bool:
+        """Whether or not output partitions have been filtered"""
+        return (
+            "_partitions" in self._parameters
```
mrocklin, line 943: I'm imagining that `"_partitions" in self._parameters` is as an expectation of this class.

dask_match/expr.py
```diff
@@ -887,6 +907,81 @@ def simplify(self):
...
+    @property
+    def npartitions(self):
+        if self._filtered:
+            # Special case: Specific partitions were selected
+            return len(self._partitions)
+        elif "npartitions" in self._parameters:
+            idx = self._parameters.index("npartitions")
+            return self.operands[idx]
+        else:
+            return len(self.divisions) - 1
```
mrocklin, line 978: This could maybe simplified, given the new definition of `divisions` above

dask_match/expr.py
```diff
@@ -887,6 +907,81 @@ def simplify(self):
...
+            if self.frame._partitions:
+                partitions = [self.frame._partitions[p] for p in self.partitions]
+            else:
+                partitions = self.partitions
+            # We assume that expressions defining a special "_partitions"
+            # parameter can internally capture the same logic as `Partitions`
+            operands = [
+                partitions if self.frame._parameters[i] == "_partitions" else op
+                for i, op in enumerate(self.frame.operands)
+            ]
```
mrocklin, line 920: We do this pattern a lot. I want to think about some dict variant of operands in the future I think

dask_match/expr.py
mrocklin: Good evidence of the fact that I am often wrong!

mrocklin (commented):
Some small thoughts below.  How are you feeling about this direction so far?

dask_match/expr.py
rjzamora: I was thinking about that as well. One complication is that we don't want to require a key for every operand (recall that variadic expressions are useful). Therefore, it may be less restrictive if we just support something like `Expr.replace_operands(**operand_replacements)` so that this logic can be isolated in one place.

PR #43, opened 2023-04-19
Add abstract Mean operation

mrocklin:
This came up in conversation with @jorisvandenbossche

It's a good example of an operation that doesn't have any concrete implementation, but instead optimizes itself down to other operations for execution.  This is our approach with shuffles, but this is probably easier to explain.


rjzamora (approved):
Makes sense!

PR #46, opened 2023-04-21
Add sort argument to `from_pandas`

rjzamora:
**Context**: While working on repartitioning, I found that I wanted `from_pandas` to give me known partitions by default.

### DIFF
--- a/dask_match/expr.py
+++ b/dask_match/expr.py
@@ -397,7 +397,9 @@ class Expr(Operation, metaclass=_ExprMeta):
         Parameters
         ----------
         substitutions:
-            mapping old terms to new terms
+            mapping old terms to new terms. Note that using
+            non-`Expr` keys may produce unexpected results,
+            and substituting boolean values is not allowed.
 
         Examples
         --------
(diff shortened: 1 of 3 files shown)


dask_match/expr.py
```diff
@@ -413,7 +415,11 @@ def substitute(self, substitutions: dict) -> "Expr":
         new = []
         update = False
         for operand in self.operands:
-            if ishashable(operand) and operand in substitutions:
+            if (
+                not isinstance(operand, bool)
```
rjzamora, line 419: My preference would be to disallow non-`Expr` replacement in `substitute`. For now I'm only disallowing boolean replacement, since there is no clean way to distinguish between `True`/`1`/`1.0` here, but I'm expecting something like `expr.substitute({<literal>: <literal>})` to be a pretty dangerous thing to do in cases where `expr` isn't trivial.

rjzamora: **Context**: Adding `sort` to `FromPandas` broke `test_substitute`, because `(df + 1).substitute({1: 2})` results in the default `sort=True` setting being changed to `sort=2` (because `1 == True`).

rjzamora: Decided to merge this PR for now since it is difficult to test repartitioning without divisions. I'll be happy to revisit this tangential question.

mrocklin: I'm ok with dropping this if you think best

mrocklin: > Decided to merge this PR for now since it is difficult to test repartitioning without divisions. I'll be happy to revisit this tangential question.

Thanks for calling this out.  I think it's fine to be loose at this stage.

PR #47, opened 2023-04-21
Add `npartitions` and `divisions`-based repartitioning

rjzamora:
Adds an abstract `Repartition` expression, and subclasses for `npartitions` and `divisions`-based repartitioning.

### DIFF
--- a/dask_match/repartition.py
+++ b/dask_match/repartition.py
@@ -83,11 +83,12 @@ class Repartition(Expr):
             raise NotImplementedError()
 
     def _simplify_up(self, parent):
-        # Move filter or column projection before repartitioning
-        if isinstance(parent, (Filter, Projection)) and not isinstance(
-            parent.operands[1], Expr
-        ):
-            op = type(parent)(self.frame, *parent.operands[1:])
+        # Reorder with column projection
+        if isinstance(parent, Projection):
+            return type(self)(self.frame[parent.columns], *self.operands[1:])
+        # Reorder with filter
+        if isinstance(parent, Filter) and not isinstance(parent.predicate, Expr):
+            op = type(parent)(self.frame, parent.predicate)
             return type(self)(op, *self.operands[1:])
 
 
--- a/dask_match/repartition.py
+++ b/dask_match/repartition.py
@@ -1,17 +1,23 @@
 import functools
 from operator import getitem
+from pprint import pformat
 
 import pandas as pd
+import numpy as np
 from pandas.api.types import is_datetime64_any_dtype, is_numeric_dtype
+from tlz import unique
 
 from dask.dataframe import methods
 from dask.dataframe.core import split_evenly
+from dask.dataframe.utils import is_series_like
 
 from dask_match.expr import Expr
 
 
 class Repartition(Expr):
-    _parameters = ["frame", "n", "new_divisions"]
+    """Abstract repartitioning expression"""
+
+    _parameters = ["frame", "n", "new_divisions", "force"]
 
     @property
     def _meta(self):
@@ -19,7 +25,7 @@ class Repartition(Expr):
 
     def _divisions(self):
         if self.n is not None:
-            return (None,) * (self.n + 1)
+            return self.simplify()._divisions()
         return self.new_divisions
 
     def simplify(self):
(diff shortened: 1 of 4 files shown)


dask_match/repartition.py
```diff
@@ -0,0 +1,332 @@
...
+
+    def _divisions(self):
+        if self.n is not None:
+            return self.simplify()._divisions()
+        return self.new_divisions
+
+    def simplify(self):
+        if self.n is not None:
+            if self.n < self.frame.npartitions:
+                return ReducePartitionCount(self.frame, self.n)
```
mrocklin, line 35: I suggest that all of these classes should maybe start with `Repartition`, like `RepartitionToMorePartitions` and `RepartitionToFewerPartitions`.  Mostly I care about Repartition being in the name, and probably the first term.

dask_match/repartition.py
```diff
@@ -0,0 +1,332 @@
...
+                isinstance(frame, cls)
+                and not isinstance(other, Expr)
+            )
+        )
+        for op in [Filter, Projection]:
+            pattern = Pattern(op(frame, other), supported)
+            yield ReplacementRule(pattern, functools.partial(reorder, op=op))
+
+
+class RepartitionImpl(Repartition):
```
mrocklin, line 102: I suggest maybe the following names

```python
Repartition  # the super-class
    RepartitionAbstract
    RepartitionToMore
    RepartitionToFewer
    RepartitionDivisions
```

mrocklin (commented):
Seems fine to me.  Thanks @rjzamora .  A small comment about names.

dask_match/repartition.py
rjzamora: Cool - Make sense to me.

dask_match/repartition.py
```diff
@@ -1,22 +1,113 @@
...
+        supported = CustomConstraint(
+            lambda frame, other: (
+                # filter-pushdown gets messy if other is an Expr
+                isinstance(frame, cls)
+                and not isinstance(other, Expr)
+            )
+        )
+        for op in [Filter, Projection]:
+            pattern = Pattern(op(frame, other), supported)
+            yield ReplacementRule(pattern, functools.partial(reorder, op=op))
```
rjzamora, line 103: Will need to rewrite this with `simplify_up` - So, I'd say we should treat #50 as a dependency for this PR.

rjzamora: Done.

dask_match/repartition.py
```diff
@@ -1,22 +1,101 @@
...
 
+    def _divisions(self):
+        if self.n is not None:
+            return self.simplify()._divisions()
+        return self.new_divisions
+
+    def _simplify_down(self):
+        if type(self) != Repartition:
+            # This simplify logic should not be inherited
+            return None
```
mrocklin, line 35: Interesting pattern

dask_match/repartition.py
```diff
@@ -1,22 +1,101 @@
...
+        elif self.new_divisions:
+            return RepartitionDivisions(self.frame, self.new_divisions, self.force)
+        else:
+            raise NotImplementedError()
+
+    def _simplify_up(self, parent):
+        # Move filter or column projection before repartitioning
+        if isinstance(parent, (Filter, Projection)) and not isinstance(
+            parent.operands[1], Expr
+        ):
```
mrocklin, line 89: Isn't the predicate of `Filter` likely to be an expression?

I wonder if it makes sense to break these two cases out and repeat a little.  Some duplication, but maybe the code is easier to understand.  Happy either way.

dask_match/repartition.py
```diff
@@ -1,22 +1,101 @@
...
+            op = type(parent)(self.frame, *parent.operands[1:])
+            return type(self)(op, *self.operands[1:])
```
mrocklin, line 91:
```python
    return self.frame[parent.columns].repartition(*self.operands[1:])
if isinstance(parent, Filter):
    ...
```

mrocklin: Or I guess we want to use the proper Op, given that we may already have lowered.  In practice though I expect `simplify_up` to be called before `simplify_down`.  It's best not to rely on that though.

mrocklin (approved):

dask_match/repartition.py
rjzamora: >Isn't the predicate of Filter likely to be an expression?

Ah right - I confused myself here.

mrocklin: A common effect of DRY code 😉

PR #48, opened 2023-04-21
Improve `pprint` for expressions containing fusion

rjzamora:
**TODO**: Add testing and consider other names/defaults. (Suggestions are welcome)

**Example Behavior**:

```python
from dask_match.datasets import timeseries
from dask_match import optimize
import pandas as pd

ddf = timeseries()
expr = optimize(ddf[ddf.index < pd.Timestamp("2000-01-20")]["id"] + ddf["x"].sum() + 123)
```

```python
expr.pprint(collapse_fusion=True)
```
```
Add(Fused with 7 expressions): right=123
  Sum:
    Projection(Fused with 1 expressions): columns='x'
```

```python
expr.pprint(collapse_fusion=False)
```
```
Add(699a0): right=123
  Add(Fused with 699a0):
    Filter(Fused with 699a0):
      Projection(Fused with 699a0): columns='id'
        Timeseries(Fused with 699a0): dtypes={'id': <class 'int'>} seed=1727522390
      LT(Fused with 699a0): right=Timestamp('2000-01-20 00:00:00')
        Index(Fused with 699a0):
          Timeseries(Fused with 699a0): seed=1727522390
  Sum:
    Projection(Fused with 1 expressions): columns='x'
```

**Old Behavior**:

```python
expr.pprint()
```
```
Fused:[Filter(frame=Timeseries(7ca40e3)['id'], predicate=Index(frame=Timeseries(c133ead)) < 2000-01-20 00:00:00) + Fused-getitem-timeseries.sum() + 123, Filter(frame=Timeseries(7ca40e3)['id'], predicate=Index(frame=Timeseries(c133ead)) < 2000-01-20 00:00:00) + Fused-getitem-timeseries.sum(), Filter(frame=Timeseries(7ca40e3)['id'], predicate=Index(frame=Timeseries(c133ead)) < 2000-01-20 00:00:00), Timeseries(7ca40e3)['id'], Timeseries(7ca40e3), Index(frame=Timeseries(c133ead)) < 2000-01-20 00:00:00, Index(frame=Timeseries(c133ead)), Timeseries(c133ead)]
  Sum:
    Fused:[Timeseries(0fd1cbf)['x'], Timeseries(0fd1cbf)]
```

### DIFF
--- a/dask_match/expr.py
+++ b/dask_match/expr.py
@@ -112,38 +112,42 @@ class Expr(Operation, metaclass=_ExprMeta):
     def __repr__(self):
         return str(self)
 
-    def _tree_repr_lines(self, indent=0):
-        header = funcname(type(self)) + ":"
+    def _tree_repr_operand(self, i, lines, header, collapse_fusion=True):
+        op = self.operands[i]
+        if isinstance(op, Expr):
+            lines.extend(op._tree_repr_lines(2, collapse_fusion=collapse_fusion))
+        else:
+            try:
+                param = self._parameters[i]
+                default = self._defaults[param]
+            except (IndexError, KeyError):
+                param = self._parameters[i] if i < len(self._parameters) else ""
+                default = "--no-default--"
+
+            if isinstance(op, pd.core.base.PandasObject):
+                op = "<pandas>"
+
+            elif repr(op) != repr(default):
+                if param:
+                    header += f" {param}={repr(op)}"
+                else:
+                    header += repr(op)
+        return lines, header
+
+    def _tree_repr_lines(self, indent=0, collapse_fusion=True):
         lines = []
-        for i, op in enumerate(self.operands):
-            if isinstance(op, Expr):
-                lines.extend(op._tree_repr_lines(2))
-            else:
-                try:
-                    param = self._parameters[i]
-                    default = self._defaults[param]
-                except (IndexError, KeyError):
-                    param = ""
-                    default = "--no-default--"
-
-                if isinstance(op, pd.core.base.PandasObject):
-                    op = "<pandas>"
-
-                elif repr(op) != repr(default):
-                    if param:
-                        header += f" {param}={repr(op)}"
-                    else:
-                        header += repr(op)
+        header = funcname(type(self)) + ":"
+        for i in range(len(self.operands)):
+            lines, header = self._tree_repr_operand(i, lines, header)
         lines = [header] + lines
         lines = [" " * indent + line for line in lines]
-
         return lines
 
-    def tree_repr(self):
-        return os.linesep.join(self._tree_repr_lines())
+    def tree_repr(self, collapse_fusion=True):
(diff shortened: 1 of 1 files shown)


dask_match/expr.py
```diff
@@ -112,38 +112,42 @@ def __str__(self):
     def __repr__(self):
         return str(self)
 
-    def _tree_repr_lines(self, indent=0):
-        header = funcname(type(self)) + ":"
+    def _tree_repr_operand(self, i, lines, header, collapse_fusion=True):
```
mrocklin, line 115: Can you say a bit more about the need to make the second method here?  

I briefly tried to implement Fused earlier, but gave up.  My approach at the time was to implement `_tree_repr_lines` on that class differently.  My preference would be if we could handle this with only one method, rather than have to split logic between two (this seems more complex to me).  Is that possible?

dask_match/expr.py
```diff
@@ -1107,6 +1111,32 @@ class Fused(Blockwise):
     def _meta(self):
         return self.exprs[0]._meta
 
+    def _tree_repr_lines(self, indent=0, collapse_fusion=True):
+        # Start with fused expressions
+        lines = []
+        root = self.exprs[0]
+        if collapse_fusion:
+            root_id = f"Fused with {len(self.exprs) - 1} expressions"
```
mrocklin, line 1119: I like the `Fused with` annotation you have now.  I agree that it's nice to distinguish these a little from the others.  I wonder if there is some more visually striking way to do this.  My first attempt at this tried to use `-` rather than ` ` as the indentation character (but only as far left as the fused object was) so it would look like this:

```
Sum
  Fused
  --Add
  ----Mul
  ------ df
```

Maybe that's too much though. Maybe just one dash?  Maybe dots `.`?  Just an idea, not a constraint.

mrocklin: Or maybe we include a vertical line

```
Sum
  Fused
  | Add
  |   Mul
  |     df
```

rjzamora: I went with the vertical-line idea since it seemed easier to read/test in my experiments. Thanks for the suggestion!

dask_match/tests/test_collection.py
```diff
@@ -323,6 +325,9 @@ def test_tree_repr(df):
     assert "None" not in s
     assert "skipna=False" in s
     assert str(df.seed) in s.lower()
+    if fuse:
+        assert "Fused" in s
+        assert s.count("|") == 9
```
mrocklin, line 330: Nit: This exact count seems a little brittle to me.  I could imagine changes to the code that changed this count but wouldn't necessarily mean an error (maybe we get more verbose in our number of terms).  

I'm going to go ahead and merge, mostly I mention this to align on style over time.

mrocklin (approved):

mrocklin:
Oooh, I like the example in the opening comment also.  I've put that into the commit message.

PR #49, opened 2023-04-22
Add private _depth method

mrocklin:


rjzamora (commented):
Seems reasonable to me. Can you leave a comment here about how you envision this method being used? I can imagine several uses, just curious what motivated the PR.

mrocklin:
Originally it was for testing.  I wanted to ensure that `read_parquet().head()` had only a single term, but then I forgot about grabbing just the last five elements (`BlockwiseHead` I think it's called).

mrocklin:
OK to merge?

PR #50, opened 2023-04-22
Enhance simplify with simplify_up, remove replacement rules

mrocklin:
Simplify is, I think, simpler and faster, but has the challenge that it only allows classes to optimize their subtree.  We use replacement rules today to optimize patterns that have to look above the term in the expression tree.

This PR renames simplify to `_simplify_down` and adds a new `_simplify_up` method that gets the parent term.  It then moves most of the replacement rules over to simplify.  I haven't yet moved over the parquet predicate pushdown rules, mosty because they're really complicated.  In general everything so far has felt simpler in this new method.

Additionally, I have a local commit which I'll push up separately which removes matchpy entirely.

### DIFF
--- a/dask_match/datasets.py
+++ b/dask_match/datasets.py
@@ -2,8 +2,6 @@ import functools
 
 import numpy as np
 import pandas as pd
-from matchpy import CustomConstraint, Pattern, ReplacementRule, Wildcard
-
 from dask.utils import random_state_data
 
 from dask_match.collection import new_collection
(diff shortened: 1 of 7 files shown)


mrocklin:
I've removed the parquet rules.  This is complete now if we want it.  It was a little painful to remove the parquet rules, but I think that it's ok.

mrocklin:
@rjzamora this is the larger of my PRs that could use review, if you have time today.

rjzamora:
These general changes make sense to me. I'll try to resolve merge conflicts.

dask_match/datasets.py
```diff
@@ -52,63 +50,48 @@ def _divisions(self):
...
-        else:
-            return random_state_data(size, self.seed)
+        return {
+            k: (
+                np.random.randint(2e9, size=npartitions)
+                if self.seed is None
+                else random_state_data(npartitions, self.seed)
+            )
+            for k in self.operand("dtypes")
+        }
```
rjzamora, line 60: Sorry - Looks like I got this seed stuff a bit wrong last time.

dask_match/expr.py
```diff
@@ -1003,7 +1016,7 @@ def optimize(expr: Expr, fuse: bool = True) -> Expr:
     matchpy
     optimize_blockwise_fusion
     """
-    expr, _ = simplify(expr)
+    expr = expr.simplify()
```
rjzamora, line 1019: Should we also add `FrameBase.simplify` so that we can avoid doing things like `df.optimize(fuse=False)` in tests/demos?

rjzamora (approved):
These changes make sense to me. Though, I didn't think too carefully about edge cases in the general simplify algorithm.

PR #51, opened 2023-04-22
Remove matchpy

mrocklin:
Builds off of #50 , currently here for motivation.  We need to remove the parquet + predicate pushdown rules first.  In general it'd be nice to remove this dependency.


mrocklin:
This works now

rjzamora:
I think I like the idea of dropping match-py if we can get what we need with variations of `simplify`. Might need a new name :)

That said, I'm not sure I understand how tests are passing here without match-py. I feel like I'm missing something?

mrocklin:
And yes, ideas for new names are welcome.

```python
import dask_hlx as dx
```

??

rjzamora:
>import dask_hlx as dx

I like `dask_hlx as dx`. Also like `dask_expr as dx`, or even just `daskx as dx`

mrocklin:
dask-expr sounds fun.

mrocklin:
I ran into this with flake8

    dask_match/datasets.py:44:6: B019 Use of `functools.lru_cache` or `functools.cache` on methods can lead to memory leaks. The cache may retain instance references, preventing garbage collection.

Did you also experience this?

rjzamora:
Woot

> Did you also experience this?

Ah, no - I've been running `black` locally, but not flake8

mrocklin:
I recommend using pre-commit

```
pip install pre-commit
pre-commit install 
```

Then it'll do these things automatically prior to committing.  I'm also adding a CI run to enforce this at https://github.com/mrocklin/dask-match/pull/54

PR #52, opened 2023-04-23
Remove some unnecessary methods

mrocklin:
These are handled by inheritance, but we had defined them anyway


mrocklin:
@rjzamora in case you want to review (this should be easy)

rjzamora (approved):

mrocklin:
Thanks for handling all of the merges

PR #54, opened 2023-04-25
Add pre-commit hooks CI run

mrocklin:


mrocklin:
I tried stealing what dask does, but it seems to be unhappy and I'm not sure why. I tried adding pyproject.toml in order to help installation, but that doesn't seem to suffice.  I'm going to ignore this for a moment and hope that someone else smarter than me comes by :)

Once this works, @rjzamora you'll likely see that there is a genuine issue in the shuffle code with an undefined variable name , and the lru_cache issue.  I can try to resolve these, but I suspect that you're better placed.

If you know anyone who can figure out pre-commit stuff that would be welcome too :)

PR #55, opened 2023-04-25
Enable partition filtering in shuffle expressions

rjzamora:
Ports `SimpleShuffle` and it's sub-classes to `PartitionsFiltered`.

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -1,0 +1,39 @@
+# flake8 doesn't support pyproject.toml yet https://github.com/PyCQA/flake8/issues/234
+[flake8]
+# References:
+# https://flake8.readthedocs.io/en/latest/user/configuration.html
+# https://flake8.readthedocs.io/en/latest/user/error-codes.html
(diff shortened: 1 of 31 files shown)


mrocklin:
I haven't deeply reviewed this (and may not).  I'm curious, how did it feel when using `PartitionsFiltered`?

rjzamora:
>I haven't deeply reviewed this (and may not). I'm curious, how did it feel when using PartitionsFiltered?

Using `PartitionsFiltered` was not particularly difficult in this case (we already needed to think through it once for HLGs).

rjzamora:
>I haven't deeply reviewed this (and may not).

I took a few passes over this PR and cleaned things up a bit.  There aren't any dramatic changes IMO, so I'll probably merge today if you don't mind.

mrocklin:
No objection.  At this stage I'm mostly going to be very picky about the structure of how we define algorithms, but pretty laissez faire about the algorithms themselves

PR #56, opened 2023-04-25
Add `pre-commit` CI job

jrbourbeau:
This gets `pre-commit` running in CI -- though there are some linting failures. @rjzamora @mrocklin feel free to push directly to this branch if you want to 

Supersedes https://github.com/mrocklin/dask-match/pull/54

### DIFF
--- a/.github/workflows/pre-commit.yml
+++ b/.github/workflows/pre-commit.yml
@@ -1,0 +1,18 @@
+name: Linting
+
+on:
+  push:
+    branches: main
+  pull_request:
+    branches: main
+
+jobs:
+  checks:
+    name: pre-commit hooks
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3.5.2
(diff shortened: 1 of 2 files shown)


rjzamora:
Thanks @jrbourbeau !

Any idea what is causing this error: `ValueError: Error code '#' supplied to 'ignore' option does not match '^[A-Z]{1,3}[0-9]{0,3}$'` ?  Doesn't seem to be related to a formatting problem in dask-expr.

jrbourbeau:
Moving to a separate `.flake8` file (like we have over in `dask/dask`) seems to do the trick

rjzamora:
Thanks @jrbourbeau ! I don't know much about the config files added here, but the changes seem reasonable to me (and CI is green). Do you think this is good to merge?

jrbourbeau (commented):
Yeah, I think this should be good to go. Can always follow-up if needed

PR #59, opened 2023-04-26
Fix mixed fusion and broadcasting

rjzamora:
While working on `Merge`, I found that we were incorrectly fusing `Blockwise` expressions with mismatched partition counts. We were also failing to broadcasts single-partition dependencies within `Fused` layers.

**Side Note**:
- This PR also adds a `clear_divisions` option to `MapPartitions` to make it possible to specify that the applied function should produce unknown divisions.


rjzamora (commented):
Note that we can always revisit the need for a `clear_divisions` operand in `MapPartitions`. The rest of this PR is a simple bug-fix.

PR #60, opened 2023-04-26
[Bug] Fix `pprint` for `Fused` expressions with unfused dependencies

rjzamora:
While adding a simple `Expr.visualize` method, I discovered that `pprint` indentation was a bit off for `Fused` expressions with unfused dependencies. This PR adds a simple fix.


rjzamora:
Note: I'll self-merge this since the change is minor.

PR #61, opened 2023-04-26
Add basic `Expr.visualize` method

rjzamora:
**Example**

```python
from dask_expr.datasets import timeseries

df = timeseries()
df["id2"] = df["id"] * 2
df2 = (df[df["id"] < 1000].shuffle("id2", backend="tasks")["x"]).partitions[0]
df2.simplify().visualize()
```

Screen Shot 2023-04-26 at 2 15 22 PM

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -1,0 +1,39 @@
+# flake8 doesn't support pyproject.toml yet https://github.com/PyCQA/flake8/issues/234
+[flake8]
+# References:
+# https://flake8.readthedocs.io/en/latest/user/configuration.html
+# https://flake8.readthedocs.io/en/latest/user/error-codes.html
+# https://pycodestyle.pycqa.org/en/latest/intro.html#error-codes
+exclude = __init__.py
+ignore =
+    # Extra space in brackets
+    E20
+    # Multiple spaces around ","
+    E231,E241
+    # Comments
+    E26
+    # Import formatting
+    E4
+    # Comparing types instead of isinstance
+    E721
+    # Assigning lambda expression
+    E731
+    # Ambiguous variable names
+    E741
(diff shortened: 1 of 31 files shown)


mrocklin:
Cool.  Quick thought, maybe we replace a term if it is repeated in the children?

So rather than ...

```
LT(Projection, 1000)
    Projection(...)
```

Maybe we have

```
LT( -- , 1000)
    Projection(...)
```

Just a thought, not a constraint

rjzamora:
>maybe we replace a term if it is repeated in the children?

I can understand the appeal of doing that. However, I actually find it useful for the label to include just enough information for the graph to be self explanatory, and using `" -- "` would not make something like `LT( -- , -- )` very clear.  In fact, I wonder if we should number the expressions, and use something like `" #2 "` instead of `" -- "`.

Thoughts on something like this?:

Screen Shot 2023-04-26 at 3 37 43 PM

rjzamora:
Or perhaps we go with something like `" -- "` (as you suggest) and rely on "tooltips" to show more-detailed information about dependencies.

mrocklin:
I think I like your original use of the operation names over numbers.  Happy to drop the `--` thought.

rjzamora:
>I think I like your original use of the operation names over numbers. Happy to drop the -- thought.

Okay. This PR is only mean to provide minimal `visualize` support, so I'm expecting us to revisit most of the details :)

rjzamora:
@mrocklin - Any objections on merging this as is with the knowledge that we will probably change things later?

mrocklin:
Do your thing.  Sorry to be MIA.  Lots of conference booth time this week.

PR #62, opened 2023-04-27
Add abstract merge API

rjzamora:
Adds `DataFrame.merge` and an abstract `Merge` expression.

Other changes in this PR:
- ~Fixes `Fused._tree_repr_lines` (again!)~ [**EDIT**: Moved to #64]
- Adds `dask_expr.shuffle.AssignPartitioningIndex` to simplify shuffle (replacing `Assign` + `PartitioningIndex`).

**TODO**:
- [x] `DataFrame.merge` docstring
- [x] Improve test coverage

### DIFF
--- a/dask_expr/expr.py
+++ b/dask_expr/expr.py
@@ -546,6 +546,9 @@ class MapPartitions(Blockwise):
         "kwargs",
     ]
 
+    def __str__(self):
+        return f"MapPartitions({funcname(self.func)})"
+
     def _broadcast_dep(self, dep: Expr):
         # Always broadcast single-partition dependencies in MapPartitions
         return dep.npartitions == 1
--- a/dask_expr/expr.py
+++ b/dask_expr/expr.py
@@ -64,13 +64,12 @@ class Expr:
     def __repr__(self):
         return str(self)
 
-    def _tree_repr_lines(self, indent=0, recurse=True):
+    def _tree_repr_lines(self, indent=0):
         header = funcname(type(self)) + ":"
         lines = []
         for i, op in enumerate(self.operands):
             if isinstance(op, Expr):
-                if recurse:
-                    lines.extend(op._tree_repr_lines(2))
+                lines.extend(op._tree_repr_lines(2))
             else:
                 try:
                     param = self._parameters[i]
@@ -1098,34 +1097,20 @@ class Fused(Blockwise):
     def _meta(self):
         return self.exprs[0]._meta
 
-    def _tree_repr_lines(self, indent=0, recurse=True):
-        header = f"Fused({self._name[-5:]}):"
-        if not recurse:
-            return [header]
-
-        seen = set()
+    def _tree_repr_lines(self, indent=0):
         lines = []
-        stack = [(self.exprs[0], 2)]
-        fused_group = [_expr._name for _expr in self.exprs]
-        dependencies = {dep._name: dep for dep in self.dependencies()}
-        while stack:
-            expr, _indent = stack.pop()
+        ext_indent = 2  # How far to indent "unfused" lines
+        header = f"Fused({self._name[-5:]}):"
+        for i, line in enumerate(self.exprs[0]._tree_repr_lines(2)):
+            if i < len(self.exprs):
+                lines.append(line.replace(" ", "|", 1))
+            else:
+                ext_indent = len(line) - len(line.lstrip(" "))
+                break
 
-            if expr._name in seen:
-                continue
-            seen.add(expr._name)
-
-            line = expr._tree_repr_lines(_indent, recurse=False)[0]
-            lines.append(line.replace(" ", "|", 1))
-            for dep in expr.dependencies():
-                if dep._name in fused_group:
-                    stack.append((dep, _indent + 2))
-                elif dep._name in dependencies:
-                    dependencies.pop(dep._name)
-                    lines.extend(dep._tree_repr_lines(_indent + 2))
-
-        for dep in dependencies.values():
-            lines.extend(dep._tree_repr_lines(2))
+        for op in self.operands[1:]:
+            if isinstance(op, Expr):
+                lines.extend(op._tree_repr_lines(ext_indent))
 
         lines = [header] + lines
         lines = [" " * indent + line for line in lines]
--- a/dask_expr/expr.py
+++ b/dask_expr/expr.py
@@ -64,12 +64,13 @@ class Expr:
     def __repr__(self):
         return str(self)
 
-    def _tree_repr_lines(self, indent=0):
+    def _tree_repr_lines(self, indent=0, recurse=True):
         header = funcname(type(self)) + ":"
         lines = []
         for i, op in enumerate(self.operands):
             if isinstance(op, Expr):
-                lines.extend(op._tree_repr_lines(2))
+                if recurse:
+                    lines.extend(op._tree_repr_lines(2))
             else:
                 try:
                     param = self._parameters[i]
@@ -861,7 +862,7 @@ class Partitions(Expr):
 
     def _simplify_down(self):
         if isinstance(self.frame, Blockwise) and not isinstance(
-            self.frame, BlockwiseIO
+            self.frame, (BlockwiseIO, Fused)
         ):
             operands = [
                 Partitions(op, self.partitions)
@@ -1097,20 +1098,34 @@ class Fused(Blockwise):
     def _meta(self):
         return self.exprs[0]._meta
 
-    def _tree_repr_lines(self, indent=0):
-        lines = []
-        ext_indent = 2  # How far to indent "unfused" lines
+    def _tree_repr_lines(self, indent=0, recurse=True):
         header = f"Fused({self._name[-5:]}):"
-        for i, line in enumerate(self.exprs[0]._tree_repr_lines(2)):
-            if i < len(self.exprs):
-                lines.append(line.replace(" ", "|", 1))
-            else:
-                ext_indent = len(line) - len(line.lstrip(" "))
-                break
+        if not recurse:
+            return [header]
 
-        for op in self.operands[1:]:
-            if isinstance(op, Expr):
-                lines.extend(op._tree_repr_lines(ext_indent))
+        seen = set()
+        lines = []
+        stack = [(self.exprs[0], 2)]
+        fused_group = [_expr._name for _expr in self.exprs]
+        dependencies = {dep._name: dep for dep in self.dependencies()}
+        while stack:
+            expr, _indent = stack.pop()
+
+            if expr._name in seen:
+                continue
+            seen.add(expr._name)
+
+            line = expr._tree_repr_lines(_indent, recurse=False)[0]
+            lines.append(line.replace(" ", "|", 1))
+            for dep in expr.dependencies():
+                if dep._name in fused_group:
+                    stack.append((dep, _indent + 2))
+                elif dep._name in dependencies:
+                    dependencies.pop(dep._name)
+                    lines.extend(dep._tree_repr_lines(_indent + 2))
+
+        for dep in dependencies.values():
+            lines.extend(dep._tree_repr_lines(2))
 
         lines = [header] + lines
         lines = [" " * indent + line for line in lines]
--- a/dask_expr/expr.py
+++ b/dask_expr/expr.py
@@ -785,9 +785,6 @@ class Projection(Elemwise):
         return f"{base}[{repr(self.columns)}]"
 
     def _simplify_down(self):
-        from dask_expr.merge import Merge
-        from dask_expr.shuffle import Shuffle
-
         if isinstance(self.frame, Projection):
             # df[a][b]
             a = self.frame.operand("columns")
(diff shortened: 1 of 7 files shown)


dask_expr/merge.py
```diff
@@ -0,0 +1,226 @@
...
+        "left_index",
+        "right_index",
+        "suffixes",
+        "indicator",
+        "backend",
+        "shuffle_backend",
+    ]
+    _defaults = {
+        "how": "inner",
+        "on": None,
```
mrocklin, line 31: I would expect `on` to not be a keyword.  I would expect it to be normalized to `left_on` and `right_on` by the user API `merge` function.  In general I'd want us to normalize all inputs we can before making an expression.

dask_expr/merge.py
```diff
@@ -0,0 +1,226 @@
...
+        how = expr.how
+        on = expr.on
+        left_on = expr.left_on
+        right_on = expr.right_on
+        left_index = expr.left_index
+        right_index = expr.right_index
+
+        for o in [on, left_on, right_on]:
+            if isinstance(o, Expr):
+                raise NotImplementedError()
```
mrocklin, line 121: Does this ever happen in practice?

dask_expr/merge.py
```diff
@@ -0,0 +1,226 @@
...
+            and not left_index
+            and not right_index
+        ):
+            on = [c for c in left.columns if c in right.columns]
+            if not on:
+                left_index = right_index = True
+
+        if on and not left_on and not right_on:
+            left_on = right_on = on
+            on = None
```
mrocklin, line 135: I would hope that this stuff could happen in the user API before we ever got to expressions.  Is that possible?

dask_expr/merge.py
```diff
@@ -0,0 +1,226 @@
...
+
+        supported_how = ("left", "right", "outer", "inner")
+        if how not in supported_how:
+            raise ValueError(
+                f"dask.dataframe.merge does not support how='{how}'."
+                f"Options are: {supported_how}."
+            )
```
mrocklin, line 142: Also in user API maybe?

dask_expr/merge.py
```diff
@@ -0,0 +1,226 @@
...
+        # Check for "trivial" broadcast (single partition)
+        npartitions = max(left.npartitions, right.npartitions)
+        if (
+            npartitions == 1
+            or left.npartitions == 1
+            and how in ("right", "inner")
+            or right.npartitions == 1
+            and how in ("left", "inner")
+        ):
+            return BlockwiseMerge(left, right, **expr._kwargs)
```
mrocklin, line 153: I would expect this (and general algorithm selection) to be in `Merge.simplify_down`.  Thoughts?

dask_expr/merge.py
```diff
@@ -0,0 +1,226 @@
...
+            # Shuffle right
+            right = Shuffle(
+                right,
+                shuffle_right_on,
+                npartitions_out=npartitions,
+                backend=expr.shuffle_backend,
+            )
+
+        # Blockwise merge
+        return BlockwiseMerge(left, right, **expr._kwargs)
```
mrocklin, line 204: This also seems like it should be in `Merge.simplify_down` to me

dask_expr/merge.py
```diff
@@ -0,0 +1,226 @@
...
+        return dep.npartitions == 1
+
+    @staticmethod
+    def _merge_partition(df, other, **kwargs):
+        return df.merge(other, **kwargs)
+
+    def _task(self, index: int):
+        return (
+            apply,
+            self._merge_partition,
```
mrocklin, line 220: Maybe replace this with `M.merge` instead?

dask_expr/shuffle.py
```diff
@@ -429,39 +434,39 @@ def _is_column_label_reference(expr, key):
...
         and key == index_name
-        and key not in getattr(expr, "columns", ())
+        and key not in getattr(df, "columns", ())
     )
 
 
-class PartitioningIndex(Blockwise):
-    """Create a partitioning index
+class AssignPartitioningIndex(Blockwise):
+    """Assign a partitioning index
```
mrocklin, line 469: +1 on this name change

mrocklin (commented):
Exciting to see!  Some questions about where to put what logic below.

dask_expr/merge.py
rjzamora: Yes, I think we can do this. In general, I agree with the desire to rely on normalization outside of the `Expr` and to keep the logic within the `Expr` relatively "clean."  However, it is important to note that `Expr`s will **often** be initialized by other `Expr`s within `simplify`. Therefore, doing input validation in the collection API does not guarantee that an `Expr` will always get normalized inputs (there can easily be a bug in another `Expr`'s `_simplify_down`/`up` implementation.

dask_expr/merge.py
rjzamora: This was translated from `dask.dataframe` code, and doesn't make the most sense here. I think we just want to raise an error if the user passes in a collection for `on`/`left_on`/`right_on` in the collection API.

dask_expr/merge.py
rjzamora: Yes - In this case we can.

dask_expr/merge.py
mrocklin: > However, it is important to note that Exprs will often be initialized by other Exprs within simplify

Understood.  I think that the principle here is "normalize inputs as early as possible".

dask_expr/merge.py
mrocklin: I'm open to changes like this in the future.  I don't think that we're closing the door on these things by putting everything in simplify_down.  

If we're going to make a new protocol or way of doing things then I'm going to be pretty strict that it's the correct choice for the future.  If we want to reuse an existing abstraction then I'm going to be pretty relaxed.  I completely agree that in the future we may want to separate out different kinds of optimization/simplification passes.  I don't think that we have that need today, and so I'd like to keep everything in as few protocols as possible.  Once we find a need for separating things out then we'll have, I think, more information with a concrete use case to help us do this separation.

mrocklin: > Users should be able to use custom logic to convert abstract expressions (like Shuffle and Merge) to other expressions. Therefore, the user needs a way to pass in this custom logic.

Also, maybe I don't have full context here.  Are there examples that you need soon like this?  Maybe for rapids work that I'm unaware of?

rjzamora: >Are there examples that you need soon like this?

I'm not sure how "soon", but we absolutely want it to be easy to plug in explicit-comms shuffle/merge algorithms. We'd also like to explore custom p2p algorithms with GPU-direct storage.  I'd like to be able to just define a simple  `Merge`/`Shuffle` class that defines the necessary logic to lower from the abstract class.

mrocklin: OK, good to know.  The approch here though doesn't provide that kind of registration ability though today, right?  

My sense is that the MergeBackend class today is unnecessary.  And that this logic is fundamentally the same as `Merge.simplify_down`.  If we want to add some ability to register new algorithms I'm happy with that.  I don't see that this class with this method helps us achieve that though.  

I still think that we should just do the dumb thing for now and put this in simplify_down, and then, once we have some concrete case to work with, think about how to add some kind of dispatch / registration logic.  I suspect that we'll find a simpler solution at that time.

mrocklin: At the time, it may also just make sense to ask if the workers have a GPU and then import from RAPIDS directly in this code too, right?  That seems like it might be a lot simpler.

rjzamora: >The approch here though doesn't provide that kind of registration ability though today, right?

Right now, one could define a `MergeBackend` class with a custom `from_abstract_merge` method, and pass it in with `backend=...`. In the future, we'd also want to implement a registration mechanism to change defaults and/or expose the algorithm in a config field.

>My sense is that the MergeBackend class today is unnecessary.

Agree that it isn't necessary yet, and even in the long run, `Merge` could just expose a `from_abstract_merge` or `lower_abstract_merge` class method.

>I still think that we should just do the dumb thing for now and put this in simplify_down, and then, once we have some concrete case to work with, think about how to add some kind of dispatch / registration logic. I suspect that we'll find a simpler solution at that time.

This is fine with me. Just being transparent that I think putting everything in `simplify_down` will probably get in the way pretty soon.

rjzamora:
Note that I spent a bunch of time trying to push column projections through `Shuffle` expressions, but ultimately decided that I'll need to wait until "lowering" logic can be distinguished from `_simplify_down`. Once we can do that, this optimization becomes trivial. With the optimizations combined, annoying details like the temporary partitioning-index makes the necessary logic much messier than it needs to be.  In other words, I'm feeling like putting everything in `_simplify_down` is officially "in the way."

dask_expr/merge.py
```diff
@@ -0,0 +1,177 @@
...
+        ) and left.known_divisions
+        merge_indexed_right = (
+            right_index or _contains_index_name(right, right_on)
+        ) and right.known_divisions
+
+        # NOTE: Merging on an index is fragile. Pandas behavior
+        # depends on the actual data, and so we cannot use `meta`
+        # to accurately predict the output columns. Once general
+        # partition statistics are available, it may make sense
+        # to drop support for left_on and right_on.
```
rjzamora, line 108: The only component of this PR that doesn't feel "ready" is that we aren't really testing the case that we want to merge on an index, but require a shuffle. However, while trying to add a test for this, I discovered that merging on an index is actually quite fragile in general :/

rjzamora: oops `left_on and right_on` should be `left_index and right_index` here.

dask_expr/merge.py
```diff
@@ -0,0 +1,177 @@
...
+                backend=shuffle_backend,
+            )
+
+        # Blockwise merge
+        return BlockwiseMerge(left, right, **self.kwargs)
+
+    def _simplify_down(self):
+        if type(self) == Merge:
+            # Only lower abstract objects
+            return self._lower()
```
mrocklin, line 159: Maybe inline this method?

mrocklin (commented):
Things here seem pretty simple to me.  What do you think?

rjzamora:
>Things here seem pretty simple to me. What do you think?

There are some unsupported corner cases here (e.g. merging on an unnamed index with unknown divisions).  However, things seem fine for an initial Merge API.

dask_expr/expr.py
```diff
@@ -796,6 +802,69 @@ def _simplify_down(self):
...
+                    or col in projection
+                    or f"{col}{right_suffix}" in projection
+                )
+            ]
+            if set(project_left) < set(left.columns) or set(project_right) < set(
+                right.columns
+            ):
+                return type(self.frame)(
+                    left[project_left], right[project_right], *self.frame.operands[2:]
+                )[self.operand("columns")]
```
mrocklin, line 866: I think that in the future we might consider putting these in simplify_up.  My thinking here is that Merge/Shuffle are more complex/specialized operations, and so the burden of defining optimization should be on them.  Otherwise I imagine that the `Projection.simplify_down` method will become very large.  Also probably most of this code is about the logic of shuffle/merge, rather than the logic of projection.  Maybe that keeps thinking/changes localized to single files.

Just a thought though, nothing immediate needed here.  I'm just testing out ideas for a future convention about where to put simplification logic.

dask_expr/shuffle.py
```diff
@@ -127,7 +132,7 @@ def from_abstract_shuffle(cls, expr: Shuffle) -> Expr:
         partitioning_index = expr.partitioning_index
         npartitions_out = expr.npartitions_out
         ignore_index = expr.ignore_index
-        options = expr.options
+        options = expr.options or {}
```
mrocklin, line 135: Nit: maybe normalize this ahead of time rather than here?  Maybe even in `_defaults` (although I know using a mutable default is sometimes worrisome).

dask_expr/merge.py
```diff
@@ -0,0 +1,173 @@
...
+from dask.dataframe.dispatch import make_meta, meta_nonempty
+from dask.utils import M, apply
+
+from dask_expr.expr import Blockwise, Expr
+from dask_expr.repartition import Repartition
+from dask_expr.shuffle import Shuffle, _contains_index_name
+
+
+class Merge(Expr):
+    """Abstract merge operation"""
```
mrocklin, line 12: Maybe instead 
```suggestion
    """Merge / join two dataframes 
    
    This is an abstract class.  It will be transformed into a concrete 
    implementation before graph construction.
    
    See Also
    --------
    BlockwiseMerge
    Shuffle
    """
```

dask_expr/merge.py
```diff
@@ -0,0 +1,173 @@
...
+                npartitions_out=npartitions,
+                backend=shuffle_backend,
+            )
+
+        # Blockwise merge
+        return BlockwiseMerge(left, right, **self.kwargs)
+
+
+class BlockwiseMerge(Merge, Blockwise):
+    """Blockwise merge class"""
```
mrocklin, line 156: This docstring seems noninformative.  

Totally ok for now while we're early, just calling it out as something that'll probably have to change before we think about bringing more folks on.  Probably as we get closer to thinking of this as a real project we'll want things to start looking more polished.

We're not there yet, but I think it's probably useful to start thinking about this.

mrocklin (approved):
Just some nits.  I'm happy to merge as-is.

mrocklin:
Fine if this a follow-up, but it looks like we need to think about the left/right renaming in merge

```python
import dask_expr as dx
df = dx.datasets.timeseries()
df.merge(df, on="id").x_x.simplify()
```

```python-traceback
Cell In[3], line 1
----> 1 df.merge(df, on="id").x_x.simplify()

File ~/workspace/dask-expr/dask_expr/collection.py:93, in FrameBase.simplify(self)
     92 def simplify(self):
---> 93     return new_collection(self.expr.simplify())

File ~/workspace/dask-expr/dask_expr/collection.py:501, in new_collection(expr)
    498 def new_collection(expr):
    499     """Create new collection from an expr"""
--> 501     meta = expr._meta
    502     if is_dataframe_like(meta):
    503         return DataFrame(expr)

File ~/mambaforge/envs/match/lib/python3.11/functools.py:1001, in cached_property.__get__(self, instance, owner)
    999 val = cache.get(self.attrname, _NOT_FOUND)
   1000 if val is _NOT_FOUND:
-> 1001     val = self.func(instance)
   1002     try:
   1003         cache[self.attrname] = val

File ~/workspace/dask-expr/dask_expr/expr.py:579, in Blockwise._meta(self)
    577 @functools.cached_property
    578 def _meta(self):
--> 579     return self.operation(
    580         *[arg._meta if isinstance(arg, Expr) else arg for arg in self.operands]
    581     )

File ~/mambaforge/envs/match/lib/python3.11/site-packages/pandas/core/frame.py:3760, in DataFrame.__getitem__(self, key)
   3758 if self.columns.nlevels > 1:
   3759     return self._getitem_multilevel(key)
-> 3760 indexer = self.columns.get_loc(key)
   3761 if is_integer(indexer):
   3762     indexer = [indexer]

File ~/mambaforge/envs/match/lib/python3.11/site-packages/pandas/core/indexes/base.py:3654, in Index.get_loc(self, key)
   3652     return self._engine.get_loc(casted_key)
   3653 except KeyError as err:
-> 3654     raise KeyError(key) from err
   3655 except TypeError:
   3656     # If we have a listlike key, _check_indexing_error will raise
   3657     #  InvalidIndexError. Otherwise we fall through and re-raise
   3658     #  the TypeError.
   3659     self._check_indexing_error(key)

KeyError: 'x_x'
```

mrocklin:
Also for context:

```python
In [4]: df.merge(df, on="id").columns
Out[4]: Index(['name_x', 'id', 'x_x', 'y_x', 'name_y', 'x_y', 'y_y'], dtype='object')
```

rjzamora:
>Fine if this a follow-up, but it looks like we need to think about the left/right renaming in merge

Oops, that was a column-projection bug. Should be fixed/tested now.

mrocklin:
🎉

PR #64, opened 2023-04-28
Add `recursive` option to `_tree_repr_lines`

rjzamora:
Follow-up/correction to #60

After already attempting to rewrite the `Fused._tree_repr_lines` method twice, I discovered that the output is still incorrect in some cases. For example:

```python
from dask_expr.datasets import timeseries

df = timeseries()
df = df[df["id"] < 1000]["id"] + df["x"].sum()
df.optimize().pprint()
```

**Current Output**:
```

Fused(23ea4):
| Add:
|   Filter:
|     Projection: columns='id'
|       Timeseries: dtypes={'id': <class 'int'>} seed=1297281255
|     LT: right=1000
        Sum:
          Fused(4d36a):
          | Projection: columns='x'
          |   Timeseries: dtypes={'x': <class 'float'>} seed=1297281255
```

There are many things wrong with this, but one example is that the `LT: right=1000` dependencies should not include a `Sum`. As far as I can tell, the only way to properly build the `pprint` output for `Fused` is to make the `recursive` nature of `_tree_repr_lines` optional, and to manually iterate over the fused `Expr` graph to make sure we can properly distinguish between fused and un-fused dependencies. Doing so gives us...

**NEW (Correct) Output**:
```
Fused(fc998):
| Add:
    Sum:
      Fused(dfc62):
      | Projection: columns='x'
      |   Timeseries: dtypes={'x': <class 'float'>} seed=160437347
|   Filter:
|     LT: right=1000
|       Projection: columns='id'
|         Timeseries: dtypes={'id': <class 'int'>} seed=160437347
```

(which **finally** seems correct to me)

### DIFF
--- a/dask_expr/expr.py
+++ b/dask_expr/expr.py
@@ -64,12 +64,13 @@ class Expr:
     def __repr__(self):
         return str(self)
 
-    def _tree_repr_lines(self, indent=0):
+    def _tree_repr_lines(self, indent=0, recursive=True):
         header = funcname(type(self)) + ":"
         lines = []
         for i, op in enumerate(self.operands):
             if isinstance(op, Expr):
-                lines.extend(op._tree_repr_lines(2))
+                if recursive:
+                    lines.extend(op._tree_repr_lines(2))
             else:
                 try:
                     param = self._parameters[i]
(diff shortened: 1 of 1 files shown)


rjzamora (commented):
I'm sure this can be cleaned up a bit, but I'm going to move ahead with this since I'm quite uncomfortable with the existing version in main.

PR #65, opened 2023-05-03
[WIP] Use abstract `MapReduce` expressions for reductions

rjzamora:
**NOTE**: I feel it is important that we adjust the ACA/Reduction foundation a bit before building a `Groupby` API on top of it. I'd like for all reductions to naturally decompose into distinct "map" and "reduce" expressions to make it easier to (1) achieve task fusion, and (2) replace the reduction phase with alternative algorithms (like shuffling). With that said, the specific design used in this PR is completely experimental. I'm expecting us to significantly change the class naming and organization.

### DIFF
--- a/dask_expr/reductions.py
+++ b/dask_expr/reductions.py
@@ -39,6 +39,11 @@ class ApplyConcatApply(Expr):
     combine_kwargs = {}
     aggregate_kwargs = {}
 
+    @property
+    def args(self):
+        # Return unnamed operands
+        return self.operands[len(self._parameters) :]
+
     def __dask_postcompute__(self):
         return toolz.first, ()
 
@@ -61,15 +66,16 @@ class ApplyConcatApply(Expr):
             combine = aggregate
             combine_kwargs = aggregate_kwargs
 
-        d = {}
-        keys = self.frame.__dask_keys__()
-
         # apply chunk to every input partition
-        for i, key in enumerate(keys):
+        d = {}
+        for i in range(self.frame.npartitions):
+            args = [(self.frame._name, i)] + [
+                (arg._name, i) if isinstance(arg, Expr) else arg for arg in self.args
+            ]
             if chunk_kwargs:
-                d[self._name, 0, i] = (apply, chunk, [key], chunk_kwargs)
+                d[self._name, 0, i] = (apply, chunk, args, chunk_kwargs)
             else:
-                d[self._name, 0, i] = (chunk, key)
+                d[self._name, 0, i] = (chunk, *args)
 
         keys = list(d)
         j = 1
@@ -97,9 +103,15 @@ class ApplyConcatApply(Expr):
     @property
     def _meta(self):
         meta = meta_nonempty(self.frame._meta)
-        meta = self.chunk(meta, **self.chunk_kwargs)
-        meta = self.combine([meta], **self.combine_kwargs)
-        meta = self.aggregate([meta], **self.aggregate_kwargs)
+        args = [
+            meta_nonempty(arg._meta) if isinstance(arg, Expr) else arg
+            for arg in self.args
+        ]
+        meta = self.chunk(meta, *args, **self.chunk_kwargs)
+        aggregate = self.aggregate or (lambda x: x)
+        combine = self.combine or aggregate
+        meta = combine([meta], **self.combine_kwargs)
+        meta = aggregate([meta], **self.aggregate_kwargs)
         return make_meta(meta)
 
     def _divisions(self):
@@ -121,12 +133,6 @@ class Reduction(ApplyConcatApply):
     methods.
     """
 
-    _defaults = {
-        "skipna": True,
-        "numeric_only": None,
-        "min_count": 0,
-        "dropna": True,
-    }
     reduction_chunk = None
     reduction_combine = None
     reduction_aggregate = None
@@ -170,6 +176,7 @@ class Reduction(ApplyConcatApply):
 
 class Sum(Reduction):
     _parameters = ["frame", "skipna", "numeric_only", "min_count"]
+    _defaults = {"skipna": True, "numeric_only": None, "min_count": 0}
     reduction_chunk = M.sum
 
     @property
@@ -190,13 +197,15 @@ class Sum(Reduction):
 
 
 class Max(Reduction):
-    _parameters = ["frame", "skipna"]
+    _parameters = ["frame", "skipna", "numeric_only"]
+    _defaults = {"skipna": True, "numeric_only": None}
     reduction_chunk = M.max
 
     @property
     def chunk_kwargs(self):
         return dict(
             skipna=self.skipna,
+            numeric_only=self.numeric_only,
         )
 
     def _simplify_up(self, parent):
@@ -244,6 +253,7 @@ class Mean(Reduction):
 
 class Count(Reduction):
     _parameters = ["frame", "numeric_only"]
+    _defaults = {"numeric_only": None}
     split_every = 16
     reduction_chunk = M.count
 
--- a/dask_expr/reductions.py
+++ b/dask_expr/reductions.py
@@ -24,10 +24,10 @@ class MapReduce(Expr):
     This pattern is commonly used for reductions, groupby-aggregations, and
     more. A `MapReduce` subclass must define the following attributes:
 
-    -   `map`: An expression class to use for the initial partition-wise
+    -   `map`: A `Map`-based class to use for the initial partition-wise
         stage of the map-reduce. This should be a `Blockwise` expression
         if possible.
-    -   `reduce`: An expression class to use for the reduction stage of
+    -   `reduce`: A `Reduce`-based class to use for the reduction stage of
         the map-reduce. This is typically a tree-reduction.
 
     See Also
@@ -59,11 +59,11 @@ class MapReduce(Expr):
 class Map(Blockwise):
     """Perform the partition-wise 'map' stage of a map-reduce
 
-    This is commonly used for reductions, groupby-aggregations, and
-    more.  It requires a `chunk` method to be implemented, which
-    should correspond to a function to be applied directly to each
-    input partition. This function should be easy to serialize, and
-    can take in keyword arguments defined in `chunk_kwargs`.
+    This class is used by `MapReduce`. It requires a `chunk` method
+    to be implemented, which should correspond to a function to be
+    applied directly to each input partition. This function should
+    be easy to serialize, and can take in keyword arguments defined
+    in `chunk_kwargs`.
 
     See Also
     --------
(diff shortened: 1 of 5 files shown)


rjzamora:
**Example**

```python
import dask_expr as dx

df = dx.datasets.timeseries()
out = (df.x + 1).sum(skipna=False)
out.pprint()
```
An un-optimized reduction expression looks the same as before:
```
Sum: skipna=False
  Add: right=1
    Projection: columns='x'
      Timeseries: seed=254559780
```
```python
out.simplify().pprint()
```
A "simplified" reduction expression breaks into separate `Map` and `Reduce` expressions:
```
SumReduce: skipna=False
  SumMap: skipna=False
    Add: right=1
      Projection: columns='x'
        Timeseries: dtypes={'x': <class 'float'>} seed=254559780
```
```python
out.optimize().pprint()
```
After fusion, the `Map` expression can be fused with adjacent `Blockwise` expressions:
```
SumReduce: skipna=False
  Fused(d7a52):
  | SumMap: skipna=False
  |   Add: right=1
  |     Projection: columns='x'
  |       Timeseries: dtypes={'x': <class 'float'>} seed=254559780
```

mrocklin:
Cool.  Busy right now, but I'll hope to take a look this afternoon.

dask_expr/reductions.py
```diff
@@ -9,72 +11,141 @@
...
+
+    def _divisions(self):
+        return (None, None)
+
+    def _simplify_down(self):
+        mapped = self.map(*self.operands)
+        return self.reduce(mapped, *self.operands[1:])
+
+
+class Map(Blockwise):
```
mrocklin, line 59: Is this the same as `MapPartitions`?

I would expect `Map` to operate elementwise, not blockwise.  

I also expect us to need a `MapPartitions` operation (if we don't have one already).  Maybe that expression could do double duty?

mrocklin: Unless we're looking to be explicit about differentiating reduction layers, then maybe `ReductionBlockwise`, `ReductionCombine`, ... or something similar

dask_expr/reductions.py
```diff
@@ -94,49 +165,94 @@ def _layer(self):
...
-        return make_meta(meta)
 
-    def _divisions(self):
-        return [None, None]
+##
+## Simple class reductions
+##
+
+
+class DataFrameReduction(MapReduce):
```
mrocklin, line 174: Does this work on Series objects?  If so, we maybe want to avoid DataFrame in a name.

dask_expr/reductions.py
```diff
@@ -151,26 +267,86 @@ def aggregate(cls, inputs, **kwargs):
...
-        return toolz.first, ()
 
-    def _divisions(self):
-        return [None, None]
+##
+## Templates
+##
+
+
+class MapReduceTemplate:
```
mrocklin, line 276: On first reading I don't understand this class.  It confuses me.  

Maybe that's alright, but it's a sign that others will likely be confused as well.

dask_expr/reductions.py
```diff
@@ -180,52 +356,72 @@ def chunk_kwargs(self):
...
-class Max(Reduction):
-    _parameters = ["frame", "skipna"]
+class Sum(_Sum, DataFrameReduction):
+    map = _Sum.make_map()
+    reduce = _Sum.make_reduce()
+
+
+class _Max(DataFrameReductionTemplate):
+    _parameters = ["frame", "skipna", "numeric_only"]
+    _defaults = {"skipna": True, "numeric_only": None}
```
mrocklin, line 371: This looks like it's making things really complex?  For every reduction we now need to define two classes?

mrocklin (commented):
Some comments and questions.  In general this feels pretty complex to me.  My sense is that we haven't found the right abstraction here.  

I'm not sure I agree with the general goal of wanting map and reduce operations, nor with the goal of fusion.  I think I do agree with the goal of supporting split-out or shuffling when doing these.  My guess is that there are simpler ways to achieve that goal.  I can try to think about this more if that would be helpful.

dask_expr/reductions.py
```diff
@@ -151,26 +267,86 @@ def aggregate(cls, inputs, **kwargs):
...
+        class MyTemplate(MapReduce):
+            # Define common attributes
+            _parameters = [...]
+            ...
+
+        # Use the template to define an expression
+        class MyExpr(MyTemplate, MapReduce):
+            # Generate `map` and `reduce`
+            map = MyTemplate.make_map()
+            reduce = MyTemplate.make_reduce()
```
rjzamora, line 297: A key question for this PR is whether we *want* something like this "template" idea to make it possible to reduce the amount of code needed to implement a specific reduction. For example, without the template, `Sum` would need to look something like this:

```python
class SumMap(DataFrameMap):
    _parameters = ["frame", "skipna", "numeric_only", "min_count"]
    _defaults = {"skipna": True, "numeric_only": None, "min_count": 0}
    reduction_chunk = M.sum

    @property
    def chunk_kwargs(self):
        return dict(
            skipna=self.skipna,
            numeric_only=self.numeric_only,
            min_count=self.min_count,
        )

class SumReduce(DataFrameReduce):
    _parameters = ["frame", "skipna", "numeric_only", "min_count"]
    _defaults = {"skipna": True, "numeric_only": None, "min_count": 0}
    reduction_aggregate = M.sum

class Sum(DataFrameReduction):
    map = SumMap
    reduce = SumReduce

            return self.frame[parent.operand("columns")].sum(*self.operands[1:])
``` 

Which is only a bit more code than we need with the "template" mechanism:

```python
class _Sum(DataFrameReductionTemplate):
    _parameters = ["frame", "skipna", "numeric_only", "min_count"]
    _defaults = {"skipna": True, "numeric_only": None, "min_count": 0}
    reduction_chunk = M.sum
    reduction_aggregate = M.sum

    @property
    def chunk_kwargs(self):
        return dict(
            skipna=self.skipna,
            numeric_only=self.numeric_only,
            min_count=self.min_count,
        )

            return self.frame[parent.operand("columns")].sum(*self.operands[1:])


class Sum(_Sum, DataFrameReduction):
    map = _Sum.make_map()
    reduce = _Sum.make_reduce()
```

rjzamora: Given this comparison, my current preference is to simply define each `Map`/`Reduce` component explicitly (and remove the over-engineered template mechanism).

dask_expr/reductions.py
rjzamora: This was an experiment that I consider a failure. I shared some thoughts here: https://github.com/mrocklin/dask-expr/pull/65#discussion_r1183989786

dask_expr/reductions.py
mrocklin: I could also be wrong about this.  Maybe lots of people use shuffle groupbys for example.  I don't have evidence of that though.  I'm open to being educated otherwise.

dask_expr/reductions.py
```diff
@@ -9,72 +11,141 @@
...
+
+##
+## Base MapReduce classes
+##
 
 
-class ApplyConcatApply(Expr):
-    """Perform reduction-like operation on dataframes
+class MapReduce(Expr):
+    """Perform a map-reduction operation on a dataframe
```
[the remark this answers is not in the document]
rjzamora: Sure - I also don't consider the `MapReduce` name to be the right fit. However, the current `ApplyConcatApply` name is also wrong, since we aren't actually concatenating unless that logic is specified in `combine`/`aggregate`.

dask_expr/reductions.py
```diff
@@ -9,72 +11,141 @@
...
 
     This pattern is commonly used for reductions, groupby-aggregations, and
-    more.  It requires three methods to be implemented:
+    more. A `MapReduce` subclass must define the following attributes:
+
+    -   `map`: An expression class to use for the initial partition-wise
+        stage of the map-reduce. This should be a `Blockwise` expression
+        if possible.
+    -   `reduce`: An expression class to use for the reduction stage of
+        the map-reduce. This is typically a tree-reduction.
```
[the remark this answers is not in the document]
rjzamora: >I'm being kinda pedantic here, but I think that others who are used to this terminology will likely be even more pedantic.

Being pedantic is fine. I am in no way attached to any naming here.

dask_expr/reductions.py
mrocklin: An alternative implementation (which I'd suggest we delay) is that we create a ShuffleApplyConcatApply expression like the following:

```python
class ApplyConcatApply(Expr):
    ...

class ShuffleApplyConcatApply(ApplyConcatApply):
    def _layer(self):  # the only thing we need to add
```

And then we subclass all of the reductions (I would hope that we find something better) but with nothing extra (or maybe an extra split_out parameter, but that's all)

```python
class ShuffleSum(Sum, ShuffleApplyConcatApply):
    pass
```

I don't particularly like this either, but it does show us that we can add shuffling in in the future while still keeping the current Sum/Max/Min reductions easy to write and understand.  The thing that changes is `_layer`.  That's all that we should really care about I think.

Maybe we can even automate the creation of the `SumShuffle` expressions by generating them on demand.

In general, I find that I'm reacting strongly to anything that makes writing the Sum/Max/Min logic more complicated.  I think that we need it to be very easy to author and maintain algorithms.

rjzamora: I don't see us needing to support a shuffle-based aggregation for most reductions. Just operations like groupby and unique/drop_duplicates.

In the long run, I suppose I'm just fairly confident that we will want to be working with abstract ACA/Reduction expressions that can be converted into something like `chunk` + `shuffle` + `aggregate` or `chunk` + `repartition` + `aggregate` or just `chunk` + `aggregate`, depending on top-level arguments (like `split_out` etc..).

I'm not a fan of the current approach in this PR - It's just an experiment. I much preferred the simple approach of splitting something like into chunkwise and reduction phases by passing in the necessary chunk/combine/aggregate components as operands. That approach did not require any changes to the existing reductions, but resulted in hard-to-read `pprint` results after `simplify` was called.

mrocklin: > Just operations like groupby and unique/drop_duplicates

In this case then maybe the `class ShuffleSum(Sum, ShuffleApplyConcatApply)` approach works well.  We make `DropDuplicates(ApplyConcatApply)` and then make a variant algorithm that inherits from a shuffle algorithm.

```python
class DropDuplicatesShuffle(ShuffleACA, DropDuplicates):
    pass
```

rjzamora:
**Small Update**:  I seem to have what I want mostly working with minimal changes to the existing ACA/Reduction code by introducing the concept of a simple `ExprReference`. More specifically, I seem to be able to lower a reduction into distinct `Chunk` and `TreeReduce` classes, by passing them a reference to the original ACA/Reduction expression (to access combine/aggregate/kwargs).  Only problem I'm facing now is that the decomposed expression is somehow blocking column projection (not sure why).

rjzamora:
Closing in favor of #66 (which probably needs some work, but is significantly simpler).

PR #66, opened 2023-05-04
Use `ExprReference` to decompose reductions

rjzamora:
Possible alternative to #65

Uses a new/simple `ExprReference` mechanism to split `ApplyConcatApply` into distinct `Count` and `TreeReduce` expressions at `simplify` time. This approach does not require any changes to how `Reductions` (like sum/min/max) need to be implemented.

### DIFF
--- a/dask_expr/reductions.py
+++ b/dask_expr/reductions.py
@@ -68,6 +68,12 @@ class ApplyConcatApply(Expr):
     combine_kwargs = {}
     aggregate_kwargs = {}
 
+    @functools.cached_property
+    def split_out(self):
+        if "split_out" in self._parameters:
+            return self.operand("split_out")
+        return 1
+
     @property
     def _meta(self):
         meta = meta_nonempty(self.frame._meta)
@@ -82,8 +88,10 @@ class ApplyConcatApply(Expr):
     def _simplify_down(self):
         aca_ref = ExprReference(self)
         chunked = Chunk(self.frame, aca_ref)
-        reduced = TreeReduction(chunked, aca_ref)
-        return reduced
+        if self.split_out > 1:
+            raise NotImplementedError("split_out > 1 is not supported yet")
+        else:
+            return TreeReduction(chunked, aca_ref, self.split_every)
 
 
 class Chunk(Blockwise):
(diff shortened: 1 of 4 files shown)


dask_expr/reductions.py
```diff
@@ -39,41 +68,118 @@ class ApplyConcatApply(Expr):
...
+    @functools.cached_property
+    def _meta(self):
+        meta = meta_nonempty(self.frame._meta)
+        chunk_kwargs = self.chunk_kwargs or {}
+        return make_meta(self.chunk(meta, **chunk_kwargs))
+
+    def _task(self, index: int):
+        args = [self._blockwise_arg(self.frame, index)]
+        chunk_kwargs = self.chunk_kwargs or {}
+        return (apply, self.chunk, args, chunk_kwargs)
```
mrocklin, line 126: For ease of debugging task graphs I recommend the following:

```suggestion
        if self.chunk_kwargs:
            return (apply, self.chunk, args, self.chunk_kwargs)
        else:
            return (self.chunk,) + tuple(args)
```

Or maybe even bring args inside

mrocklin:
When we go to write an operation, groupby-mean for example, do we need this PR to be in, or are the chunk/aggregate/combine functions going to be independent of the tree-reduction/shuffle choice (I hope that they'll be independent).  If so, are we able to delay merging this PR in?  My hope is that the fact that we've found a solution that separates the tree-reduction/shuffle choice from the groupby-mean/max/count stuff allows us to go forward defining groupby operations knowing that we're future-proofed.

rjzamora:
>When we go to write an operation, groupby-mean for example, do we need this PR to be in, or are the chunk/aggregate/combine functions going to be independent of the tree-reduction/shuffle choice (I hope that they'll be independent). If so, are we able to delay merging this PR in? My hope is that the fact that we've found a solution that separates the tree-reduction/shuffle choice from the groupby-mean/max/count stuff allows us to go forward defining groupby operations knowing that we're future-proofed.

I *think* the answer to this question is yes. However, I'd like to take a bit of time this morning to add `split_out>1` support to make sure that my intuition is correct.  Once I have done that sanity check, I can try to implement `groupby` from `main` (without supporting `split_out>1` or changing any `_layer` logic).

rjzamora:
>I'd also be fine if split_out triggered the use of a shuffle groupby all the time.

I do think this is a good default (especially for pandas-backed data).

>I'm also curious about your motivations here in particular, do you end up running into split_out situations frequently with other work that you do? It seems you and I care differently about this topic. I'm guessing that that's because the recommender system work that you do relies on this a lot. Is that the case?

There are certainly many workflows requiring an aggregation to produce multiple partitions in the RAPIDS/RecSys world. However, whether or not a shuffle or tree-based algorithm is optimal is pretty case-dependent. At this point, the `split_out` question is just another "example" of why it is convenient to have clear mechanisms for decomposing general/abstract expressions.

I do appreciate your interest in not wasting time on details that are not technically "blocking" our high-priority goals (e.g. groupby). I think you are asking the right questions.

rjzamora:
This is quite stale at this point. Closing in favor of https://github.com/dask-contrib/dask-expr/pull/232

PR #74, opened 2023-05-09
Run CI against multiple Python versions

jrbourbeau:
I think the failures in https://github.com/coiled/benchmarks/pull/837 are related to lack of Python 3.9 support. Giving that a try here.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -18,6 +18,11 @@ defaults:
 jobs:
   test:
     runs-on: ubuntu-latest
+    strategy:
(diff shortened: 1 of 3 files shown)


jrbourbeau:
Okay, seeing the same `TypeError: unsupported operand type(s) for |: 'type' and 'type'` error here

jrbourbeau:
Alright, that was an easy fix. This should be ready for review whenever folks get a moment

jrbourbeau:
Thanks for reviewing! I can't push the green button here -- would you mind doing the honors?

PR #75, opened 2023-05-09
Update package finding

jrbourbeau:
Fixes the `E   ModuleNotFoundError: No module named 'dask_expr.io'` error seen over in https://github.com/coiled/benchmarks/pull/837 (see [this CI build](https://github.com/coiled/benchmarks/actions/runs/4928923042/jobs/8807903176?pr=837)). Note this update is based on what we're doing over in `dask/dask` and `dask/distributed` 

cc @rjzamora

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -31,3 +31,7 @@ dependencies = [
 ]
 
 dynamic = ["version"]
+
+[tool.setuptools.packages.find]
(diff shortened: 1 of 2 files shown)


mrocklin:
I don't know what any of this means, but merging anyway!

Also @jrbourbeau you now have write privileges.

PR #76, opened 2023-05-09
Add `isna`

jrbourbeau:
This looked like an easy thing to try and add when looking at failures over in https://github.com/coiled/benchmarks/pull/837. It took about 5 minutes minutes to figure out what to do and write a test (kudos on making things modular). Let me know if I missed something or if this is just totally wrong

### DIFF
--- a/dask_expr/tests/test_collection.py
+++ b/dask_expr/tests/test_collection.py
@@ -123,6 +123,8 @@ def test_conditionals(func, pdf, df):
         lambda df: df.apply(lambda row, x, y=10: row * x + y, x=2),
         lambda df: df[df.x > 5],
         lambda df: df.assign(a=df.x + df.y, b=df.x - df.y),
+        lambda df: df.isna(),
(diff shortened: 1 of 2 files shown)


dask_expr/tests/test_collection.py
```diff
@@ -123,6 +123,8 @@ def test_conditionals(func, pdf, df):
         lambda df: df.apply(lambda row, x, y=10: row * x + y, x=2),
         lambda df: df[df.x > 5],
         lambda df: df.assign(a=df.x + df.y, b=df.x - df.y),
+        lambda df: df.isna(),
+        lambda df: df.x.isna(),
```
rjzamora, line 127: We may want `pdf` to have at least one null for features like this. Or perhaps we have a `pdf_nulls` fixture?

rjzamora (approved):
Thanks @jrbourbeau - We can add null testing in a follow-up.

PR #78, opened 2023-05-09
Fix `Projection._meta`

rjzamora:
- Fixes `Projection._meta` when the operand is a series-like expression
- Also modifies `_tree_repr_lines` to account for the fact that the underlying data may be something other than pandas (e.g. cudf)


mrocklin:
Yup.  Sorry for the delay in review

PR #79, opened 2023-05-11
Revise predicate-pushdown behavior in `ReadParquet`

rjzamora:
The current implementation of `Parquet._simplify_up` works for some simple cases, but it does not enforce proper disjunctive-normal form (DNF) using the `List[List[Tuple]]` convention in `dask.dataframe`.

This PR introduces an overhaul of the logic used to add new filters to an existing `ReadParquet` expression.

### DIFF
--- a/dask_expr/io/parquet.py
+++ b/dask_expr/io/parquet.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import itertools
 import operator
 from functools import cached_property
 
@@ -12,7 +13,7 @@ from dask.dataframe.io.parquet.core import (
 from dask.dataframe.io.parquet.utils import _split_user_options
 from dask.utils import natural_sort_key
 
-from dask_expr.expr import EQ, GE, GT, LE, LT, NE, Expr, Filter, Projection
+from dask_expr.expr import AND, EQ, GE, GT, LE, LT, NE, OR, Expr, Filter, Projection
 from dask_expr.io import BlockwiseIO, PartitionsFiltered
 
 NONE_LABEL = "__null_dask_index__"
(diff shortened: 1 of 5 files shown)


dask_expr/io/parquet.py
```diff
@@ -265,3 +283,82 @@ def _filtered_task(self, index: int):
...
+
+class And(frozenset):
+    """Helper class for filter conjunctions"""
+
+    def to_list_tuple(self):
+        # NDF "and" is List[Tuple]
+        return tuple(
+            val.to_list_tuple() if hasattr(val, "to_list_tuple") else val
+            for val in self
+        )
```
mrocklin, line 319: These classes confuse me a little.  They're very similarly named to AND and OR above, and do very little I think.  My programming brain wants to find some way to remove them.  

However, I'm also happy to avoid that work and I'll care much less about them if we're able to isolate these classes to the `to_dnf` function (defined) and so make them very private.  Is that doable?

rjzamora: Yes, I was thinking something similar. These classes are a very useful way to make the DNF logic easier to think through, but but it seems critical that these "helper" classes don't get confused with proper expressions.

When I take another pass over this, I'll try to refactor things a bit to make the distinction clearer.

mrocklin: If we move them within the `to_dnf` function then I'll stop caring entirely about them :)

rjzamora: It was difficult to move everything into a function, but I was able to isolate everything within a general `_DNF` utility class.

rjzamora:
I'm going to merge this since it is mostly isolated from other code, and I'd like to work on some other things that are partially blocked by this.  Happy to revisit the logic/design later if needed.

PR #84, opened 2023-05-16
Add `Expr.statistics` method based on `Statistics`

rjzamora:
Alternate take on https://github.com/mrocklin/dask-expr/pull/40

- Adds `dask_expr.statistics` module (defining a simple `dataclass` structure for statistics)
- Adds `Expr.statistics` (which utilizes `dask_expr.statistics`)
- Adds `FrameBase.__len__` (which utilizes `Expr.statistics` when possible)

**Rational for requiring `Expr.statistics: dict` to be `dask_expr.statistics.Statistics` instances**: It think we will be wanting to add/leverage many different kinds of "statistics." Therefore, I think we will need a class structure like this to simplify and isolate the logic that dictates if/how a specific `Expr` "parent" can assume a specific `Statistics` object from its child.

### DIFF
--- a/dask_expr/statistics.py
+++ b/dask_expr/statistics.py
@@ -24,8 +24,8 @@ class Statistics:
         """Statistics that a "parent" Expr may assume
 
         A return value of `None` (the default) means that
-        `type(Expr)` is not eligable to assume these kind
-        of statistics.
+        `parent` is not eligable to assume this kind of
+        statistics.
         """
         return None
 
--- a/dask_expr/statistics.py
+++ b/dask_expr/statistics.py
@@ -10,7 +10,7 @@ from dask_expr.expr import Elemwise, Expr, Partitions
 
 @dataclass(frozen=True)
 class Statistics:
-    """Abstract expression-statistics class
+    """Abstract class for expression statistics
 
     See Also
     --------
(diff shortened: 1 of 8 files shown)


rjzamora:
I'm interested to know your thoughts on the general approach here @mrocklin

dask_expr/statistics.py
```diff
@@ -0,0 +1,74 @@
+from __future__ import annotations
```
mrocklin, line 1: I could use some help understanding this module.  

I think that in general we have yet to define what kinds of statistics we're going to capture, and how we plan to encode those.  There are lots of options here.  

I _think_ what I'm seeing here is that your response is "we'll just make different classes for all the different kinds of things that people might want to encode".  Is that correct?  If so, I'm not totally bought into this just yet.  

I think that the question of "how do we encode dataframe-level or partition-level statistics" is a big open one.  I'm ok with us not having a clear answer on this before we move forward, but I want the level of sophistication of our solution to be correlated with our confidence.  This feels like a somewhat sophisticated/specific solution (a few classes with some specific method APIs) but I don't have confidence that it's correct (or at least I don't know enough to be confident).  Can you help me understand here?

PR #85, opened 2023-05-17
Add ``any`` and ``all`` implementations

phofl:
Added any and all

### DIFF
--- a/dask_expr/reductions.py
+++ b/dask_expr/reductions.py
@@ -225,6 +225,36 @@ class Max(Reduction):
             return self.frame[parent.operand("columns")].max(skipna=self.skipna)
 
 
+class Any(Reduction):
+    _parameters = ["frame", "skipna"]
+    reduction_chunk = M.any
+
+    @property
+    def chunk_kwargs(self):
+        return dict(
+            skipna=self.skipna,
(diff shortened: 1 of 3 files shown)


dask_expr/reductions.py
```diff
@@ -225,6 +225,36 @@ def _simplify_up(self, parent):
...
+
+    @property
+    def chunk_kwargs(self):
+        return dict(
+            skipna=self.skipna,
+        )
+
+    def _simplify_up(self, parent):
+        if isinstance(parent, Projection):
+            return self.frame[parent.operand("columns")].any(skipna=self.skipna)
```
mrocklin, line 240: I wonder, is it possible to move this up to `Reduction`?  

(genuine question, the answer here might be "no")

mrocklin: It looks like the `Sum` implementation might be a bit more general.  I think that this simplification probably applies in most cases, maybe all?  My guess @phofl is that you're best placed actually to answer that question.

phofl: Yes we should be able to move this up, Wanted to look into this next. I guess that we have to generalise the implementations a bit, but should work

PR #86, opened 2023-05-17
Handle reduction to scalar when calculating Index

### LINKED ISSUE #81: Selecting single row after reduction raises RecursionError when checking the index in ``assert_eq``
```
import pandas as pd
from dask.dataframe import assert_eq

from dask_expr import from_pandas

df = pd.DataFrame({"a": [1, 2, 3], "bb": 1})
ddf = from_pandas(df)

assert_eq(ddf.max()["a"], df.max()["a"])
```

raises

```
  File "/Users/patrick/PycharmProjects/dask_dev/dask-expr/dask_expr/expr.py", line 134, in __getattr__
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/patrick/PycharmProjects/dask_dev/dask-expr/dask_expr/expr.py", line 841, in _meta
    return self.frame._meta.index
           ^^^^^^^^^^^^^^^^
  File "/Users/patrick/PycharmProjects/dask_dev/dask-expr/dask_expr/reductions.py", line 100, in _meta
    meta = self.chunk(meta, **self.chunk_kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/patrick/PycharmProjects/dask_dev/dask-expr/dask_expr/reductions.py", line 144, in chunk
    out = cls.reduction_chunk(df, **kwargs)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/patrick/mambaforge/envs/dask-expr/lib/python3.11/site-packages/dask/utils.py", line 1105, in __call__
    return getattr(__obj, self.method)(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/patrick/mambaforge/envs/dask-expr/lib/python3.11/site-packages/pandas/core/generic.py", line 11646, in max
    return NDFrame.max(self, axis, skipna, numeric_only, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/patrick/mambaforge/envs/dask-expr/lib/python3.11/site-packages/pandas/core/generic.py", line 11185, in max
    return self._stat_function(
           ^^^^^^^^^^^^^^^^^^^^
  File "/Users/patrick/mambaforge/envs/dask-expr/lib/python3.11/site-packages/pandas/core/generic.py", line 11158, in _stat_function
    return self._reduce(
           ^^^^^^^^^^^^^
  File "/Users/patrick/mambaforge/envs/dask-expr/lib/python3.11/site-packages/pandas/core/series.py", line 4666, in _reduce
    return op(delegate, skipna=skipna, **kwds)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/patrick/mambaforge/envs/dask-expr/lib/python3.11/site-packages/pandas/core/nanops.py", line 158, in f
    result = alt(values, axis=axis, skipna=skipna, **kwds)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/patrick/mambaforge/envs/dask-expr/lib/python3.11/site-packages/pandas/core/nanops.py", line 421, in new_func
    result = func(values, axis=axis, skipna=skipna, mask=mask, **kwargs)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/patrick/mambaforge/envs/dask-expr/lib/python3.11/site-packages/pandas/core/nanops.py", line 1083, in reduction
    values, mask, dtype, dtype_max, fill_value = _get_values(
                                                 ^^^^^^^^^^^^
  File "/Users/patrick/mambaforge/envs/dask-expr/lib/python3.11/site-packages/pandas/core/nanops.py", line 313, in _get_values
    values = extract_array(values, extract_numpy=True)  # type: ignore[assignment]
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/patrick/mambaforge/envs/dask-expr/lib/python3.11/site-packages/pandas/core/construction.py", line 441, in extract_array
    if isinstance(obj, (ABCIndex, ABCSeries)):
       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/patrick/mambaforge/envs/dask-expr/lib/python3.11/site-packages/pandas/core/dtypes/generic.py", line 44, in _instancecheck
    return _check(inst) and not isinstance(inst, type)
           ^^^^^^^^^^^^
  File "/Users/patrick/mambaforge/envs/dask-expr/lib/python3.11/site-packages/pandas/core/dtypes/generic.py", line 38, in _check
    return getattr(inst, attr, "_typ") in comp
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
RecursionError: maximum recursion depth exceeded in comparison
```

phofl:
- [x] closes #81

I'll open an issue about the comment in the test

### DIFF
--- a/dask_expr/expr.py
+++ b/dask_expr/expr.py
@@ -20,6 +20,7 @@ from dask.dataframe.core import (
     is_index_like,
     is_series_like,
 )
+from dask.dataframe.dispatch import meta_nonempty
 from dask.utils import M, apply, funcname, import_required, is_arraylike
 
 replacement_rules = []
@@ -815,8 +816,10 @@ class Projection(Elemwise):
         if is_dataframe_like(self.frame._meta):
             return super()._meta
         # if we are not a DataFrame and have a scalar, we reduce to a scalar
-        if pd.api.types.is_scalar(self.operands[1]):
-            return 0
+        if not isinstance(self.operands[1], list) and not hasattr(
+            self.operands[1], "dtype"
+        ):
+            return meta_nonempty(self.frame._meta).iloc[0]
         # Avoid column selection for Series/Index
         return self.frame._meta
 
--- a/dask_expr/expr.py
+++ b/dask_expr/expr.py
@@ -816,8 +816,8 @@ class Projection(Elemwise):
         if is_dataframe_like(self.frame._meta):
             return super()._meta
         # if we are not a DataFrame and have a scalar, we reduce to a scalar
-        if not isinstance(self.operands[1], list) and not hasattr(
-            self.operands[1], "dtype"
+        if not isinstance(self.operand("columns"), list) and not hasattr(
+            self.operand("columns"), "dtype"
         ):
             return meta_nonempty(self.frame._meta).iloc[0]
         # Avoid column selection for Series/Index
--- a/dask_expr/expr.py
+++ b/dask_expr/expr.py
@@ -288,6 +288,8 @@ class Expr:
     def __getitem__(self, other):
         if isinstance(other, Expr):
             return Filter(self, other)  # df[df.x > 1]
+        # elif pd.api.types.is_scalar(other):
+        #     return GetitemScalar(self, other)
         else:
             return Projection(self, other)  # df[["a", "b", "c"]]
 
@@ -814,6 +816,9 @@ class Projection(Elemwise):
     def _meta(self):
         if is_dataframe_like(self.frame._meta):
             return super()._meta
+        # if we are not a DataFrame and have a scalar, we reduce to a scalar
+        if pd.api.types.is_scalar(self.operands[1]):
+            return 0
         # Avoid column selection for Series/Index
         return self.frame._meta
 
--- a/dask_expr/expr.py
+++ b/dask_expr/expr.py
@@ -288,8 +288,6 @@ class Expr:
     def __getitem__(self, other):
         if isinstance(other, Expr):
             return Filter(self, other)  # df[df.x > 1]
-        # elif pd.api.types.is_scalar(other):
-        #     return GetitemScalar(self, other)
         else:
             return Projection(self, other)  # df[["a", "b", "c"]]
 
--- a/dask_expr/expr.py
+++ b/dask_expr/expr.py
@@ -850,7 +850,11 @@ class Index(Elemwise):
 
     @property
     def _meta(self):
-        return self.frame._meta.index
+        meta = self.frame._meta
+        # Handle scalar results
+        if is_series_like(meta) or is_dataframe_like(meta):
+            return self.frame._meta.index
+        return meta
 
     def _task(self, index: int):
         return (
--- a/dask_expr/reductions.py
+++ b/dask_expr/reductions.py
@@ -296,10 +296,6 @@ class Count(Reduction):
 class Min(Max):
     reduction_chunk = M.min
 
-    def _simplify_up(self, parent):
-        if isinstance(parent, Projection):
-            return self.frame[parent.operand("columns")].min(skipna=self.skipna)
-
 
 class Mode(ApplyConcatApply):
     """
--- a/dask_expr/reductions.py
+++ b/dask_expr/reductions.py
@@ -306,6 +306,10 @@ class Count(Reduction):
 class Min(Max):
     reduction_chunk = M.min
 
+    def _simplify_up(self, parent):
+        if isinstance(parent, Projection):
+            return self.frame[parent.operand("columns")].min(skipna=self.skipna)
+
 
 class Mode(ApplyConcatApply):
     """
--- a/dask_expr/tests/test_collection.py
+++ b/dask_expr/tests/test_collection.py
@@ -94,8 +94,6 @@ def test_dask(pdf, df):
 def test_reductions(func, pdf, df):
     assert_eq(func(df), func(pdf))
     assert_eq(func(df.x), func(pdf.x))
-    # re-conversion to Series is necessary since Dask does not reduce. Trying to
-    # figure out whether this was intended or not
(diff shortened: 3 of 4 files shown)


dask_expr/reductions.py
```diff
@@ -306,6 +306,10 @@ def reduction_aggregate(cls, df):
 class Min(Max):
     reduction_chunk = M.min
 
+    def _simplify_up(self, parent):
+        if isinstance(parent, Projection):
+            return self.frame[parent.operand("columns")].min(skipna=self.skipna)
```
rjzamora, line 311: This follows the convention used in other `Reduction` implementations, but as discussed in https://github.com/mrocklin/dask-expr/pull/88#discussion_r1196734567, these can all be written the same way:

```python
            return type(self)(self.frame[parent.operand("columns")], *operands[1:])
```

phofl: Yeah this needs a rebase when the other PR is merged.

dask_expr/tests/test_collection.py
```diff
@@ -94,6 +94,11 @@ def test_dask(pdf, df):
 def test_reductions(func, pdf, df):
     assert_eq(func(df), func(pdf))
     assert_eq(func(df.x), func(pdf.x))
+    # re-conversion to Series is necessary since Dask does not reduce. Trying to
+    # figure out whether this was intended or not
+    # check_dtype False because sub-selection of columns that is pushed through
+    # is not reflected in the meta calculation
+    assert_eq(func(df)["x"], pd.Series(func(pdf)["x"]), check_dtype=False)
```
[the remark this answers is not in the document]
phofl: Yeah the main problem we are facing here is:

```
df = pd.DataFrame({"a": [11, 12, 31], "b": 1.0})
ddf = from_pandas(df, npartitions=1)
ddf.a.max() # reduce
ddf.max()["a"] # does not
```

This is what happens in Dask. Since we are transforming the second one into the first one we lose these information. I think Dask should reduce in the second case as well. I'll open an issue in Dask. Wdyt?

dask_expr/reductions.py
phofl: Rebased

dask_expr/tests/test_collection.py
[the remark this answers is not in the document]
rjzamora: >I'll open an issue in Dask. Wdyt?

Yes, I think this makes sense as a dask issue.

For the case of dask-expr, `df.a.max()` does reduce, and `df.max()["a"]` should too (but doesn't). As far as I can tell, the problem with the second case is that we are calling `compute` on a dask-expr `Series`, but the underlying expression is being optimized into a `Scalar` collection within that `compute` call (and so the `__dask_postcompute__` logic is converting the scalar back to a `Series`)... Still thinking about the proper "fix" for this.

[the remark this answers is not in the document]
phofl: Yeah exactly. The cast-back happens in ``_concat``. One option would be to not cast back anymore if we have len(args) == 1, but this breaks other things. So will have to take a closer look how we can do this efficiently.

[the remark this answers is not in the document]
rjzamora: Looks like we may be able to change `FrameBase.__dask_postcompute__` to the following?:

```python
    def __dask_postcompute__(self):
        state = self.simplify()
        if type(self) != type(state):
            return state.__dask_postcompute__()
        return _concat, ()
```

Wdyt? Does this help with the test?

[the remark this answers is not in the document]
phofl: We don't need a new class, that was crab.

Adjusted the meta creation in Projection, which does the trick. Wdyt?

Are we ok with using a pandas utility to check for scalars?

dask_expr/tests/test_collection.py
```diff
@@ -94,6 +94,11 @@ def test_dask(pdf, df):
 def test_reductions(func, pdf, df):
     assert_eq(func(df), func(pdf))
     assert_eq(func(df.x), func(pdf.x))
+    # re-conversion to Series is necessary since Dask does not reduce. Trying to
+    # figure out whether this was intended or not
```
rjzamora, line 98: Not relevant anymore?

dask_expr/expr.py
```diff
@@ -814,6 +814,9 @@ def columns(self):
     def _meta(self):
         if is_dataframe_like(self.frame._meta):
             return super()._meta
+        # if we are not a DataFrame and have a scalar, we reduce to a scalar
+        if pd.api.types.is_scalar(self.operands[1]):
+            return 0
```
rjzamora, line 819: I'd like to avoid anything too pandas specific here (otherwise, I'll probably end up needing to change it to get the cudf backend working). Also, can't the scalar type be something other than an `int`?

Maybe if `self.operand("columns")` is not a `list`, we just return `meta_nonempty(self.frame._meta).iloc[0]`?

phofl: Yes definitely, forgot to commit the meta_nonemtpy part 

list is not sufficient, but we can create an explicit check

dask_expr/tests/test_collection.py
phofl: Removed, thx

PR #88, opened 2023-05-17
Move ``simplify_up`` implementation to Reductions class

phofl:
This is one approach to removing the duplicated code parts

### DIFF
--- a/dask_expr/reductions.py
+++ b/dask_expr/reductions.py
@@ -138,6 +138,7 @@ class Reduction(ApplyConcatApply):
     reduction_chunk = None
     reduction_combine = None
     reduction_aggregate = None
+    meth = None
 
     @classmethod
     def chunk(cls, df, **kwargs):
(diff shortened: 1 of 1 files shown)


dask_expr/reductions.py
```diff
@@ -264,6 +258,9 @@ def _simplify_down(self):
             child = max(self.frame.dependencies(), key=lambda expr: expr.npartitions)
             return Len(child)
 
+    def _simplify_up(self, parent):
+        return
```
rjzamora, line 262: Can't we push down a `Project` to come before `Len` as well?

rjzamora (commented):
Thanks for looking into this @phofl !  I'm hoping we can avoid adding the new `meth` attribute.

dask_expr/reductions.py
rjzamora: Cool - Makes sense to me.

rjzamora (approved):
Thanks @phofl !

PR #90, opened 2023-05-18
Add ``idxmin`` and ``idxmax``

phofl:


dask_expr/reductions.py
```diff
@@ -2,6 +2,9 @@
 import toolz
 from dask.dataframe.core import (
     _concat,
+    idxmaxmin_agg,
```
phofl, line 5: Are we ok with dispatching like this?

mrocklin: You mean using dask.dataframe code?  If so, yes.  In some future swap we'll bring all of these functions into this library.

mrocklin (approved):
Looks good to me.

dask_expr/reductions.py
mrocklin: I've been rewriting overly-complex things that have been about how we define algorithms, but leaving functions-used-as-tasks alone.  At this stage I care a lot about how we organize operations and very little about specific execution code.  

So for example I really enjoyed rewriing the ApplyConcatApply logic (this is used in reductions, groupby-aggregations, and more).

Not sure if that makes sense or not.

phofl: No that helps, thx for the explanation. Makes judging stuff easier

PR #93, opened 2023-05-18
Implement ``nbytes`` for ``Series``

phofl:


dask_expr/tests/test_collection.py
```diff
@@ -98,6 +98,12 @@ def test_reductions(func, pdf, df):
     assert_eq(func(df.x), func(pdf.x))
 
 
+def test_nbytes(pdf, df):
+    with pytest.raises(NotImplementedError, match="nbytes is not implemented"):
+        df.nbytes
+    assert_eq(df.x.nbytes.compute(), pdf.x.nbytes)
```
mrocklin, line 104: Do you need the `.compute` here or can we get this at expression generation time?

mrocklin (approved):

dask_expr/tests/test_collection.py
phofl: I don't think so, if there is something like a ``unique`` call before, we have to wait for the result to know the size of our object

mrocklin: Ah sorry, I misread this.  I thought that this was triggering within the pytest.raises line.  I see above now that you do have just `df.nbytes` under the `pytest.raises` context manager.  I'm good.

mrocklin, line 104: Well, actually, we should drop the `.compute()` I would expect `assert_eq` to handle this.

phofl: Yikes, yes of course. Thx

PR #95, opened 2023-05-18
Implement ``value counts``

phofl:


dask_expr/expr.py
```diff
@@ -382,6 +382,11 @@ def idxmax(self, skipna=True, numeric_only=None):
     def mode(self, dropna=True):
         return Mode(self, dropna=dropna)
 
+    def value_counts(self, sort=None, ascending=False, dropna=True, normalize=False):
+        if is_dataframe_like(self.frame):
+            raise NotImplementedError("value_counts not implemented for DataFrame")
```
mrocklin, line 387: Nothing to do in this PR, but this pattern has me a little concerned.  I like that we don't think about the Series/Dataframe separation in expr.py, but these dynamic checks feel like a bad pattern to start seeing.

mrocklin (approved):

dask_expr/expr.py
phofl: Yes I agree, but I am afraid that we will need a bunch of them if we don't solve this differently. Keywords that exist only for DataFrame methods come to mind for example. I guess this is something we have to solve at some point

PR #96, opened 2023-05-18
Implement ``nlargest`` and ``nsmallest``

phofl:
This is one of the keyword cases I mentioned on the other pr.

### DIFF
--- a/dask_expr/reductions.py
+++ b/dask_expr/reductions.py
@@ -267,6 +267,46 @@ class IdxMax(IdxMin):
     _fn = "idxmax"
 
 
+class NLargest(Reduction):
+    _defaults = {"n": 5, "_columns": None}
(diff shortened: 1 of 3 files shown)


dask_expr/reductions.py
```diff
@@ -268,6 +268,46 @@ class IdxMax(IdxMin):
...
+    _parameters = ["frame", "n", "_columns"]
+    reduction_chunk = M.nlargest
+    reduction_aggregate = M.nlargest
+
+    @classmethod
+    def chunk(cls, df, **kwargs):
+        return cls.reduction_chunk(df, **kwargs)
+
+    @classmethod
+    def combine(cls, inputs: list, **kwargs):
```
phofl, line 282: Will rewrite this logic into a more generic class after this is in

PR #97, opened 2023-05-18
Switch default of ``numeric_only`` for consistency

phofl:
If we only support pandas 2.0 and up then this is the best way to move forward

### DIFF
--- a/dask_expr/expr.py
+++ b/dask_expr/expr.py
@@ -355,16 +355,16 @@ class Expr:
     def __or__(self, other):
         return Or(other, self)
 
-    def sum(self, skipna=True, numeric_only=None, min_count=0):
+    def sum(self, skipna=True, numeric_only=False, min_count=0):
         return Sum(self, skipna, numeric_only, min_count)
 
-    def prod(self, skipna=True, numeric_only=None, min_count=0):
+    def prod(self, skipna=True, numeric_only=False, min_count=0):
         return Prod(self, skipna, numeric_only, min_count)
 
-    def mean(self, skipna=True, numeric_only=None, min_count=0):
+    def mean(self, skipna=True, numeric_only=False, min_count=0):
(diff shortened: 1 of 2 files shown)


rjzamora:
Just a note that cudf does not support Pandas-2.0 just yet.  So, I don't love this change (or that we only support Pandas-2).

mrocklin:
Ah, sorry to push quickly.  Thanks for speaking up @rjzamora . 

I'm curious, what are cudf's plans on Pandas 2.0?  What is the timeline there?

In general I'd like for dask-expr to be pretty free of legacy concerns if we can and "live in the future".  I also acknowledge the importance of cudf.  Do you have suggestions?

rjzamora:
>I'm curious, what are cudf's plans on Pandas 2.0? What is the timeline there?

My understanding is the Pandas-2.0 support was bumped from the June release, and August may not even be "safe".

With that said, there is a cudf feature branch where **most** "important" issues have been resolved ([`pandas_2.0_feature_branch`](https://github.com/rapidsai/cudf/tree/pandas_2.0_feature_branch)).  I don't think we will run into too many issues if we use that branch for development/testing.

Overall, I think it's fine to use/assume pandas-2.0, but it would be nice to be general in cases where it doesn't require the code to be too ugly.

mrocklin:
> Overall, I think it's fine to use/assume pandas-2.0, but it would be nice to be general in cases where it doesn't require the code to be too ugly

I think that we're pretty aligned here.  We should aim towards 2.0.  If it's trivial to support cudf we should do so.  In general given that this won't hit users for a few months we don't need to worry too much about things as long as we expect cudf to align with 2.0 at roughly the same timeline, and as long as we can reasonably test (which it sounds like maybe we can).

phofl:
Sounds good, is there an issue where you are tracking incompatibilities with 2.0 that aren't resolved yet?

PR #99, opened 2023-05-19
Refactor reducers that keep the dimension

phofl:
This makes the abstraction a bit nicer, since these reducers behave different than all the others

### DIFF
--- a/dask_expr/reductions.py
+++ b/dask_expr/reductions.py
@@ -369,6 +369,12 @@ class Mode(ApplyConcatApply):
 
 
 class ReductionConstantDim(Reduction):
+    """
+    Some reductions reduce the number of rows in your object but keep the original
+    dimension, e.g. a DataFrame stays a DataFrame instead of getting reduced to
+    a Series.
+    """
+
     @classmethod
     def chunk(cls, df, **kwargs):
         return cls.reduction_chunk(df, **kwargs)
--- a/dask_expr/reductions.py
+++ b/dask_expr/reductions.py
@@ -268,46 +268,6 @@ class IdxMax(IdxMin):
     _fn = "idxmax"
 
 
-class NLargest(Reduction):
-    _defaults = {"n": 5, "_columns": None}
-    _parameters = ["frame", "n", "_columns"]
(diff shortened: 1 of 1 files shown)


mrocklin:
No immediate thoughts.  @rjzamora any thoughts?

dask_expr/reductions.py
```diff
@@ -408,7 +368,49 @@ def aggregate_kwargs(self):
...
+class ReductionConstantDim(Reduction):
+    @classmethod
+    def chunk(cls, df, **kwargs):
+        return cls.reduction_chunk(df, **kwargs)
+
+    @classmethod
+    def combine(cls, inputs: list, **kwargs):
+        func = cls.reduction_combine or cls.reduction_aggregate or cls.reduction_chunk
+        df = _concat(inputs)
+        return func(df, **kwargs)
```
rjzamora, line 380: If I understand correctly, the only difference is that we don't want to convert a series back to dataframe.
Can we just add an `enforce_dataframe = True` class property to `Reduction`, and just set it to `False` for `ValueCounts`, etc?

phofl: Personally, I find this more readable and clear than a flag that's buried in the functions, but no strong opinion. Happy to change if you prefer

A flag should do the trick, yes

rjzamora: Ah, that's fair. In hindsight, I think I just find the `ReductionConstantDim` name a bit confusing. Could this be `SeriesReduction`? Or, am I misunderstanding the meaning of `ReductionConstantDim`?

phofl: Maybe the name is not ideal, this means that you reduce the number of rows but the dimension stays the same. Nlargest for example reduces the number of rows in your DataFrame but the result will still be a DataFrame, while sum reduces to a Series

rjzamora: Okay, I see. Maybe just add a description in the doc-string and I'll be fine with the name then.

phofl: Added a docstring

rjzamora (approved):

PR #100, opened 2023-05-19
Implement ``memory_usage``

phofl:
I think we need a general mechanism to pop not-needed keywords from the kwargs properties, otherwise we will run into similar problems in a bunch of methods.

### DIFF
--- a/dask_expr/expr.py
+++ b/dask_expr/expr.py
@@ -398,17 +398,6 @@ class Expr:
             raise NotImplementedError("value_counts not implemented for DataFrame")
         return ValueCounts(self, sort, ascending, dropna, normalize)
 
-    def memory_usage(self, deep=False, index=no_default):
-        if is_index_like(self._meta):
-            if index is not no_default:
-                raise TypeError(
-                    "index key-word is not supported when calculating memory_usage of index"
-                )
-            return MemoryUsage(self, deep=deep)
-        if index is no_default:
-            index = True
-        return MemoryUsage(self, deep=deep, _index=index)
-
     def min(self, skipna=True, numeric_only=False, min_count=0):
         return Min(self, skipna, numeric_only, min_count)
 
@@ -1397,7 +1386,6 @@ from dask_expr.reductions import (
     IdxMin,
     Max,
     Mean,
-    MemoryUsage,
     Min,
     Mode,
     NBytes,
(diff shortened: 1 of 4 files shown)


dask_expr/reductions.py
```diff
@@ -441,3 +442,26 @@ def aggregate_kwargs(self):
...
         return
+
+
+class MemoryUsage(Reduction):
+    _parameters = ["frame", "deep", "_index"]
+    _defaults = {"deep": False, "_index": True}
+    reduction_chunk = M.memory_usage
+    reduction_combine = staticmethod(
+        lambda x, is_dataframe: x.groupby(x.index).sum() if is_dataframe else x.sum()
+    )
```
mrocklin, line 453: This seems complicated enough that maybe it deserves a normal `def ...` method

```python
@staticmethod
def reduction_combine(x, is_dataframe):
    ...
```

dask_expr/expr.py
```diff
@@ -398,6 +398,17 @@ def value_counts(self, sort=None, ascending=False, dropna=True, normalize=False)
...
+    def memory_usage(self, deep=False, index=no_default):
+        if is_index_like(self._meta):
+            if index is not no_default:
+                raise TypeError(
+                    "index key-word is not supported when calculating memory_usage of index"
+                )
+            return MemoryUsage(self, deep=deep)
+        if index is no_default:
+            index = True
+        return MemoryUsage(self, deep=deep, _index=index)
```
mrocklin, line 410: An alternative is to stop putting these methods on `Expr` and instead put them on `DataFrame/Series/Index`.  This makes the `Expr` class a little less useful on its own, but maybe that's ok because it's rarely used on its own.  

I propose that we use this reduction as an example, and see how this would look if we put the methods on the DataFrame/Series/Index classes directly and see if that feels cleaner.

Thoughts?

mrocklin (commented):
Thanks for putting this up and calling out the unpleasantness around types.  A suggestion below

dask_expr/reductions.py
phofl: Sounds good

dask_expr/expr.py
phofl: Added it to Series/Index/Frame, this is probably the cleanest we can come up with with regards to if/else stuff. Thoughts?

PR #101, opened 2023-05-19
Implement ``drop_duplicates`` and ``unique``

phofl:


dask_expr/reductions.py
```diff
@@ -118,6 +118,67 @@ def _divisions(self):
...
+    def _subset_kwargs(self):
+        if is_series_like(self.frame._meta):
+            return {}
+        return {"subset": self.subset}
+
+    @property
+    def chunk_kwargs(self):
+        return {"ignore_index": self.ignore_index, **self._subset_kwargs()}
+
+    def _simplify_up(self, parent):
```
phofl, line 178: We can simplify here in some cases, but would prefer to do as follow up

dask_expr/tests/test_collection.py
```diff
@@ -456,3 +457,29 @@ def test_repartition_divisions(df, opt):
...
             assert part.max() < df2.divisions[p + 1]
+
+
+def test_drop_duplicates(df, pdf):
+    assert_eq(df.drop_duplicates(), pdf.drop_duplicates())
+    assert_eq(
+        df.drop_duplicates(ignore_index=True), pdf.drop_duplicates(ignore_index=True)
+    )
+    assert_eq(df.drop_duplicates(subset=["x"]), pdf.drop_duplicates(subset=["x"]))
+    assert_eq(df.x.drop_duplicates(), pdf.x.drop_duplicates())
```
mrocklin, line 468: For tests like this I wonder if we should be relying on one giant parametrized test.

A lot of our tests are just `assert_eq(func(df), fund(pdf))`.  Maybe there is something we should be doing to make this easier / lighter weight?

dask_expr/expr.py
```diff
@@ -404,6 +404,18 @@ def min(self, skipna=True, numeric_only=False, min_count=0):
...
+        # let pandas raise errors, e.g. not defined on DataFrame
+        self._meta.unique()
+        return Unique(self)
+
+    def drop_duplicates(self, subset=None, ignore_index=False):
+        if subset is not None:
+            # Fail early if subset is not valid
+            meta_nonempty(self._meta).drop_duplicates(subset=subset)
+
+        return DropDuplicates(self, subset=subset, ignore_index=ignore_index)
```
mrocklin, line 417: If these are only on `Series/DataFrame` then maybe the answer is that we don't put them on `Expr`, but instead only put them on `Series/DataFrame`.

See https://github.com/mrocklin/dask-expr/issues/98#issuecomment-1556191825 for this same suggestion on a larger issue.

mrocklin: Ah, I forgot that we already had this conversation on another PR as well.  Thoughts here?

dask_expr/tests/test_collection.py
mrocklin: I'm also happy keeping these things separate, especially when there are lots of options like this to test.  Just a thought.

phofl: Generally, I prefer tests where the parametrisation isn't too complicated. Historically, we had some tests in pandas where the whole logic was encoded into the parametrisation, which made debugging a real PITA when something broke

dask_expr/expr.py
phofl: Yep makes sense, will update

phofl: Updated

PR #102, opened 2023-05-19
Simple statistics

rjzamora:
Supersedes #84

- Implements `__len__`
- Adds a `_lengths` property and `_partitioning` method to `Expr`. These are essentially mechanisms to track "known" partition lengths and partitioning information.
- Adds new logic to `ReadParquet` to use collect and use parquet-metadata statistics to implement `ReadParquet._lengths` and `ReadParquet._partitioning`


dask_expr/io/parquet.py
```diff
@@ -56,7 +53,7 @@ class ReadParquet(PartitionsFiltered, BlockwiseIO):
         "categories": None,
         "index": None,
         "storage_options": None,
-        "calculate_divisions": False,
+        "calculate_divisions": True,
```
mrocklin, line 56: Is this asking for statistics by default?  If so, this has caused problems in the past.  Is it wise?  (genuine question, you know more than anyone here)

dask_expr/io/parquet.py
```diff
@@ -246,9 +241,106 @@ def _filtered_task(self, index: int):
             return (operator.getitem, tsk, self.columns[0])
         return tsk
 
+    @cached_property
+    def _lengths(self):
+        """Return known partition lengths using parquet statistics"""
+        if not self.filters:
+            _update_length_statistics(self)
```
mrocklin, line 248: Should this function be a method?

dask_expr/expr.py
```diff
@@ -616,6 +618,61 @@ def visualize(self, filename="dask-expr.svg", format=None, **kwargs):
...
+          - Reverse-sorted data: `("decreasing", <divisions>)`
+          - Shuffled data: `("hash", <npartitions>)`
+
+        Note that un-named index columns must be specified as
+        `"__index__"` (`None` is not supported).
+
+        Return
+        ------
+        partitioning: dict
+        """
```
mrocklin, line 646: I'd welcome a conversation about this.  

My initial thought is that it made sense to store some baseline information like ...

1.  row counts of each partition
2. min/max values of each column in each partition

These are similar to what comes out of parquet.  Then, when we wanted to ask something, we would consult that raw data.  

This feels like we're now storing derivative values off of that data.  This makes me slightly nervous because it opens the door to tracking lots of state.  I would be more comfortable if we were to track the underlying state (counts, mins, maxes) and then decided to compute quantities like these on the fly.  That feels more tightly scoped to me.  

Thoughts?

dask_expr/expr.py
```diff
@@ -616,6 +618,61 @@ def visualize(self, filename="dask-expr.svg", format=None, **kwargs):
...
+    def _divisions(self):
+        return (None, None)
+
+    @property
+    def _meta(self):
+        return make_meta(self.value)
+
+    def _task(self, index: int):
+        assert index == 0
+        return self.value
```
mrocklin, line 674: 🤔 this is interesting.

dask_expr/io/parquet.py
rjzamora: Oops! This is actually a mistake. It is less common for metadata gathering to be painful now, but it is still nice to avoid it by default (when we know the files are sufficiently small, and are not calculating divisions or filtering).

dask_expr/expr.py
rjzamora: This PR does a few different things, and some of those "things" I am much more confident in than others.

1. We add an optional `_lengths` attribute the `Expr` to store "known" partition lengths.
2. We add an optional `_partitioning` method to `Expr` so that an expression can check if the underlying collection is partitioned by a specific set of columns (even if that column does not include an index with known divisions).
3. We add logic to some `Expr` classes (mostly `ReadParquet`) to "lazily" collect the necessary statistics when `_lengths` or `_partitioning` information is requested.

The primary reason this PR is still marked as "draft" is that the current iteration will **always** attempt to go back and collect statistics in `ReadParquet` when `_lengths` or `_partitioning` are called (and the necessary statistics are missing). While it will always make sense to collect partition-length statistics in support of something like `len(df)`, it may not always be the best idea to collect statistics.  In fact, I'm already a bit uncomfortable with the fact that column-projection and predicate-pushdown optimizations currently require us to repeat the initial dataset processing, which can be slow on some systems (this is something I'd like to address separately).

Note that I also think the specific API can be improved, but the "eagerness" of the lazy-metadata collection feels like the most challenging short-term blocker.

What you seem to be uncomfortable with is the fact that we are not adding something like `Expr._mins` and `Expr._maxes`, but are instead exposing a method to provide more general (derivative) information about how the collection is partitioned.  I'm very open to other approaches. My current proposal here was just the natural result of attempting to store mins/maxes, and finding that my personal attempt at doing so was not particularly clean or useful.  In most cases, the original `ReadParquet` expression will not collect useful min/max statistics.  When the expression **does** collect min/max statistics, the only reason we care about them is to tell us how/if the collection is partitioned.  For this reason, I found it most natural to allow specific classes (like `ReadParquet` and `Shuffle`) to worry about what kinds of statistics they want to collect/track (if any).

I'll think a bit more about this.

mrocklin: > I'm already a bit uncomfortable with the fact that column-projection and predicate-pushdown optimizations currently require us to repeat the initial dataset processing, which can be slow on some systems (this is something I'd like to address separately).

I noticed this recently.  I wonder if the parquet code could benefit from a module-level lru-cache

mrocklin: > What you seem to be uncomfortable with is the fact that we are not adding something like Expr._mins and Expr._maxes, but are instead exposing a method to provide more general (derivative) information about how the collection is partitioned

Yeah, I'm comparing this to database world where you have a reference table which is the single point of truth (SPOT) and then views on that table.  This feels like we're storing the views as concrete tables.  Bad things tend to result from that behavior.

As an example.  I could imagine future applications aside from sortedness.  We've mentioned a couple of these including filtering / partition pruning and optimizations that are based on the values.  I think that storing the underlying data is more future-proof.  

I probably wouldn't have separate protocols for `_maxes` and `_mins` but maybe a protocol that includes `_min_maxes` or all column-based statistics (if they're likely to be consistent across all systems that provide this information (it might make sense to look at what Snowflake, Parquet, and Delta all provide, for example)).

dask_expr/expr.py
rjzamora: I struggled a bit to come up with something better than `Literal`. The overall purpose is to allow us to replace an `Expr` with a literal ("known") value in `_simplify_down`/`up`.  However, the `simplify` logic would need to be uglier if we wanted to to allow a non-`Expr` object to be returned.

rjzamora:
@mrocklin - Let me know if you are concerned by any of the changes remaining in this PR.  I agree with your off-line suggestion that we could also introduce another method like `Expr._maxima` to track min/max statistics. However, I'd like to tackle that in a follow-up.

mrocklin (commented):
OK, I'm sorry to have this drag on for so long, but please bear with me here.  

What if we don't create a new protocol on expressions here, but instead make a new expression, `Lengths`

```python
class Lengths(Expr):
    _parameters = ["frame"]

    @property
    def _meta(self):
        return []

    def _simplify_down(self):
        if isinstance(self.frame, Elemwise):
            return Lengths(self.frame.operands[0])

        if isinstance(self.frame, ReadParquet):
            return Literal(self.frame._get_lengths())

    def _layer(self):
        return {self._name: [(self.frame._name, i): (len, (self.frame._name, i) for i in range(self.frame.npartitions]}
```

Does this give you what you want?  If so, I like it because we use an existing extension mechanism (Operations/Expr subclasses) rather than make something new.  I'm also ok adding new methods, but this seems like it might work and invent one fewer thing.

rjzamora:
>Does this give you what you want?

I'm not completely sure. I was pushing on a similar design earlier on, but moved away from it when I started considering how I would need min/max statistics to be accessed/used.

To understand the limitations of an `Expr`-based approach, consider the case that the user calls `set_index`, and you want to be able to check if the collection is already sorted by the selected column. If we rely on something like `MinMax(Expr)`, (I think) we would also need some other mechanism to tell us if **computing** this expression would require "real" data to be read into memory.  That is, I would only want to use `MinMax` to optimize a `set_index` operation if I knew it would **never** require any real IO/computation.

Perhaps one reasonable approach would be to call these expressions something like `MinMaxStats` and `LengthStats`, and explicitly prohibit them from reading in "real" data. This way, optimization logic can use/compute a `MinMaxStats` expression to query for known statistics without risking any real/expensive IO. We could also introduce the `Lengths` expression you have suggested, but allow `ReadParquet` to replace  `Lengths` with `LengthStats` in `_simplify_up`.

Overall, I think I'm saying that I like the idea of leveraging the `Expr` foundation to query/leverage statistics. However, I think it is important that we clearly distinguish "pure" statistics-scanning expressions in some way. What do you think?

mrocklin:
Yeah, so I think that this gets at my other question of "exactly what are the semantics around the `force=` keyword?"  I agree generally with your framing. 

Using the Expr approach in an operation like `set_index` we would probably do something like the following:

```python
def set_index(self, column):
    minmaxes = self.minmax(column).optimize()
    if isinstance(minmaxes, Literal):  # 🎉
        ...  do something easy
    if hasattr(minmaxes.frame, "minmaxes"):
        ... consider calling this method
    ...
```

>  If we rely on something like MinMax(Expr), (I think) we would also need some other mechanism to tell us if computing this expression would require "real" data to be read into memory. That is, I would only want to use MinMax to optimize a set_index operation if I knew it would never require any real IO/computation.

Agreed.  I think that that information is present in the optimized expression tree.

dask_expr/expr.py
```diff
@@ -1031,6 +1049,32 @@ def _task(self, index: int):
...
+    @property
+    def _meta(self):
+        return tuple()
+
+    def _divisions(self):
+        return (None, None)
+
+    def _simplify_down(self):
+        if isinstance(self.frame, Elemwise):
+            return Lengths(self.frame.operands[0])
```
mrocklin, line 1066: I was simplifying things before.  You'll have to watch out for situations like `x.sum() + x`.  I think that if you look at the implementation for `Len` we handle this well.

dask_expr/io/parquet.py
```diff
@@ -104,6 +116,11 @@ def _simplify_up(self, parent):
                 kwargs["filters"] = filters.combine(kwargs["filters"]).to_list_tuple()
                 return ReadParquet(**kwargs)
 
+        if isinstance(parent, Lengths):
+            _lengths = self._get_lengths()
+            if _lengths:
+                return Literal(_lengths)
```
mrocklin, line 122: maybe this?

```suggestion
        super()._simplify_up(parent)
```

(or however super works)

dask_expr/reductions.py
```diff
@@ -346,7 +346,10 @@ class Len(Reduction):
     reduction_aggregate = sum
 
     def _simplify_down(self):
-        if isinstance(self.frame, Elemwise):
+        _lengths = Lengths(self.frame).optimize()
+        if isinstance(_lengths, Literal):
+            return Literal(sum(_lengths.value))
+        elif isinstance(self.frame, Elemwise):
             child = max(self.frame.dependencies(), key=lambda expr: expr.npartitions)
             return Len(child)
```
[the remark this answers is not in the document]
mrocklin: It may be that there are cases where we can move Len but not Lengths.  For example:

```python
df = dx.read_csv(...)
df = df.set_index("...")
len(df)
```

I'm not going to be able to pass `Lengths` through the `set_index` call, but I can pass `Len` through (the number of rows is the same after the full shufle).  Because of this I think that they probably need to remain separate operations with separate optimization paths.  I don't think that we can define one in terms of the other.

Also, I suspect that `Len` is likely to be more common than `Lengths`, so I'm disinclined to have it inherit any weaknesses from the other.

dask_expr/io/tests/test_io.py
```diff
@@ -207,6 +207,16 @@ def test_parquet_complex_filters(tmpdir):
...
 
 
+def test_parquet_len(tmpdir):
+    df = read_parquet(_make_file(tmpdir))
+    pdf = df.compute()
+
+    assert len(df[df.a > 5]) == len(pdf[pdf.a > 5])
+
+    s = (df["b"] + 1).astype("Int32")
+    assert len(s) == len(pdf)
```
mrocklin, line 217: Do you want to verify that we can do this without any actual execution somehow?

Maybe we can call something like the following

```python
expr = Len(s)
assert isinstance(s.optimize(), Literal)
```

dask_expr/io/parquet.py
```diff
@@ -365,3 +472,123 @@ def extract_pq_filters(cls, pq_expr: ReadParquet, predicate_expr: Expr) -> _DNF:
...
 
         return _DNF(_filters)
+
+
+#
+# Parquet-statistics handling
+#
+
+
+def _collect_pq_statistics(
```
mrocklin, line 482: Heads-up, I'm not reviewing the parquet specific bits here.

dask_expr/collection.py
```diff
@@ -85,6 +85,13 @@ def _meta(self):
     def size(self):
         return new_collection(self.expr.size)
 
+    def __len__(self):
+        return self._len
+
+    @functools.cached_property
+    def _len(self):
+        return sum(new_collection(expr.Lengths(self.expr)).compute())
```
mrocklin, line 93: I think that we'll still need to use `Len` here.  

I'm not sure how I feel about caching.  Ok to leave this as-is for now, and this probably makes sense to do, but it does open some questions about caching more generally.

dask_expr/io/parquet.py
rjzamora: Not sure I understand this suggestion? None of `ReadParquet`'s super-classes cover the the `_get_lengths` optimization.

mrocklin: I thought you had it on the `IO` class, but maybe this was github's diff misleading me and actually it was on `FromPandas`

rjzamora: Ah right, that makes sense. Yeah, I don't want to assume that all IO classes can leverage statistics.

dask_expr/io/tests/test_io.py
rjzamora: good call

dask_expr/collection.py
rjzamora: Went back to using `Len` and removed the caching.

dask_expr/reductions.py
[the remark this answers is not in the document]
mrocklin: I think it would be better to keep the `Len(Elemwise)` branch of this on top.  Otherwise we're constructing a `Lengths` object and optimizing it every pass through optimizations (and there are likely to be several of these).  

In principle, I wouldn't mind removing `Lengths` from this entirely, and instead have a `ReadParquet._simplify_up(Len)` case which is similar to but simpler than the `Lengths` case.

dask_expr/io/parquet.py
mrocklin, line 122: Maybe also add

```python
if isinstance(parent, Len):
    _lengths = self._get_lengths()
    if _lengths:
        return Literal(sum(_lengths))
```

mrocklin (approved):
Minor comment.  Generally I'm good to merge.  

Does this satisfy your needs in projects like Merlin?

mrocklin (approved):

PR #104, opened 2023-05-20
Implement ``to_timestamp``

phofl:


dask_expr/expr.py
```diff
@@ -768,6 +771,17 @@ class Elemwise(Blockwise):
...
 
+class ToTimestamp(Elemwise):
+    _parameters = ["frame", "freq", "how"]
+    _defaults = {"freq": None, "how": "start"}
+    operation = M.to_timestamp
+
+    def _divisions(self):
+        return tuple(
+            pd.Index(self.frame.divisions).to_timestamp(freq=self.freq, how=self.how)
+        )
```
mrocklin, line 782: I was surprised by us changing the divisions too, but I see that this is what dask dataframe does (and so presumably pandas)

mrocklin (approved):

dask_expr/expr.py
phofl: The method name isn't ideal, since ``to_timestamp`` operates exclusively on the Index, so have to adapt the divisions as well.

PR #108, opened 2023-05-21
Implement ``dropna`` and fix ``Blockwise`` apply bug

phofl:
A bunch of methods in pandas are kwargs only since 2.0 came out, Blockwise wasn't able to handle this right now. We can not convert everything to a kwargs dict, because some methods are positional only...

Happy to hear suggestions whether there is a better way of achieving this, my helper creates args and kwargs based on a variable. 

Blockwise had a small bug in the apply branch, which is fixed here

### DIFF
--- a/dask_expr/expr.py
+++ b/dask_expr/expr.py
@@ -624,14 +624,27 @@ class Blockwise(Expr):
 
     @functools.cached_property
     def _meta(self):
-        args, kwargs = self._get_args_kwargs_from_operands(
-            lambda x: x._meta if isinstance(x, Expr) else x
-        )
+        func = lambda x: x._meta if isinstance(x, Expr) else x
+        args = [func(op) for op in self._args]
+        kwargs = {key: func(op) for key, op in self._kwargs}
         return self.operation(*args, **kwargs)
 
-    @property
-    def _kwargs(self):
+    @functools.cached_property
+    def _kwargs(self) -> dict:
+        if self._keyword_only:
+            return {
+                p: self.operand(p) for p in self._parameters if p in self._keyword_only
+            }
         return {}
+
+    @functools.cached_property
+    def _args(self) -> list:
+        if self._keyword_only:
+            args = [
+                self.operand(p) for p in self._parameters if p not in self._keyword_only
+            ] + self.operands[len(self._parameters) :]
+            return args
+        return self.operands
 
     def _broadcast_dep(self, dep: Expr):
         # Checks if a dependency should be broadcasted to
@@ -681,10 +694,9 @@ class Blockwise(Expr):
         -------
         task: tuple
         """
-        args, kwargs = self._get_args_kwargs_from_operands(
-            lambda x: self._blockwise_arg(x, index)
-        )
-        if self._kwargs or kwargs:
+        args = [self._blockwise_arg(op, index) for op in self._args]
+        kwargs = {key: self._blockwise_arg(op, index) for key, op in self._kwargs}
+        if kwargs:
             return (
                 apply,
                 self.operation,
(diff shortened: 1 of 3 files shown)


mrocklin:
@rjzamora can I ask you to review this one?  It looks like it requires slightly-more-than-trivial thinking

dask_expr/expr.py
```diff
@@ -679,12 +681,29 @@ def _task(self, index: int):
...
+            return (
+                apply,
+                self.operation,
+                (tuple, list(args)),
+                {**self._kwargs, **kwargs},
+            )
         else:
             return (self.operation,) + args
 
+    def _get_args_kwargs_from_operands(self, op_func) -> tuple[tuple, dict]:
```
rjzamora, line 697: I don't think there is anything wrong with the proposal here, but I do worry about "readability".  It *may* be a little easier to interpret if we just define `_args` and `_kwargs` properties, and use them throughout `Blockwise` (in lieu of `self.operands`. What do you think of something like this?:


```python
class Blockwise(Expr):
    ...
    _keyword_only = None

    @functools.cached_property
    def _meta(self):
        return self.operation(
            *[arg._meta if isinstance(arg, Expr) else arg for arg in self._args],
            **self._kwargs,
        )

    @functools.cached_property
    def _kwargs(self) -> dict:
        if self._keyword_only:
            return {
                p: self.operand(p)
                for p in self._parameters
                if p in self._keyword_only
            }
        return {}

    @functools.cached_property
    def _args(self) -> list:
        if self._keyword_only:
            args = [
                self.operand(p)
                for p in self._parameters
                if p not in self._keyword_only
            ] + self.operands[len(self._parameters):]
            return args
        return self.operands

    ....

    def _task(self, index: int):
        args = [self._blockwise_arg(op, index) for op in self._args]
        if self._kwargs:
            return (apply, self.operation, args, self._kwargs)
        else:
            return (self.operation, *args)
```

phofl: We can do this as well, but we have to apply ``_blockwise_arg`` to the kwargs values as well, which loses some of the advantages

rjzamora: Hmm. It makes me a bit nervous that  you are saying Kwargs may hold other expressions now.

rjzamora: Can you share an example where this is the case (sorry if it’s right in front of me)?

phofl: pandas deprecated a bunch of positional arguments in 2.0, ``clip`` is one example where this could be problematic (there are probably more)

rjzamora: Hmm. I’m not sure I understand the clip example. Is this because upper or lower can be another expression?

phofl: Yes exactly, that's a possibility.

rjzamora: Ah, okay. I understand where you are coming from then.

Keep in mind that you cannot include a key in the kwargs argument to `apply`. You can only include a keys within the `args`. For example, this assertion will fail:

```python
import dask
from dask.utils import apply

def func(x, y=None):
    return y


dsk = {
    "key": (apply, func, ["key-a"], {"y": ("key-b", 0)}),
    ("key-a", 0): "expr-a",
    ("key-b", 0): "expr-b",
}

assert dask.get(dsk, "key") == "expr-b"
```
Therefore, I don't think there is any good reason to apply `_blockwise_arg` to the kwargs. If we want to allow users to include another collection/expression in the kwargs, we will need to introduce a new version of `apply` (which I think we can probably punt on here).

phofl: I wasn't aware of this. In that case we can avoid the function for kwargs. Agreed, let's punt on this here

dask_expr/expr.py
```diff
@@ -679,11 +694,16 @@ def _task(self, index: int):
...
         task: tuple
         """
-        args = tuple(self._blockwise_arg(op, index) for op in self.operands)
+        args = [self._blockwise_arg(op, index) for op in self._args]
         if self._kwargs:
-            return (apply, self.operation, args, self._kwargs)
+            return (
+                apply,
+                self.operation,
+                (tuple, list(args)),
```
rjzamora, line 702:
```suggestion
                args,
```
Pretty sure we can just pass the list here?

phofl: No, this was a bug previously. The dask_dependency solving does not work when you pass a list here. I took this workaround from dask/dask

phofl: Ah, thx. That makes sense. Changed

rjzamora (approved):
Thanks!

PR #110, opened 2023-05-21
Add ``abs``

phofl:


dask_expr/expr.py
```diff
@@ -787,6 +790,11 @@ class Round(Elemwise):
     operation = M.round
 
 
+class Abs(Blockwise):
```
mrocklin, line 793: Elemwise here is better I think. That was operations like Len will be able to pass through.

phofl: Oh yes that makes sense. I was a bit confused why this wasn't elemwise in Dask as well, but thought maybe I am missing something

PR #111, opened 2023-05-22
Implement _simplify_up for drop_duplicates

### LINKED ISSUE #109: Pushing a ``Projection`` through an operation that operates on a subset of columns
A relatively common operation is something like the following:

```
df.dropna(subset=["a"]).b
```

We can optimise this to something like

```
df[["a", "b"]].dropna(subset=["a"]).b
```

Doing this with our Projection class would cause a RecursionError though, since we would try to push b up over and over again. Thoughts on how we can solve this?

- Implement another class that isn't used in the simplify_up steps
- Add an attribute that controls whether or not we can simplify
- ...

phofl:
- [x] closes #109 

This is what we discussed in the issue

### DIFF
--- a/dask_expr/reductions.py
+++ b/dask_expr/reductions.py
@@ -178,6 +178,8 @@ class DropDuplicates(Unique):
     def _simplify_up(self, parent):
         if self.subset is not None and isinstance(parent, Projection):
             columns = parent.columns
+            if not isinstance(columns, list) and not hasattr(columns, "dtype"):
+                columns = [columns]
             columns = set(columns).union(self.subset)
             if columns == set(self.frame.columns):
                 # Don't add unnecessary Projections, protects against loops
--- a/dask_expr/reductions.py
+++ b/dask_expr/reductions.py
@@ -178,10 +178,7 @@ class DropDuplicates(Unique):
     def _simplify_up(self, parent):
         if self.subset is not None and isinstance(parent, Projection):
             columns = parent.columns
-            subset = self.subset
-            if not isinstance(subset, list) and not hasattr(subset, "dtype"):
-                subset = [subset]
-            columns = set(columns).union(subset)
+            columns = set(columns).union(self.subset)
             if columns == set(self.frame.columns):
                 # Don't add unnecessary Projections, protects against loops
                 return
--- a/dask_expr/reductions.py
+++ b/dask_expr/reductions.py
@@ -176,7 +176,20 @@ class DropDuplicates(Unique):
         return {"ignore_index": self.ignore_index, **self._subset_kwargs()}
 
     def _simplify_up(self, parent):
-        return
+        if self.subset is not None and isinstance(parent, Projection):
+            columns = parent.operand("columns")
+            if isinstance(columns, str):
+                columns = [columns]
+            subset = self.subset if isinstance(self.subset, list) else [self.subset]
+            columns = set(columns).union(subset)
+            if columns == set(self.frame.columns):
+                # Don't add unnecessary Projections, protects against loops
+                return
+
+            return type(parent)(
+                type(self)(self.frame[sorted(columns)], *self.operands[1:]),
+                parent.operand("columns"),
+            )
 
 
 class Reduction(ApplyConcatApply):
--- a/dask_expr/reductions.py
+++ b/dask_expr/reductions.py
@@ -178,9 +178,11 @@ class DropDuplicates(Unique):
     def _simplify_up(self, parent):
         if self.subset is not None and isinstance(parent, Projection):
             columns = parent.operand("columns")
-            if isinstance(columns, str):
+            if not isinstance(columns, list) and not hasattr(columns, "dtype"):
                 columns = [columns]
-            subset = self.subset if isinstance(self.subset, list) else [self.subset]
+            subset = self.subset
+            if not isinstance(subset, list) and not hasattr(subset, "dtype"):
+                subset = [subset]
             columns = set(columns).union(subset)
             if columns == set(self.frame.columns):
                 # Don't add unnecessary Projections, protects against loops
--- a/dask_expr/reductions.py
+++ b/dask_expr/reductions.py
@@ -178,7 +178,7 @@ class DropDuplicates(Unique):
     def _simplify_up(self, parent):
         if self.subset is not None:
             columns = parent.columns
-            if not isinstance(columns, list) and not hasattr(columns, "dtype"):
+            if not isinstance(columns, pd.Index):
                 columns = [columns]
             columns = set(columns).union(self.subset)
             if columns == set(self.frame.columns):
@@ -187,7 +187,7 @@ class DropDuplicates(Unique):
 
             return type(parent)(
                 type(self)(self.frame[sorted(columns)], *self.operands[1:]),
-                parent.columns,
+                *parent.operands[1:],
             )
 
 
--- a/dask_expr/reductions.py
+++ b/dask_expr/reductions.py
@@ -176,7 +176,7 @@ class DropDuplicates(Unique):
         return {"ignore_index": self.ignore_index, **self._subset_kwargs()}
 
     def _simplify_up(self, parent):
-        if self.subset is not None and isinstance(parent, Projection):
+        if self.subset is not None:
             columns = parent.columns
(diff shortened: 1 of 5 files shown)


dask_expr/reductions.py
```diff
@@ -176,7 +176,22 @@ def chunk_kwargs(self):
         return {"ignore_index": self.ignore_index, **self._subset_kwargs()}
 
     def _simplify_up(self, parent):
-        return
+        if self.subset is not None and isinstance(parent, Projection):
+            columns = parent.operand("columns")
+            if not isinstance(columns, list) and not hasattr(columns, "dtype"):
+                columns = [columns]
```
mrocklin, line 182: I'll bet that this is just 

```suggestion
            columns = parent.columns
```

dask_expr/reductions.py
```diff
@@ -176,7 +176,22 @@ def chunk_kwargs(self):
...
+            subset = self.subset
+            if not isinstance(subset, list) and not hasattr(subset, "dtype"):
+                subset = [subset]
```
mrocklin, line 185: Should this be normalized somewhere else?  Maybe in the collection `drop_duplicates` code?

mrocklin (approved):

dask_expr/reductions.py
phofl: Not sure, pandas can deal with both. Maybe it's easier not to worry about it there?

phofl: Yes both are equivalent.

Ok sounds good, moved it over.

phofl: We have to normalise the columns though, because scalar columns are a legal input

dask_expr/reductions.py
```diff
@@ -176,7 +176,19 @@ def chunk_kwargs(self):
         return {"ignore_index": self.ignore_index, **self._subset_kwargs()}
 
     def _simplify_up(self, parent):
-        return
+        if self.subset is not None and isinstance(parent, Projection):
```
mrocklin, line 179: Maybe we don't need the `isinstance(parent, Projection)` condition here.  If we have the same columns then there's no need to project, regardless of the type of the parent operation

dask_expr/tests/test_collection.py
```diff
@@ -232,6 +232,17 @@ def test_columns_traverse_filters(pdf, df):
     assert str(result) == str(expected)
 
 
+@pytest.mark.parametrize("projection", ["zz", ["zz"], ["zz", "x"], "zz"])
+@pytest.mark.parametrize("subset", ["x", ["x"]])
+def test_drop_duplicates_subset_optimizing(pdf, subset, projection):
```
mrocklin, line 237: One more note, there's been an effort in the project to keep testing times very fast (we were sub-two-seconds for a long time, that went away when we added shuffling).  Seeing many-parameter tests for simple functionality has me mildly concerned, but maybe it's appropriate at this point to start caring more about coverage than about speed.

cc @rjzamora in case he wants to weigh in

Mostly I want to convey "tests are currently fast, I'd love it if they stayed that way"

dask_expr/reductions.py
```diff
@@ -176,7 +176,19 @@ def chunk_kwargs(self):
...
+            columns = parent.columns
+            if not isinstance(columns, list) and not hasattr(columns, "dtype"):
+                columns = [columns]
```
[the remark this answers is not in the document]
phofl: No, that would change behavior. Everything that’s not a scalar would return a DataFrame instead of a series 

df[[a]] gives you back a frame instead of a Series

dask_expr/reductions.py
phofl: This might open us up to some edge case where parent isn’t a frame anymore? Saying might, not totally sure

dask_expr/tests/test_collection.py
phofl: Yeah this is mostly a coverage question, will defer to both of you there. One note though: none of these are redundant coverage wise

dask_expr/reductions.py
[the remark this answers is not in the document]
mrocklin: Hrm, indeed

```python
In [1]: import dask_expr as dx

In [2]: df = dx.datasets.timeseries()

In [3]: df.x.expr.columns
Out[3]: 'x'

In [4]: df.expr.columns
Out[4]: Index(['name', 'id', 'x', 'y'], dtype='object')
```

I wonder if we want to get cleaner here in some way.  Using `.columns` as a signal for "this is a series" seems like we're mixing a couple of things into one variable.  Maybe expr.columns should always be a pandas index (or even just a list).  

@rjzamora any thoughts?

dask_expr/reductions.py
mrocklin: Even then I'd hope that we would use `isisntance(parent.columns, pd.Index)` or something similar.  

I think that we shouldn't introduce a Projection if it doesn't help.  I could imagine future optimizations, for example, which remove unnecessary projections, and then we'd be back in a cycle.

dask_expr/tests/test_collection.py
mrocklin: Probably the right answer is to prefer coverage at this point (even if it makes me a little sad)

dask_expr/reductions.py
phofl: This doesn't work in all cases unfortunately,

if you do

```
ddf.drop_duplicates(subset="a").index.compute()
```
then parent it an ``Index`` which doesn't have a columns attribute, we could check ``hasattr(parent, "columns")``, but that is kinda ugly as well

mrocklin: Exprs don't really have a concept of frame vs series vs index.  They're all the same.  Ideally we can avoid this distinction while in expr-land entirely.

phofl: So you wouldn't do something like

```
ddf.drop_duplicates(subset="a").index
```
?

That wasn't my idea, but ``assert_eq`` is checking the index and it's meta information as well.

mrocklin: > So you wouldn't do something like

You would, but the resulting object would still just be an Expr.  In general we don't know the collection type (DataFrame/Series/Index) of exprs.  We treat them all the same in terms of what API they expose. 

This was one of the original challenges we found in early experiments.  We encountered an explosion of types like `SeriesAdd`, `FrameAdd`, `IndexAdd` which became challenging quickly.  The solution was to have two different class hierarchies, `DataFrame/Series/Index` which understood the different in APi but didn't track any user intent and the `Expr` class hierarchy, which captured what the user wrote down but didn't try to track the `DataFrame/Series/Index` difference at all.  Exprs only track what the user wrote down, and don't try to infer much from it.  Sometimes optimizations may ask `meta`, but even that is a derived quantity rather than something innate about the Expr.

phofl: Yeah this makes sense (and this is also what I understood so far). But the error I am getting is the following (this all happens on the expression level, but I might be missing something):

If I do 
```
ddf.drop_duplicates(subset="a").index
```

then ``parent`` in ``_simplify_up`` of ``DropDuplicates`` is the expression ``Index``, e.g. ``class Index(Elemwise):``. If I do not check for ``isintance(parent, Projection)``, then I will try to access ``parent.columns``. But the ``columns`` property returns ``self._meta.columns`` and ``self._meta`` is an actual ``pd.Index``, which does not have ``columns``. This is the error I am seeing. So we have to ensure somehow that ``parent`` is an expression where we can access columns in ``_simplify_up``?

dask_expr/reductions.py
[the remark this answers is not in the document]
rjzamora: Right, I remember struggling with this a bit. From the collection API, we want the behavior of `columns` to be consistent with that of a pandas object. However, for the `Expr`, it seems reasonable to always return a `list` (it could also simplify column-projection logic if we always know that `Expr.columns` will return a list).

dask_expr/tests/test_collection.py
rjzamora: Coverage should probably be the priority. However, we should certainly try to minimize run time to whatever extent we can. What is the "delta" on the new test coverage here?

phofl: The whole test runs in a bit under 30ms locally, so should be good? ``test_setitem`` takes over 100ms as a comparison.

Issue #92, opened 2023-05-18
Make uber-lyft example work


mrocklin:
I (and others) often use this example when giving Dask dataframe examples: https://github.com/coiled/examples/blob/main/uber-lyft.ipynb

I tried running it and, to my surprise, some things actually worked!  Not everything though, in particular:

-  value_counts doesn't exist
- Groupby aggregations like `df.groupby("hvfhs_license_num").tipped.mean().compute()` didn't seem to work

```python-traceback
AttributeError                            Traceback (most recent call last)
Cell In[11], line 3
      1 df["tipped"] = df.tips != 0
----> 3 df.groupby("hvfhs_license_num").tipped.mean().compute()

AttributeError: 'GroupBy' object has no attribute 'tipped'
```

My thinking is that by going after common workflows like this we'll find ourselves with a good objective function to help us prioritize.

phofl:
The notebook seems to work now. Closing here.

PR #112, opened 2023-05-22
Implement ``GroupBy.__getitem__``

phofl:
The logic is mostly from dask/dask.

The initial meta calculation couldn't work since it would always reduce to a Series with name None, passing the slice through should be more efficient. 

xref #92

### DIFF
--- a/dask_expr/groupby.py
+++ b/dask_expr/groupby.py
@@ -1,14 +1,8 @@
 import functools
 
 import numpy as np
-from dask.dataframe.core import (
-    _concat,
-    is_dataframe_like,
-    is_series_like,
-    make_meta,
-    meta_nonempty,
-    no_default,
-)
+from dask import is_dask_collection
+from dask.dataframe.core import _concat, is_dataframe_like, is_series_like, no_default
 from dask.dataframe.groupby import (
     _agg_finalize,
     _apply_chunk,
@@ -18,7 +12,7 @@ from dask.dataframe.groupby import (
     _groupby_apply_funcs,
     _normalize_spec,
 )
-from dask.utils import M
+from dask.utils import M, is_index_like
 
 from dask_expr.collection import DataFrame, new_collection
 from dask_expr.reductions import ApplyConcatApply
@@ -69,12 +63,14 @@ class SingleAggregation(ApplyConcatApply):
         "dropna",
         "chunk_kwargs",
         "aggregate_kwargs",
+        "_slice",
     ]
     _defaults = {
         "observed": None,
         "dropna": None,
         "chunk_kwargs": None,
         "aggregate_kwargs": None,
+        "_slice": None,
     }
 
     groupby_chunk = None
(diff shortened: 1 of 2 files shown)


mrocklin:
cc @rjzamora for review if you have the time

dask_expr/groupby.py
```diff
@@ -321,9 +329,27 @@ def _single_agg(
...
         )
 
+    def __getattr__(self, key):
+        try:
+            return self[key]
+        except KeyError as e:
+            raise AttributeError(e) from e
+
+    def __getitem__(self, key):
+        g = GroupBy(
```
rjzamora, line 343: We didn't feel the need to distinguish between `DataFrame`/`SeriesGroupBy` before this PR. Curious if you still think we can use the same class for both?

dask_expr/groupby.py
```diff
@@ -285,9 +275,27 @@ def __init__(
...
+        by_ = by if isinstance(by, (tuple, list)) else [by]
+        self._slice = slice
+        # Check if we can project columns
+        projection = None
+        if (
+            np.isscalar(slice)
+            or isinstance(slice, (str, list, tuple))
+            or (
+                (is_index_like(slice) or is_series_like(slice))
+                and not is_dask_collection(slice)
```
rjzamora, line 289: What are the constraints on `slice` here? Is it even possible for `slice` to be a Dask collection?

dask_expr/groupby.py
```diff
@@ -285,9 +275,27 @@ def __init__(
...
+            )
+        ):
+            projection = set(by_).union(
+                {slice} if (np.isscalar(slice) or isinstance(slice, str)) else slice
+            )
+            projection = [c for c in obj.columns if c in projection]
+
         self.by = [by] if np.isscalar(by) else list(by)
-        self.obj = obj
+        self.obj = obj[projection] if projection is not None else obj
```
rjzamora, line 298: Perhaps the implied column projection could also be added in `_simplify_down` (i.e. add a `Projection` if `self.obj` includes more columns than necessary). It seems like it could be nice to separate out the optimization from the core logic. Does that make sense?

dask_expr/groupby.py
phofl: This PR doesn't really change anything, since we control the result via ``slice``, but there are some methods that are only defined on ``SeriesGroupBy`` and not on ``DataFrameGroupBy``, so we _might_ need both classes later on.

dask_expr/groupby.py
phofl: It could be a Dask Index for example, so yes this should be possible. I took the logic from dask/dask though.

dask_expr/groupby.py
phofl: I think so, will check this tomorrow

phofl: Rick and I chatted offline. We can add this step to ``SingleAggregation`` and ``GroupbyAggregation`` but not to ``GroupBy``, since it's not an Expression. not a blocker here and will tackle in a follow up

PR #113, opened 2023-05-23
Implement ``DataFrame.rename``

phofl:
Implemented this on DataFrame, what confused me a bit is that Dask disallows mutating the Index for a DataFrame but allows it on a Series. We should maybe think high level whether this is what we want before we invest time implementing anything on the Index.

### DIFF
--- a/dask_expr/expr.py
+++ b/dask_expr/expr.py
@@ -5,6 +5,7 @@ import numbers
 import operator
 import os
 from collections import defaultdict
+from collections.abc import Mapping
 
 import dask
 import pandas as pd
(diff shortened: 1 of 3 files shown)


dask_expr/expr.py
```diff
@@ -780,6 +781,25 @@ class DropnaFrame(Blockwise):
     operation = M.dropna
 
 
+class RenameFrame(Blockwise):
+    _parameters = ["frame", "columns"]
+    _keyword_only = ["columns"]
+    operation = M.rename
+
+    def _simplify_up(self, parent):
+        if isinstance(parent, Projection) and isinstance(self.columns, Mapping):
```
mrocklin, line 790: Historically @rjzamora has used `self.operand("columns")` when we wants to refer to an operand that collides with a property name.  I'm surprised that this isn't necessary here.

phofl: Thx, this is necessary. Test coverage wasn't good enough to show this. Added a test and cleaned this up

PR #115, opened 2023-05-23
Implement ``clip``

phofl:


dask_expr/expr.py
```diff
@@ -769,6 +772,19 @@ def _task(self, index: int):
...
+    _parameters = ["frame", "lower", "upper"]
+    _defaults = {"lower": None, "upper": None}
+    operation = M.clip
+
+    def _simplify_up(self, parent):
+        if isinstance(parent, Projection):
+            if self.frame.columns.equals(parent.columns):
+                # Don't introduce unnecessary projections
+                return
+            return type(self)(self.frame[parent.operand("columns")], *self.operands[1:])
```
mrocklin, line 785: Is this pattern likely to be common?  If so, I wonder if there is some nice way to generalize it.  

We don't have to figure that out now, but it would be good to have it in our brains I think.

dask_expr/expr.py
```diff
@@ -769,6 +772,19 @@ def _task(self, index: int):
             )
 
 
+class Clip(Blockwise):
```
mrocklin, line 775:
```suggestion
class Clip(Elemwise):
```

dask_expr/tests/test_collection.py
```diff
@@ -249,6 +251,18 @@ def test_columns_traverse_filters(pdf, df):
     assert str(result) == str(expected)
 
 
+def test_clip_traverse_filters(df):
+    result = optimize(df.clip(lower=10).y, fuse=False)
+    expected = df.y.clip(lower=10)
+
+    assert str(result) == str(expected)
```
mrocklin, line 258: I used str early on before we had deterministic hashing.  I think that now the right way to do this is by comparing `._name` (although we'd also welcome a better way if you have thoughts)

dask_expr/expr.py
phofl: Yeah that's already on my Todo list, but don't have a good solution right now

PR #119, opened 2023-05-24
Implement ``align``

phofl:
Not super happy about the structure here but should be ok for a first attempt

### DIFF
--- a/dask_expr/expr.py
+++ b/dask_expr/expr.py
@@ -24,6 +24,7 @@ from dask.dataframe.core import (
 )
 from dask.dataframe.dispatch import meta_nonempty
 from dask.utils import M, apply, funcname, import_required, is_arraylike
+from tlz import merge_sorted, unique
 
 replacement_rules = []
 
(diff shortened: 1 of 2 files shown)


dask_expr/collection.py
```diff
@@ -44,6 +44,10 @@ def _wrap_expr_api(*args, wrap_api=None, **kwargs):
     )
     if isinstance(result, expr.Expr):
         return new_collection(result)
+    elif isinstance(result, tuple):
+        return tuple(
```
phofl, line 48: We could also do this in ``align`` itself

dask_expr/expr.py
```diff
@@ -871,6 +875,29 @@ class Replace(Blockwise):
     operation = M.replace
 
 
+class _Align(Blockwise):
+    _parameters = ["frame", "other", "join", "fill_value"]
+    _defaults = {"join": "outer", "fill_value": None}
+    _keyword_only = ["join", "fill_value"]
+    operation = M.align
```
mrocklin, line 882: Is this truly blockwise though?  What if one index values is on one partition in one dataframe, but in another partition in the other dataframe?

phofl: I forgot adding a Repartition call before this class. The whole operation isn't blockwise, but this class is

dask_expr/expr.py
```diff
@@ -439,6 +440,30 @@ def apply(self, function, *args, **kwargs):
...
+                "partitions. Please use `set_index` "
+                "to set the index."
+            )
+
+        divisions = list(unique(merge_sorted(*[df.divisions for df in dfs])))
+        if len(divisions) == 1:  # single value for index
+            divisions = (divisions[0], divisions[0])
+
+        left = Repartition(self, new_divisions=divisions, force=True)
+        other = Repartition(other, new_divisions=divisions, force=True)
```
mrocklin, line 460: If the divisions are the same then this should probably be a no-op.  That logic should either be in the `.repartition` method or in a `simplify_down` method of the `Repartition` class.

demo.ipynb
```diff
@@ -18,12 +18,23 @@
   },
   {
    "cell_type": "code",
-   "execution_count": null,
+   "execution_count": 39,
```
mrocklin, line 21: It looks like the notebook was modified for this PR.  Maybe an errant `git commit -a`

dask_expr/expr.py
```diff
@@ -439,6 +440,30 @@ def apply(self, function, *args, **kwargs):
...
+        aligned = _Align(left, other, join=join, fill_value=fill_value)
```
mrocklin, line 461: Interesting, so this intermediate object isn't a dataframe, it's a tuple of two dataframes?  Totally fine, just interesting.  It breaks our current model a little (but in a way that's interesting to think about).

dask_expr/expr.py
mrocklin: Happy if this is future work though.

mrocklin (approved):

dask_expr/expr.py
phofl: I'll put it on my list

dask_expr/expr.py
phofl: Yes exactly. I think this is the cleanest in the sense that we don't deviate from the pandas model. But we could certainly change this at some point to merge both classes together

phofl:
Addressed the repartition suggestion in #162, so this is ready

PR #120, opened 2023-05-24
Implement ``combine``

phofl:


dask_expr/expr.py
```diff
@@ -778,6 +778,19 @@ def _task(self, index: int):
...
+    _parameters = ["frame", "other", "func", "fill_value", "overwrite"]
+    _defaults = {"fill_value": None, "overwrite": no_default}
+    _keyword_only = ["fill_value", "overwrite"]
+    operation = M.combine
+
+    @functools.cached_property
+    def _meta(self):
+        args = [op._meta if isinstance(op, Expr) else op for op in self._args]
+        args[0] = meta_nonempty(self.frame._meta)
+        return self.operation(*args, **self._kwargs)
```
rjzamora, line 791: Was there a specific case that motivated you to avoid relying on `Blockwise._meta` here? That is, I believe that the `meta_nonempty` step could be important, but it doesn't look like it is important for the current tests.

Another note:  We will need to wrap the final output in `make_meta` to ensure that it is "empty."

phofl: Should have added a test, the function could change your dtype but thinking about this again, I'd rather have incorrect meta information than executing an udf on synthetic data, this might brake user code.

That said, combine will probably be deprecated in pandas (the Series vs DataFrame inconsistencies are very confusing), so closing this for now. Will revisit in case it isn't deprecated soonish

PR #124, opened 2023-05-25
Add mechanism to search an expression graph for a given type

rjzamora:
While starting to add a `to_parquet` implementation, I found that it would be useful to have a simple mechanism to check if a specific `Expr` type is in the expression graph. For example, when overwriting a parquet dataset, it's good to make sure that we aren't reading from the same path earlier in the graph.

### DIFF
--- a/dask_expr/expr.py
+++ b/dask_expr/expr.py
@@ -5,7 +5,7 @@ import numbers
 import operator
 import os
 from collections import defaultdict
-from collections.abc import Mapping, Set
+from collections.abc import Mapping
 
 import dask
 import pandas as pd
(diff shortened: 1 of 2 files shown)


rjzamora:
@mrocklin - I didn't think very long about the name/API added here. Please feel free to suggest something different.

dask_expr/expr.py
```diff
@@ -623,6 +623,36 @@ def visualize(self, filename="dask-expr.svg", format=None, **kwargs):
         graphviz_to_file(g, filename, format)
         return g
 
+    def find_subtype(self, subtype: type) -> Set[Expr]:
```
mrocklin, line 626: I'll suggest that this be an iterator, like `os.path.walk`.  I think that the path dependence might be interesting in the future.

Names?  🤔 I don't like `subtype`.  We're looking for instances of a type.  `find_operations`?  `find_??`

rjzamora: >I'll suggest that this be an iterator, like os.path.walk. I think that the path dependence might be interesting in the future.

Could you explain this a bit more? Are you just saying that we should return an iterator? (I get the feeling you're saying more than that)

mrocklin: I suggest that rather than return a set we yield results as we find them.

I suspect that what you have here is a depth first search ordering.  That's a sensible ordering, probably more sensible than `set`'s hashing ordering or whatnot.  

Yielding results just makes more intuitive sense to me for traversing operations like this

mrocklin: It's kind of an aesthetic thing though I guess

dask_expr/expr.py
```diff
@@ -635,31 +635,32 @@ def visualize(self, filename="dask-expr.svg", format=None, **kwargs):
...
         while stack:
             node = stack.pop()
             if node._name in seen:
                 continue
             seen.add(node._name)
 
-            if isinstance(node, subtype):
-                nodes.add(node)
+            if isinstance(node, operation):
+                nodes.append(node)
```
mrocklin, line 663: Just to be clear, my recommendation was ...

```suggestion
                yield node
```

I'm open to not doing this though if you'd like to avoid iterators.  I don't have a strong desire here and maybe this is too complicated.

rjzamora: yes, sorry. I was adding commits to do both. Sorry, you saw only the first commit so quickly :)

rjzamora: To clarify, my latest commit returns a generator, and the one you are commenting on here returns a list (to only capture depth-first ordering).  I have no strong preference either way.

Issue #71, opened 2023-05-08
Add to_parquet


mrocklin:
If we're looking to have a realistic workload then my guess is that the `read_parquet -> shuffle -> to_parquet` pipeline seems like a good first one.  @rjzamora maybe something to think about?

rjzamora:
Closed by #137

PR #125, opened 2023-05-25
Add `Expr.reset_index`

rjzamora:
**Context**: Found a need for this while pushing on #71

### DIFF
--- a/dask_expr/expr.py
+++ b/dask_expr/expr.py
@@ -283,6 +283,14 @@ class Expr:
     def index(self):
         return Index(self)
 
+    def reset_index(self, drop=False):
+        if is_dataframe_like(self._meta) or is_series_like(self._meta):
+            return ResetIndex(self, drop)
+        raise TypeError(
+            f"`reset_index` requires dataframe-like or series-like data."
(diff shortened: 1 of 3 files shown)


dask_expr/expr.py
```diff
@@ -283,6 +283,14 @@ def optimize(self, **kwargs):
     def index(self):
         return Index(self)
 
+    def reset_index(self, drop=False):
+        if is_dataframe_like(self._meta) or is_series_like(self._meta):
+            return ResetIndex(self, drop)
```
rjzamora, line 288: We can either add `Expr.reset_index` with a `_meta` check like this, **or** we can add `reset_index` to both the `DataFrame` and `Series` APIs. Wasn't sure which was "better" in this case.

mrocklin: I'll suggest that we start deferring API to the collections whenever there's any ambiguity.  Are you ok with that convention?

mrocklin (approved):

dask_expr/expr.py
rjzamora: Yeah, that works for me.

PR #127, opened 2023-05-31
Pass reductions through shuffles

mrocklin:
This is here mostly for comments.  In many (but not all) cases we should be able to pass reductions through shuffle operations.  For example:

```python
len(df.set_index(...)) == len(df)
```

This is easy to write, although we probably want to allow-list these rather than disallow them as we do today (IdxMax is a good example of a reduction where this isn't doable).

Additionally, we should also be able to do this with a projection in the middle:

```python
df.set_index(...).x.sum() == df.x.sum()
```

However we can't actually express this with today's optimizations because `_simplify_up` doesn't reach far enough up.  Interestingly this is a case where matchpy would have given us more than we can achieve with our current optimization infrastructure.

Probably short-term we should do the simple things and merge a cleaner version of this in.  I thought I'd leave it in this state though for discussion for a moment first.

### DIFF
--- a/dask_expr/shuffle.py
+++ b/dask_expr/shuffle.py
@@ -22,6 +22,7 @@ from dask_expr.reductions import (
     All,
     Any,
     Count,
+    DropDuplicates,
     Len,
     Max,
     Mean,
@@ -125,6 +126,7 @@ class Shuffle(Expr):
             parent,
             (
                 Unique,
+                DropDuplicates,
                 Sum,
                 Prod,
                 Max,
--- a/dask_expr/shuffle.py
+++ b/dask_expr/shuffle.py
@@ -18,7 +18,25 @@ from dask.dataframe.shuffle import (
 from dask.utils import digit, get_default_shuffle_algorithm, insert
 
 from dask_expr.expr import Blockwise, Expr, PartitionsFiltered, Projection
-from dask_expr.reductions import Reduction
+from dask_expr.reductions import (
+    All,
+    Any,
+    Count,
+    Len,
+    Max,
+    Mean,
+    MemoryUsage,
+    Min,
+    Mode,
+    NBytes,
+    NLargest,
+    NSmallest,
+    Prod,
+    Size,
+    Sum,
+    Unique,
+    ValueCounts,
+)
 from dask_expr.repartition import Repartition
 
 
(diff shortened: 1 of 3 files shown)


dask_expr/shuffle.py
```diff
@@ -102,6 +103,9 @@ def _simplify_up(self, parent):
                     parent.operand("columns")
                 ]
 
+        if isinstance(parent, Reduction) and "idx" not in type(parent).__name__:
```
mrocklin, line 106: I think that this should be changed to have an explicit list of allowable reductions, rather than assume everything unseen is fine.

dask_expr/tests/test_shuffle.py
```diff
@@ -114,3 +114,12 @@ def test_shuffle_column_projection():
...
     assert "y" not in df2.expr.operands[0].columns
+
+
+def test_shuffle_reductions():
+    pdf = pd.DataFrame({"x": list(range(20)) * 5, "y": range(100)})
+    df = from_pandas(pdf, npartitions=10)
+
+    assert df.shuffle("x").sum().optimize()._name == df.sum()._name  # This passes
+
+    assert df.shuffle("x").y.sum().optimize()._name == df.y.sum()._name  # This fails
```
mrocklin, line 125: We can't support this right now 😢

mrocklin: That would solve this problem, but it doesn't feel ideal.  I think that the underlying issue here is that we can't write optimizations that look up more than one expression.  I think I'm going to just fail to do this well today for now and leave it as future work.

dask_expr/tests/test_shuffle.py
```diff
@@ -114,3 +114,12 @@ def test_shuffle_column_projection():
...
     assert "y" not in df2.expr.operands[0].columns
+
+
+def test_shuffle_reductions():
+    pdf = pd.DataFrame({"x": list(range(20)) * 5, "y": range(100)})
+    df = from_pandas(pdf, npartitions=10)
+
+    assert df.shuffle("x").sum().optimize()._name == df.sum()._name  # This passes
+
+    # assert df.shuffle("x").y.sum().optimize()._name == df.y.sum()._name  # This fails
```
rjzamora, line 125: Maybe we should XFAIL this?

rjzamora (approved):

dask_expr/tests/test_shuffle.py
mrocklin: Yeah, I'm fine with this not performing great for a while in a way that I think we weren't ok with IO not performing great.

dask_expr/tests/test_shuffle.py
mrocklin: Sure.   Done.

PR #129, opened 2023-06-01
Add optimization to remove needless projections

mrocklin:


dask_expr/expr.py
```diff
@@ -1042,6 +1042,9 @@ def __str__(self):
         return f"{base}[{repr(self.columns)}]"
 
     def _simplify_down(self):
+        if str(self.frame.columns) == str(self.columns):
+            # TODO: we should get more precise around Expr.columns types
+            return self.frame
```
rjzamora, line 1047: I think it is generally a good idea to drop redundant column projections.  I also agree that this is related to our earlier discussions about formalizing the output of `Expr.columns`.

At one point we had loosely agreed that we should always be returning a `list`. In that case, we would need to be careful about dropping projections that are converting dataframe-like expressions into series-like expressions. I'm pretty sure this will be as simple as checking that both `Expr.columns` and `Expr.ndim` match.

mrocklin: Yeah, I think that this should be cleaner once we determine a more strict definition around `.columns`.  In the meantime, `str` seems to do everything we want here I think.

rjzamora: yeah, I think `str` is a reasonable workaround for now. It shouldn't be difficult to modify this when we formalize `.columns`

PR #134, opened 2023-06-02
Len(df) -> Len(df.index), also Filter-Index optimization

mrocklin:


mrocklin:
I found that operations like 

```python
df = dx.read_parquet(...)
df = df[df.x + df.y > 1]
len(df)
```

Caused a complete read of all of the data.

Now `Len(Filter)` turns into `Len(Index(Filter))`, and we also make `Index(ReadParquet)` do zero-columns

mrocklin:
@rjzamora I'd welcome your thoughts on this.  

It's obviously not good to merge.  I think it conflicts with some of the other parquet optimizations.  If I recall correctly the conflict was interesting.

dask_expr/io/parquet.py
```diff
@@ -102,6 +103,13 @@ def columns(self):
             return pd.Index(_list_columns(columns_operand))
 
     def _simplify_up(self, parent):
+        if isinstance(parent, Index):
+            # Column projection
+            operands = list(self.operands)
+            operands[self._parameters.index("columns")] = []
+            operands[self._parameters.index("_series")] = False
+            return ReadParquet(*operands)
```
rjzamora, line 111: This feels similar to a stalled experiment of mine in dask/dask: https://github.com/dask/dask/pull/10072#issuecomment-1470962862

In that PR, I found that making an empty column projection here can be dangerous.

mrocklin:
I think that this is good to go if either @rjzamora or @phofl wants to review.  I'll probably self-merge in 24 hours if no one comments.

phofl (approved):
This looks good to me. Can't comment on the comment of @rjzamora

dask_expr/io/parquet.py
rjzamora: I like the optimization, and have no immediate concerns. I'm just pointing this out in case we end up seeing strange `len`/`index` behavior.

mrocklin: Grand.  Thanks!

PR #136, opened 2023-06-05
Implement `fillna` (simplified)

j-bennet:
Part of https://github.com/mrocklin/dask-expr/issues/68.

This is a simplest possible version, without `method`, `axis`, or `limit`.

### DIFF
--- a/dask_expr/collection.py
+++ b/dask_expr/collection.py
@@ -570,6 +570,11 @@ class DataFrame(FrameBase):
             expr.DropnaFrame(self.expr, how=how, subset=subset, thresh=thresh)
         )
 
+    def fillna(self, value=None, method=None, limit=None, axis=None):
+        return new_collection(
+            expr.Fillna(self.expr, value=value, method=method, limit=limit, axis=axis)
+        )
+
     def rename(self, columns):
         return new_collection(RenameFrame(self.expr, columns=columns))
 
@@ -619,6 +624,11 @@ class Series(FrameBase):
 
     def dropna(self):
         return new_collection(expr.DropnaSeries(self.expr))
+
+    def fillna(self, value=None, method=None, limit=None, axis=None):
+        return new_collection(
+            expr.Fillna(self.expr, value=value, method=method, limit=limit, axis=axis)
+        )
 
     def between(self, left, right, inclusive="both"):
         return new_collection(
--- a/dask_expr/collection.py
+++ b/dask_expr/collection.py
@@ -693,11 +693,6 @@ class DataFrame(FrameBase):
             expr.Sample(self.expr, state_data=state_data, frac=frac, replace=replace)
         )
 
-    def fillna(self, value=None, method=None, limit=None):
-        return new_collection(
-            expr.Fillna(self.expr, value=value, method=method, limit=limit)
-        )
-
     def rename(self, columns):
         return new_collection(expr.RenameFrame(self.expr, columns=columns))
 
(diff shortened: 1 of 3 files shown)


dask_expr/collection.py
```diff
@@ -570,6 +570,11 @@ def dropna(self, how=no_default, subset=None, thresh=no_default):
             expr.DropnaFrame(self.expr, how=how, subset=subset, thresh=thresh)
         )
 
+    def fillna(self, value=None, method=None, limit=None, axis=None):
```
phofl, line 573: This can go to ``Expr``. We try to add things there as long as the signatures are consistent. 

We did not add the axis keyword for now to keep it simple, you can keep it out here too if you don't want to deal with it now.

dask_expr/expr.py
```diff
@@ -853,6 +853,13 @@ class DropnaFrame(Blockwise):
     operation = M.dropna
 
 
+class Fillna(Blockwise):
+    _parameters = ["frame", "value", "method", "limit", "axis"]
+    _defaults = {"value": None, "method": None, "limit": None, "axis": None}
+    _keyword_only = ["method", "limit", "axis"]
+    operation = M.fillna
```
phofl, line 860: Unfortunately this won't work here (at least not in all cases). We can not inherit from ``Blockwise``. Fillna is implemented using ``map_overlap`` in dask/dask, which is necessary for properly filling NaN values when method is set and one block finished with ``NaN``. That's one of the reasons I skipped it for now.

phofl (commented):
Thx for taking a stab at this, couple of comments.

dask_expr/expr.py
j-bennet: Ah. I should have known that things are not so simple, or you would have implemented it already.

Do you have a reproduction case for when it would not work?

phofl: Something like:

```
df = pd.DataFrame({"a": [1, 2, np.nan, 3]})
ddf = from_pandas(df, npartitions=2)
ddf.fillna(method="ffill")
```

should reproduce. First row from one of the partitions should be missing and you have to specify a method that relies on the previous values.

j-bennet:
Simplified it to make it Blockwise.

dask_expr/expr.py
```diff
@@ -983,6 +986,12 @@ def _simplify_up(self, parent):
             )
 
 
+class Fillna(Blockwise):
```
mrocklin, line 989: Blockwise is correct, but Elemwise is more correct (this operation operates block by block, but also row-by-row).  This is useful for some optimizations.  For example `Len(Fillna(x)) -> Len(x)`

```suggestion
class Fillna(Elemwise):
```

dask_expr/tests/test_collection.py
```diff
@@ -154,6 +154,14 @@ def test_dropna(pdf):
     assert_eq(df.y.dropna(), pdf.y.dropna())
 
 
+def test_fillna():
+    pdf = pd.DataFrame({"x": [1, 2, None, None, 5, 6]})
+    df = from_pandas(pdf, npartitions=2)
+    actual = df.fillna(value=100).compute()
```
mrocklin, line 160: It's better not to call `.compute()` explicitly in Dask tests unless we have to. Instead we let `assert_eq` do this.  That way `assert_eq` can do things like verify that the columns and dtypes that Dask thought it would get are what actually happened.  It can also run various sanity checks on the task graph to make sure that names are sensible and so on.

If you call compute ahead of time then `assert_eq` doesn't have the Dask graph to actually ask questions of

```suggestion
    actual = df.fillna(value=100)
```

mrocklin (commented):
Thanks @j-bennet !  A couple of code suggestions below.  In general this seems good to me though.  It looks like we're implementing a simple version of the operation first and we can add more methods support in the future.

phofl:
> It looks like we're implementing a simple version of the operation first and we can add more methods support in the future.

yep, @j-bennet and I chatted earlier and I think having fillna without limit is good enough for a bit (method is deprecated in pandas anyway)

phofl (approved):

phofl:
thx @j-bennet

PR #137, opened 2023-06-09
Implement `DataFrame.to_parquet`

rjzamora:
Note that I experimented with a few variations on this, and ultimately decided to keep the `Expr` component of `to_parquet` as simple as possible.

### DIFF
--- a/dask_expr/collection.py
+++ b/dask_expr/collection.py
@@ -573,6 +573,11 @@ class DataFrame(FrameBase):
     def rename(self, columns):
         return new_collection(RenameFrame(self.expr, columns=columns))
 
+    def to_parquet(self, path, **kwargs):
+        from dask_expr.io.parquet import to_parquet
+
+        return to_parquet(self, path, **kwargs)
+
 
 class Series(FrameBase):
     """Series-like Expr Collection"""
--- a/dask_expr/io/parquet.py
+++ b/dask_expr/io/parquet.py
@@ -28,6 +28,49 @@ def _list_columns(columns):
     return columns
 
 
+class ToParquet(Expr):
+    _parameters = [
+        "frame",
+        "path",
+        "compression",
+        "write_index",
+        "append",
+        "overwrite",
+        "ignore_divisions",
+        "partition_on",
+        "storage_options",
+        "custom_metadata",
+        "write_metadata_file",
+        "compute",
+        "compute_kwargs",
+        "schema",
+        "name_function",
+        "filesystem",
+        "kwargs",
+    ]
+    _defaults = {
+        "compression": "snappy",
+        "write_index": True,
+        "append": False,
+        "overwrite": False,
+        "ignore_divisions": False,
+        "partition_on": None,
+        "storage_options": None,
+        "custom_metadata": None,
+        "write_metadata_file": None,
+        "compute": True,
+        "compute_kwargs": None,
+        "schema": "infer",
+        "name_function": None,
+        "filesystem": None,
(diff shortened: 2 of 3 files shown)


dask_expr/io/parquet.py
```diff
@@ -62,6 +68,309 @@ def normalize_pa_schema(schema):
...
+                        self.path,
+                    ],
+                    {"append": append, "compression": compression},
+                )
+            }
+        else:
+            return {(self._name, 0): (lambda x: None, self.frame.__dask_keys__())}
+
+
+def to_parquet(
```
rjzamora, line 155: NOTE: This function is mostly copied from `dd.DataFrame.to_parquet`. I didn't find any benefit in moving any option validation of filesystem/overwriting logic into an `Expr` class.

dask_expr/io/parquet.py
```diff
@@ -62,6 +68,309 @@ def normalize_pa_schema(schema):
...
+                engine,
+                i_offset,
+                partition_on,
+                write_metadata_file,
+                name_function,
+                toolz.merge(
+                    kwargs,
+                    {"compression": compression, "custom_metadata": custom_metadata},
+                    extra_write_kwargs,
+                ),
```
rjzamora, line 359: Only change to `to_parquet` is here.

dask_expr/io/parquet.py
```diff
@@ -62,6 +68,309 @@ def normalize_pa_schema(schema):
...
+            # output directory, since deleting those files now would result in
+            # errors or incorrect results.
+            for read_op in df.expr.find_operations(ReadParquet):
+                read_path_with_slash = str(read_op.path).rstrip("/") + "/"
+                write_path_with_slash = path.rstrip("/") + "/"
+                if read_path_with_slash.startswith(write_path_with_slash):
+                    raise ValueError(
+                        "Cannot overwrite a path that you are reading "
+                        "from in the same task graph."
+                    )
```
rjzamora, line 225: and here.

dask_expr/io/parquet.py
```diff
@@ -62,6 +68,309 @@ def normalize_pa_schema(schema):
...
+class ToParquetBarrier(Expr):
+    _parameters = ToParquet._parameters
+
+    @property
+    def _meta(self):
+        return None
+
+    def _divisions(self):
+        return (None, None)
+
```
mrocklin, line 133: Maybe we need to be ok with `Expr`s that aren't dataframes.  

Nothing to do at the moment, but this reinforces a questions we've had for a while.  Just pointing it out here.

dask_expr/collection.py
```diff
@@ -573,6 +573,11 @@ def dropna(self, how=no_default, subset=None, thresh=no_default):
...
         return new_collection(RenameFrame(self.expr, columns=columns))
 
+    def to_parquet(self, path, **kwargs):
+        from dask_expr.io.parquet import to_parquet
+
+        return to_parquet(self, path, **kwargs)
+
 
 class Series(FrameBase):
     """Series-like Expr Collection"""
```
mrocklin, line 583: Is the implementation above only on DataFrames?  If so do we need a Series implementation (which I assume is just `to_frame().to_parquet(...)`

mrocklin (approved):
Broadly it seems fine to me.  Thanks for doing this @rjzamora

dask_expr/collection.py
mrocklin: Ah, ok, that makes sense

PR #139, opened 2023-06-12
Implement ``DataFrame.sample``

phofl:
Do we have a solution for choosing a specific argument per partition? This looks a bit clunky.

### DIFF
--- a/dask_expr/expr.py
+++ b/dask_expr/expr.py
@@ -886,6 +886,24 @@ class RenameFrame(Blockwise):
             return type(self)(self.frame[columns], *self.operands[1:])
 
 
+class Sample(Blockwise):
+    _parameters = ["frame", "state_data", "frac", "replace"]
+    operation = staticmethod(methods.sample)
+
+    @functools.cached_property
+    def _meta(self):
+        args = [self.operands[0]._meta] + [self.operands[1][0]] + self.operands[2:]
+        return self.operation(*args)
+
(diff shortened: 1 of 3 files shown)


dask_expr/expr.py
```diff
@@ -886,6 +886,24 @@ def _simplify_up(self, parent):
...
+    @functools.cached_property
+    def _meta(self):
+        args = [self.operands[0]._meta] + [self.operands[1][0]] + self.operands[2:]
+        return self.operation(*args)
+
+    def _task(self, index: int):
+        args = (
+            [self._blockwise_arg(self.operands[0], index)]
+            + [self.operands[1][index]]
+            + self.operands[2:]
```
mrocklin, line 902: I would use names here:

```suggestion
            [self._blockwise_arg(self.frame, index)]
            + [self.state_data[index], self.frac, self.replace]
```

mrocklin (commented):
> Do we have a solution for choosing a specific argument per partition? This looks a bit clunky.

Can you say more?  I'm not sure I understand.  Or maybe implicitly I'm saying that this doesn't look that clunky to me.

phofl:
I am happy if this solution is ok :) Just wasn't sure whether we have a general class that might handle this.

mrocklin:
You mean the `self.state_data[index]` thing?  No, I think that the code here is pretty simple and reflective of what we want to do.

PR #141, opened 2023-06-12
Set default of ``numeric_only`` to ``False``

phofl:
Same as for the regular reductions


mrocklin:
You're in charge here.  Merge at will.

PR #149, opened 2023-06-13
Implement ``_simplify_up`` for ``assign``

phofl:
Ran into this with the taxi dataset. Something like:

```
df = read_parquet(...)
df["tipped"] = df.tips > 0
df.groupby(....).agg({"tipped": "mean", "base_passenger_fare": "sum"})
```

doesn't push the projection through, which hinders us when demoing somewhere (my assumption)

### DIFF
--- a/dask_expr/expr.py
+++ b/dask_expr/expr.py
@@ -1015,6 +1015,14 @@ class Assign(Elemwise):
     def _node_label_args(self):
         return [self.frame, self.key, self.value]
 
+    def _simplify_up(self, parent):
+        if isinstance(parent, Projection):
+            if self.key not in parent.columns:
+                return self.frame[parent.columns]
+
+            columns = list(set(parent.columns) - {self.key})
+            return type(self)(self.frame[columns], *self.operands[1:])
+
 
 class Filter(Blockwise):
     _parameters = ["frame", "predicate"]
--- a/dask_expr/expr.py
+++ b/dask_expr/expr.py
@@ -1104,12 +1104,7 @@ class Assign(Elemwise):
 
     def _simplify_up(self, parent):
         if isinstance(parent, Projection):
-            columns = (
-                parent.columns
-                if isinstance(parent.columns, pd.Index)
-                else [parent.columns]
-            )
-            columns = set(columns) - {self.key}
+            columns = set(parent.columns) - {self.key}
             if columns == set(self.frame.columns):
                 # Protect against pushing the same projection twice
                 return
--- a/dask_expr/expr.py
+++ b/dask_expr/expr.py
@@ -1017,11 +1017,20 @@ class Assign(Elemwise):
 
     def _simplify_up(self, parent):
         if isinstance(parent, Projection):
-            if self.key not in parent.columns:
-                return self.frame[parent.columns]
+            columns = (
+                parent.columns
+                if isinstance(parent.columns, pd.Index)
+                else [parent.columns]
+            )
+            columns = set(columns) - {self.key}
+            if columns == set(self.frame.columns):
+                # Protect against pushing the same projection twice
+                return
 
-            columns = list(set(parent.columns) - {self.key})
-            return type(self)(self.frame[columns], *self.operands[1:])
+            return type(parent)(
+                type(self)(self.frame[sorted(columns)], *self.operands[1:]),
+                *parent.operands[1:],
+            )
 
 
 class Filter(Blockwise):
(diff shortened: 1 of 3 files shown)


mrocklin:
> Are we okay with simply removing whole operations if they are not necessary?

I think I'm fine with this.  Presumably people focused on side effects will be sad, but I think that's fine.

dask_expr/expr.py
```diff
@@ -1015,6 +1015,23 @@ class Assign(Elemwise):
     def _node_label_args(self):
         return [self.frame, self.key, self.value]
 
+    def _simplify_up(self, parent):
+        if isinstance(parent, Projection):
+            columns = (
+                parent.columns
+                if isinstance(parent.columns, pd.Index)
+                else [parent.columns]
```
mrocklin, line 1023: I'm becoming increasingly concerned about the `list/pd.Index/string` type uncertainty on columns.

If you have any interest in looking into this and making a proposal I think it would be time well spent (I can do it too though).

phofl: Can do it tomorrow!

phofl:
> > Are we okay with simply removing whole operations if they are not necessary?
> 
> I think I'm fine with this. Presumably people focused on side effects will be sad, but I think that's fine.

Sounds good in general, this killed one of our tests though and the failure wasn’t immediately obvious. That’s why I removed it for now, will take a closer look in the coming days

phofl:
Refactored to get rid of the str/Index inconsistency, so should be ready for another look

dask_expr/expr.py
```diff
@@ -1102,6 +1102,18 @@ class Assign(Elemwise):
     def _node_label_args(self):
         return [self.frame, self.key, self.value]
 
+    def _simplify_up(self, parent):
+        if isinstance(parent, Projection):
+            columns = set(parent.columns) - {self.key}
```
rjzamora, line 1107: **Just a thought**: If `self.key` is **not** in `parent.columns`, then perhaps the `Assign` expression should be dropped altogether? In that case, we are essentially assigning a column and then immediately dropping it.

rjzamora: You don't need to worry about this now. Just noting that it would be a relatively simple optimization to add.

dask_expr/tests/test_collection.py
```diff
@@ -769,3 +769,21 @@ def test_align_unknown_partitions():
...
     with pytest.raises(ValueError, match="Not all divisions"):
         df.align(df)
+
+
+def test_assign_simplify(pdf):
+    df = from_pandas(pdf)
+    df2 = from_pandas(pdf)
+    df["new"] = df.x > 1
+    result = optimize(df[["x", "new"]], fuse=False)
+    expected = optimize(df2[["x"]].assign(new=df2.x > 1), fuse=False)
```
rjzamora, line 779: I think I prefer using `.simplify()` instead of `optimize(..., fuse=False)` here, but either way is technically fine. (I realize we started using `optimize` in other places before `simplify` was added)

dask_expr/tests/test_collection.py
```diff
@@ -769,3 +769,21 @@ def test_align_unknown_partitions():
...
+    assert result._name == expected._name
+
+
+def test_assign_simplify_series(pdf):
+    df = from_pandas(pdf)
+    df2 = from_pandas(pdf)
+    df["new"] = df.x > 1
+    result = optimize(df.new, fuse=False)
+    expected = optimize(df2[[]].assign(new=df2.x > 1).new, fuse=False)
+    assert result._name == expected._name
```
rjzamora, line 789: Do we want to use `assert_eq` anywhere, or do you feel that the result of `assign` is already well tested?

rjzamora (approved):
I left a few thoughts and suggestion, but the PR also looks good as is.

dask_expr/expr.py
phofl: I was struggling with this when putting up the PR, but I think I fixed the bug that caused my issues since then in another PR. Works like a charm now!

dask_expr/tests/test_collection.py
phofl: Probably tested reasonably well, but the actual check is cheap so shouldn't hurt to add it

dask_expr/tests/test_collection.py
phofl: Changed to simplify, should probably make a follow-up PR to clean this up globally.

dask_expr/expr.py
rjzamora: Nice!

PR #151, opened 2023-06-14
Change columns of Projection to return an Index

phofl:
We've discussed this a couple of weeks ago. Generally, we can make ``.columns`` always return and Index/list (Index makes sense in case we want to use the built-in methods (difference, union, intersection), otherwise we can simply use a list @rjzamora could you weigh in here whether using Index specific methods is ok for cudf?)

Most cases don't care about the difference for columns, e.g. whether it's a list-like or a scalar. These paths can just use ``parent.columns`` and be done with it. We have to be careful that we don't use ``parent.columns`` when pushing a projection up, since this would change behaviour, e.g.

```
df["a"] -> Series
df[["a"]] -> DataFrame
```

Using ``parent.operand("columns")`` consistently solves this problem.

We have a few other cases where the difference between a list-like and a scalar matters, e.g.:

```
pdf = pd.DataFrame({"a": [1, 2, 3]})
df = from_pandas(pdf)

df[["a"]] -> this should kill the projection since it is a no-op
df["a"] -> this should preserve the projection, since we coerce to a Series
```

Simply checking ``self.frame.columns == self.columns`` won't work anymore, since we cast ``"a"`` to ``pd.Index(["a"])``, which makes them seem equal. 

There are probably a few ways in which we can solve this, the one I've chosen for now is to check ``type(self.frame._meta) == type(self._meta)`` for equality. This is a pattern we should probably generalise, since we will need this in different places.

### DIFF
--- a/dask_expr/repartition.py
+++ b/dask_expr/repartition.py
@@ -84,7 +84,7 @@ class Repartition(Expr):
     def _simplify_up(self, parent):
         # Reorder with column projection
         if isinstance(parent, Projection):
-            return type(self)(self.frame[parent.columns], *self.operands[1:])
+            return type(self)(self.frame[parent.operand("columns")], *self.operands[1:])
 
 
 class RepartitionToFewer(Repartition):
--- a/dask_expr/datasets.py
+++ b/dask_expr/datasets.py
@@ -72,10 +72,7 @@ class Timeseries(PartitionsFiltered, BlockwiseIO):
 
     def _simplify_up(self, parent):
         if isinstance(parent, Projection) and len(self.dtypes) > 1:
-            if isinstance(parent.columns, (list, pd.Index)):
-                dtypes = {col: self.operand("dtypes")[col] for col in parent.columns}
-            else:
-                dtypes = {parent.columns: self.operand("dtypes")[parent.columns]}
+            dtypes = {col: self.operand("dtypes")[col] for col in parent.columns}
             out = Timeseries(
                 self.start,
                 self.end,
--- a/dask_expr/expr.py
+++ b/dask_expr/expr.py
@@ -875,7 +875,7 @@ class RenameFrame(Blockwise):
             self.operand("columns"), Mapping
         ):
             reverse_mapping = {val: key for key, val in self.operand("columns").items()}
-            if not isinstance(parent.columns, pd.Index):
+            if is_series_like(parent._meta):
                 # Fill this out when Series.rename is implemented
                 return
             else:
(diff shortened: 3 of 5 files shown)


rjzamora:
I'm going to think on this a bit more, but my initial reaction is to prefer `Expr.columns -> list` over anything pandas-specific. My preference is more about an interest in keeping `Expr` general than it is about causing problems for `cudf`.  That said, I can probably be persuaded otherwise.

>the one I've chosen for now is to check type(self.frame._meta) == type(self._meta) for equality

Can we just check that `.columns` are equal and `.ndim` match?

phofl:
> Can we just check that .columns are equal and .ndim match?

good point, this should work. Might not when we reduce to a scalar, but this is broken right now anyway, so will add the ndim check for now. Also not sure whether we can hit this path in that case.

I am neutral regarding Index vs list as columns, but this is a bit more complicated to change, since in ``assert_eq`` we expected ``dsk.columns`` to be an Index, so I would rather do this as a follow-up when we are certain that this is what we want.

rjzamora:
>since in assert_eq we expected dsk.columns to be an Index

Ah right - Now I recall that this is the primary reason `Expr.columns` currently returns an Index. Hopefully we will be able to simply convert to Index in `DataFrame.columns`

> I would rather do this as a follow-up when we are certain that this is what we want

Makes sense to me

phofl:
@rjzamora Are you happy with this PR apart from this? I'll poke around the list/Index stuff in a follow-up

dask_expr/repartition.py
```diff
@@ -84,7 +84,7 @@ def _simplify_down(self):
     def _simplify_up(self, parent):
         # Reorder with column projection
         if isinstance(parent, Projection):
-            return type(self)(self.frame[parent.columns], *self.operands[1:])
+            return type(self)(self.frame[parent.operand("columns")], *self.operands[1:])
```
rjzamora, line 87: **Side thought** (nothing to do here): I wonder if it makes sense to avoid doing this if/when `partition_size` is used (once it is implemented).

rjzamora (approved):
This makes sense to me. I'd like to see the pandas-specific `Expr.columns` type go away, but agree that it doesn't need to be in *this* PR.

phofl:
Thx for taking a look. Will look into the pd.Index stuff tomorrow

mrocklin:
@rjzamora I think that I agree with you about preferring list over pd.Index, but for different reasons.  I like lists because they're simpler / lower tech (and I like low-tech things).

I'm curious, for cudf, do you all store the column names on the GPU in a `cudf.Index`?  It seems to me that using a `pd.Index` might be better for cudf in this case.  Presumably the CPU is better at metadata.  Was there some reason why cuDF deviated from `pd.Index` here?

rjzamora:
Using pd.Index is no problem for cudf. My distaste for using something pandas specific is that it is more-specific (and “higher tech”) than it needs to be. So, I’m probably thinking along similar lines as you are.

PR #153, opened 2023-06-14
Add ``join``

phofl:
on top of #152


rjzamora (approved):
Looks good to me. **Disclaimer**: I almost always use `merge` instead of `join`, so I don't have much (if any) knowledge about the subtle differences between the two methods :)

phofl:
I'll go ahead and merge, happy to address potential comments in a follow up

PR #154, opened 2023-06-15
Implement ``concat``

phofl:


dask_expr/expr.py
```diff
@@ -494,6 +504,8 @@ def __dask_graph__(self):
             for operand in expr.operands:
                 if isinstance(operand, Expr):
                     stack.append(operand)
+                elif not isinstance(expr, Fused) and isinstance(operand, list):
```
phofl, line 507: I think the check for ``list`` is too generic (same above). We could convert a list of Expr objects into a custom object that we could check for here (just as one example).

Is there already a solution for handling a list of ``Expr`` objects for one argument? ``concat([df, df2])`` or ``join`` for example

dask_expr/collection.py
```diff
@@ -803,3 +804,46 @@ def read_parquet(
...
+        raise ValueError("No objects to concatenate")
+    if len(dfs) == 1:
+        return dfs[0]
+
+    if join not in ("inner", "outer"):
+        raise ValueError("'join' must be 'inner' or 'outer'")
+
+    try:
+        # remove any empty DataFrames
+        dfs = [df for df in dfs if len(df.columns) > 0]
```
mrocklin, line 836: This should only be in the axis=1 case?

dask_expr/concat.py
```diff
@@ -0,0 +1,121 @@
...
+            ignore_order=self.ignore_order,
+            _kwargs=self._kwargs,
+        )
+
+    def _simplify_up(self, parent):
+        # TODO: implement passing projections through
+        return
+
+
+class StackPartition(Concat):
```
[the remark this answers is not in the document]
phofl: Yes exactly,

- axis=1
- interleave_partitions=True
- ignore_unknown_divisions

Although the first two go through the same path in dask/dask

phofl:
@mrocklin gentle ping. I think this is ready for another look

dask_expr/collection.py
```diff
@@ -803,3 +804,47 @@ def read_parquet(
...
+        raise TypeError("dfs must be a list of DataFrames/Series objects")
+    if len(dfs) == 0:
+        raise ValueError("No objects to concatenate")
+    if len(dfs) == 1:
+        return dfs[0]
+
+    if join not in ("inner", "outer"):
+        raise ValueError("'join' must be 'inner' or 'outer'")
+
+    if axis == 1:
```
mrocklin, line 834: You've got `if axis == 1` a couple times here.  I guess we have an optimization that will be useful once we turn on this capability?

dask_expr/concat.py
```diff
@@ -0,0 +1,121 @@
...
+from dask_expr.expr import AsType, Expr
+
+
+class Concat(Expr):
+    _parameters = ["join", "ignore_order", "_kwargs"]
+    _defaults = {"join": "outer", "ignore_order": False, "_kwargs": {}}
+
+    @property
+    def _frame(self):
+        return self.operands[len(self._parameters) :]
```
mrocklin, line 18: Pluralize into `_frames`?

You could also use `.dependencies()` here instead, which is just all `Expr`s in operands.  Maybe that's cleaner.

dask_expr/concat.py
```diff
@@ -0,0 +1,121 @@
...
+
+    @functools.cached_property
+    def _meta(self):
+        meta = make_meta(
+            methods.concat(
+                [meta_nonempty(df._meta) for df in self._frame],
```
mrocklin, line 24: I'm curious, why nonempty?  What broke?

dask_expr/concat.py
```diff
@@ -0,0 +1,121 @@
...
+    _defaults = {"join": "outer", "ignore_order": False, "_kwargs": {}}
+
+    def _layer(self):
+        dsk, i = {}, 0
+        for df in self._frame:
+            try:
+                check_meta(df._meta, self._meta)
+                match = True
+            except (ValueError, TypeError):
+                match = False
```
mrocklin, line 102: Did you already handle this above with the `astype` calls?

dask_expr/concat.py
phofl: The dtype inference in pandas is buggy with empty dfs, so would rather not rely on it

dask_expr/collection.py
phofl: I think they are all gone for now. Will try to integrate this better when implementing axis=1

PR #156, opened 2023-06-15
Handle shuffle method name for upcoming ``dask`` release

jrbourbeau:
This method will change names in the upcoming `dask` release next week (xref https://github.com/dask/dask/pull/10344). Handling that name change early here.

### DIFF
--- a/ci/environment.yml
+++ b/ci/environment.yml
@@ -4,7 +4,9 @@ channels:
 dependencies:
   - pytest
   - pytest-cov
-  - dask
+  - dask  # overridden by git tip below
   - pyarrow
   - pandas>=2
   - pre-commit
+  - pip:
+      - git+https://github.com/dask/dask
--- a/dask_expr/shuffle.py
+++ b/dask_expr/shuffle.py
@@ -15,7 +15,13 @@ from dask.dataframe.shuffle import (
     shuffle_group_2,
     shuffle_group_get,
 )
-from dask.utils import digit, get_default_shuffle_algorithm, insert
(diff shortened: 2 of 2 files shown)


phofl:
Personally, I'd prefer not adding too many of these version consistency things at this stage. I am happy as long as we are compliant with the latest Dask release until we are in a more mature state.

rjzamora:
@phofl makes a reasonable point. We don't really have "users" yet. So, perhaps it's best to test against `main` and simply fix breakages like this as they show up?

phofl:
I'd rather test against the latest release, this makes running benchmarks and stuff a bit easier without having to install from main all the time. We release all 2 weeks in dask, so this shouldn't be too bad?

rjzamora:
I wouldn't want to make benchmarking too tricky.

I was just thinking that it becomes tricky to use `dask:main` with `dask-expr:main` (something I personally do often) if we  fix breakages after a dask release.

One possibility is to follow the convention used in this PR, but to immediately raise an issue to remove the `try`/`except` after the next release?

phofl:
Then let's go with main, my opinion isn't strong here.

rjzamora:
>Then let's go with main, my opinion isn't strong here.

I pushed a commit to move to `main`. I also don't have super strong opinions about this, so we can always move back to the latest release (or agree to handle both `main` and the latest release).

rjzamora (approved):
Thanks @jrbourbeau

I'll merge this now, and we can change our stance if relying on `main` causes problems.

PR #157, opened 2023-06-15
Test against dev `dask` and `distributed`

jrbourbeau:
I just happened to catch https://github.com/dask-contrib/dask-expr/pull/156 by chance. This PR takes a lightweight approach to covering dev versions of `dask` and `distributed` (just a single CI build).

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -38,6 +38,12 @@ jobs:
           extra-specs: |
             python=${{ matrix.python-version }}
 
+      - name: Upstream packages
+        if: ${{ matrix.python-version == '3.11' }}
+        run: |
+          python -m pip install git+https://github.com/dask/distributed
+          python -m pip install git+https://github.com/dask/dask
+
       - name: Install Dask-Match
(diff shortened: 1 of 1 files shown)


jrbourbeau:
Note https://github.com/dask-contrib/dask-expr/pull/156 fixes the failure here

.github/workflows/test.yaml
```diff
@@ -38,6 +38,12 @@ jobs:
           extra-specs: |
             python=${{ matrix.python-version }}
 
+      - name: Upstream packages
+        if: ${{ matrix.python-version == '3.11' }}
+        run: |
+          python -m pip install git+https://github.com/dask/distributed
+          python -m pip install git+https://github.com/dask/dask
```
rjzamora, line 45: #156 was recently merged, and it now installs dask dev from source in `environment.yml`. Should we also add distributed to that file?

jrbourbeau: Yeah, I think so. Let's close this PR in favor of https://github.com/dask-contrib/dask-expr/pull/190

PR #158, opened 2023-06-16
Refactor ``Expr`` class

### LINKED ISSUE #142: Remove DataFrame assumptions from Expr
Eventually we'll have all sorts of Exprs like arrays, bags, etc..  Today we're mostly focused on dataframe and that's good I think.

However, even today we have non-DataFrame exprs like scalars, and the results of `to_parquet` calls.  Maybe it's time to pull out some of the dataframe metadata assumptions like `divisions` and maybe even `meta` from the `Expr` class and make some new `Frame` class (or some better name) from which dataframe-like-things (dataframe, series, index) inherit.

phofl:
This is a first suggestion. There is still stuff that needs to be done as follow up (e.g. adapting new_collection for example), but this will be a pain to synch, so want to keep scope as small as possible

- [x] closes #142


dask_expr/expr.py
```diff
@@ -33,8 +19,7 @@
 class Expr:
     """Primary class for all Expressions
 
-    This mostly includes Dask protocols and various Pandas-like method
-    definitions to make us look more like a DataFrame.
+    This mostly includes Dask protocols.
     """
 
     commutative = False
```
mrocklin, line 25: Unrelated to this PR, but we can remove commutativve / associative

mrocklin: These are left over from the matchpy days

dask_expr/expr.py
```diff
@@ -206,8 +132,35 @@ def _layer(self) -> dict:
...
+        # (e.g. if the key is reserved by a method/property)
+        return self.operands[type(self)._parameters.index(key)]
+
+    def dependencies(self):
+        # Dependencies are `Expr` operands only
+        return [operand for operand in self.operands if isinstance(operand, Expr)]
+
+    @property
+    def npartitions(self):
+        raise NotImplementedError
```
mrocklin, line 163: Maybe npartitions should move over?  Not a huge deal I guess if collections like scalar or delayed return 1 though.

dask_expr/expr.py
```diff
@@ -674,1087 +433,3 @@ def find_operations(self, operation: type) -> Generator[Expr]:
...
-    def _divisions(self):
-        return (None, None)
-
-    @property
-    def _meta(self):
-        return make_meta(self.value)
-
-    def _task(self, index: int):
-        assert index == 0
-        return self.value
```
mrocklin, line 693: Maybe Literal stays?  I wonder if we can remove divisions/meta and still have things work.

dask_expr/frameexpr.py
```diff
@@ -0,0 +1,1375 @@
+from __future__ import annotations
```
mrocklin, line 1: I think I'm not a fan of the name `frameexpr.py`.  Maybe `frame.py`, maybe `dataframe/expr.py`?  Thoughts?

dask_expr/frameexpr.py
```diff
@@ -0,0 +1,1375 @@
...
+
+class FrameExpr(Expr):
+    """Primary class for all Expressions
+
+    This mostly includes pandas-like method
+    definitions to make us look more like a DataFrame.
+    """
+
+    def __hash__(self):
+        return hash(self._name)
```
mrocklin, line 40: I'm curious about your thoughts on where `hash` should live.  Is this because `_name` may not be on all expressions? (maybe it should?)

dask_expr/expr.py
```diff
@@ -120,55 +97,13 @@ def __hash__(self):
...
-        try:
-            return object.__getattribute__(self, key)
-        except AttributeError as err:
-            # Allow operands to be accessed as attributes
-            # as long as the keys are not already reserved
-            # by existing methods/properties
-            _parameters = type(self)._parameters
-            if key in _parameters:
-                idx = _parameters.index(key)
-                return self.operands[idx]
```
mrocklin, line 145: Probably we want this top bit in `Expr`

dask_expr/frameexpr.py
```diff
@@ -0,0 +1,1375 @@
...
+        return {(self._name, i): self._task(i) for i in range(self.npartitions)}
+
+    def _simplify_down(self):
+        return
+
+    def _simplify_up(self, parent):
+        return
+
+    def optimize(self, **kwargs):
+        return optimize(self, **kwargs)
```
mrocklin, line 97: These should maybe be in `Expr`

dask_expr/frameexpr.py
```diff
@@ -0,0 +1,1375 @@
...
+    def _task(self, index: int):
+        return self._filtered_task(self._partitions[index])
+
+    def _filtered_task(self, index: int):
+        raise NotImplementedError()
+
+
+@normalize_token.register(FrameExpr)
+def normalize_expression(expr):
+    return expr._name
```
mrocklin, line 1126: This should probably be in expr

dask_expr/frameexpr.py
mrocklin: Oh, I see, optimize is dataframe specific. The simplify methods then at least should probably be moved up.

dask_expr/frameexpr.py
```diff
@@ -0,0 +1,1375 @@
...
+from dask.utils import M, apply, funcname
+
+from dask_expr.expr import Expr
+
+replacement_rules = []
+
+no_default = "__no_default__"
+
+
+class FrameExpr(Expr):
```
mrocklin, line 32: Maybe?

```python
class Frame(Expr)
```

Or is that weird for scalars?  (maybe scalars shouldn't be Frames?)

In general I dislike.  

```python
class FooBar(Bar)
```

It always feels needlessly wordy to me.  This is subjective though and I'm happy to be overruled.

mrocklin (commented):
Some comments

mrocklin:
I'm curious if we can make Scalar inherit from Expr rather than FrameExpr.  This will likely stress things like `Add` and `simplify` and `meta` checks.  I think that that'll open up a lot of interesting questions though.  

Maybe that's follow-up work though if we want to keep things low-friction.

dask_expr/expr.py
phofl: I checked this with the dask/dask implementation and for stuff like bag this is given through __init__, that's why I ordered it like this

dask_expr/expr.py
phofl: I can take a look as a follow-up if ok?

dask_expr/frameexpr.py
phofl: This is very weird, we have a lot of tests failing if I add this only to Expr, but it was hard to investigate with all the other changes as well, that's why I put it here as well for now.

mrocklin: You might need to move over `__eq__` as well.  Python has weird rules around hash and eq together.

dask_expr/frameexpr.py
phofl: The simplify steps live in both actually, missed deleting them here

dask_expr/expr.py
phofl: Oh good point, moved up

mrocklin:
OK, so let's imagine that we make it so that scalars don't inherit from FrameExpr, but just Expr.  And then we want to add a scalar to a series:

```python
scalar + series
```

First, we check `Scalar.__add__`, which probably uses `expr.Add` normally, but instead notices that the `other` object has its own `Add` class and uses that.  I guess we have lots of diffeent Adds?

```python
class Add(Expr):
    ...

class Add(FrameExpr, expr.Add):
    ...
```

This is starting to get a litlte weird.  Maybe it's ok though.  

Part of me thinks "we should just use one `Add` class, expressions are designed to show what the user typed in, and not capture user object types (we get into terrible many-class multiple-inheritance as above if we try to capture types as well as operations).  But then we have to figure out lots of things dynamically like "which optimization functions should I use in this expression?" and "should this class have frame methods like `map_partitions` and maybe more.   Historically when we've run into these situations we've said "ok, expresions should not have dataframe/series/type sensitive methods.  Those should live on the collection".  If we now make expressions generic to the point where they may or may not be dataframes then we probably lose all of those shared methods.

I bring this up because it's possible that while this path seems generally pretty good now, that it leads us down some odd multi-class hell.  This is some small motivation to not merge this yet until we've looked a bit farther down this path.  (not pushing for that hard though).

@rjzamora when you wake up you might want to take a look at this PR.

dask_expr/frameexpr.py
phofl: Renamed

dask_expr/frameexpr.py
phofl: Moved

PR #164, opened 2023-06-20
Add ``storage_options`` to ``read_csv``

phofl:
Not sure whether adding a test here makes sense.


mrocklin:
I'll defer review here to @rjzamora I think if that's ok

rjzamora (approved):
Seems fine to me. Thanks

PR #165, opened 2023-06-20
Add ``divisions`` to Reduction

phofl:
This is in the way of some alignment stuff I was looking into.


dask_expr/groupby.py
```diff
@@ -294,6 +294,9 @@ def chunk_kwargs(self):
     def combine_kwargs(self):
         return {"levels": self.levels}
 
+    def _divisions(self):
+        return (None, None)
```
rjzamora, line 298: Just a small note that this will eventually need to depend on the value of `split_out`.

PR #166, opened 2023-06-20
Implement alignment for BinOps

phofl:
Sits on top of #165

### DIFF
--- a/dask_expr/collection.py
+++ b/dask_expr/collection.py
@@ -5,7 +5,7 @@ import warnings
 from numbers import Number
 
 import numpy as np
-from dask.base import DaskMethodsMixin, named_schedulers
+from dask.base import DaskMethodsMixin, is_dask_collection, named_schedulers
 from dask.dataframe.core import (
     _concat,
     _Frame,
@@ -20,6 +20,7 @@ from fsspec.utils import stringify_path
 from tlz import first
 
 from dask_expr import expr
+from dask_expr.concat import Concat
 from dask_expr.expr import no_default
 from dask_expr.merge import Merge
 from dask_expr.reductions import (
@@ -524,6 +525,32 @@ class DataFrame(FrameBase):
             )
         )
 
+    def join(
+        self,
+        other,
+        on=None,
+        how="left",
+        lsuffix="",
+        rsuffix="",
+        shuffle_backend=None,
+    ):
+        if not is_dataframe_like(other._meta) and hasattr(other._meta, "name"):
+            other = new_collection(expr.ToFrame(other.expr))
+
+        if not is_dataframe_like(other._meta):
+            # TODO: Implement multi-join
+            raise NotImplementedError
+
+        return self.merge(
+            right=other,
+            left_index=on is None,
+            right_index=True,
+            left_on=on,
+            how=how,
+            suffixes=(lsuffix, rsuffix),
+            shuffle_backend=shuffle_backend,
+        )
+
     def __setitem__(self, key, value):
         out = self.assign(**{key: value})
         self._expr = out._expr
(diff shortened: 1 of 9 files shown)


dask_expr/repartition.py
```diff
@@ -313,3 +313,45 @@ def _is_single_last_div(x):
...
+            "Not all divisions are known, can't align "
+            "partitions. Please use `set_index` "
+            "to set the index."
+        )
+
+    divisions = list(unique(merge_sorted(*[df.divisions for df in dfs])))
+    if len(divisions) == 1:  # single value for index
+        divisions = (divisions[0], divisions[0])
+
+    return [Repartition(df, new_divisions=divisions, force=True) for df in dfs]
```
mrocklin, line 357: This is interesting.  We're taking possibly-lazily-computed-metadata and putting into the expression tree.  This kind of locks us in a little.  

For example, if we, after this, perform a filter and then pass that Filter down through this expression into a ReadParquet then we'll want to change the divisions.  This `Repartition` call will then become inaccurate.

I think that we're discovering a principle that inputs to Expressions should only reflect inputs provided by the user, never something that we compute based on the current form of the expressions.

mrocklin: Maybe this makes sense to do if we're deep in the expression process, turning very abstract expressions into more concrete ones.  I don't have confidence yet that we don't want exactly what's here, I mostly want us to agree on the "expression operands should reflect user inputs" principle.

phofl: I can see your point, I was struggling with something related when starting to look into ``set_index``.

I am just not sure what alternatives we have for operations that require aligning the data. Disallowing them in general could be an option, but that wouldn't be a good solution from a usability perspective. Requiring explicit divisions could be an option for some operations, but this is also not great usability-wise.

> Maybe this makes sense to do if we're deep in the expression process, turning very abstract expressions into more concrete ones. 

I think this touches an interesting point. We could update the ``Repartition`` call when we push something like a filter down. This would keep the divisions updated with the lazily computed ones. Not totally sure whether this would cause problems in other areas though.

mrocklin: > We could update the Repartition call when we push something like a filter down.  

That seems hard to track well. 

Instead, my thinking is that we try hard never to create `Repartition(divisions=)` expressions, except after most optimization is done.  I expect these expressions to be the results of operations like the following:

```python
x + y
df.repartition("1 GiB")
```

We should keep these expresssions as close as possible to what the user wrote down:

```python
Add(x, y)
RepartitionSize(df, "1 GiB")
```

It should be easy to optimize around these.  We should try hard not to lower these too early into things like the following:

```python
Add(Repartition(x, divisions=...), Repartition(y, divisions=...))
RepartitionDivisions(df, [...])
```

Once we do that then the expression is kind of locked in and we can't optimize through these terms.  Things are concrete rather than fluid.  

So I think that the `Repartition(divisions=...)` expression is OK, but we want to avoid engaging it as much as possible.  I would be sad if we were lowering to that immediately in user API, for example.  We probably also need to start thinking about when different optimizations happen.  `Repartition(size=...) -> Repartition(divisions=...)` wants to be polite and let other optimizations go first.  Once all of them are done, then it'll go ahead and execute.  We don't currently have any mechanism to control the order of optimizations unfortunately.  

@rjzamora thought about something this a little bit similar (Rick, I'm thinking of simplify vs lower, which I think is a special case of this problem).  Pinging him.

dask_expr/collection.py
```diff
@@ -58,7 +58,14 @@ def _wrap_expr_op(self, other, op=None):
     assert op is not None
     if isinstance(other, FrameBase):
         other = other.expr
-    return new_collection(getattr(self.expr, op)(other))
+    left = self.expr
+
+    if not isinstance(other, expr.Expr):
+        pass
+    else:
+        left, other = _maybe_align_partitions([left, other])
```
mrocklin, line 66: Given my recent comment, I think that we maybe want some abstract `AlignDivisions` expression here or something.  We don't specify exactly what the divisions should be (because we genuinely don't know them yet and don't want to claim that we do) but we know that, when the time comes, we'll want to align the inputs along those divisions.

Maybe we keep the check if all of the divisions are the same though. This isn't quite correct (maybe one side's divisions could shift from the other's) but I wouldn't want to inject a ton of AlignDivisions expressions everywhere.  Thoughts?

phofl: This ``AlignDivisions`` expression would set its ``_divisions`` to ``(None, None)`` then? It would only calculate its divisions when we construct the actual graph after all optimisations are through? Just making sure that I understand you correctly.

Couple of things come to mind:

- We potentially end up injecting this expression in a lot of places where it might not be necessary? e.g.:

```
x = a - b. # injects abstract ``AlignDivisions`` because they don't have the same structure 
y = x + c  # ``divisions`` of ``x``  are unknown, so inject another one?
```

- I am assuming that we would use the same pattern for ``align`` and friends where the alignment was specifically requested?
- I am wondering whether or not we shoot ourselves in the foot because follow up operations don't know the divisions of the previous object?

mrocklin: I think that broadly we should try to use divisions as little as possible when building and operating on expressions.  Divisions are lazy for a reason.  They should mostly be used when constructing the graph.

phofl:
See #193

PR #168, opened 2023-06-20
Basic p2p shuffle support

rjzamora:


dask_expr/tests/test_distributed.py
rjzamora, line 1: @jrbourbeau - This will be the first instance of distributed testing within dask-expr. Please feel free to advise

phofl (commented):
looks good to me generally, but someone who is more familiar with p2p shuffling should take a look as well

dask_expr/shuffle.py
```diff
@@ -462,6 +461,55 @@ def _layer(self):
...
+        for i in range(self.frame.npartitions):
+            transfer_keys.append((name, i))
+            dsk[(name, i)] = (
+                shuffle_transfer,
+                (self.frame._name, i),
+                token,
+                i,
+                self.npartitions_out,
+                self.partitioning_index,
+                parts_out,
```
hendrikmakait, line 495: `shuffle_transfer` currently expects `parts_out` to be a `set`. If that matters, let's convert to `set` here. If it doesn't matter, let's relax the typing in `dask/distributed`.

dask_expr/shuffle.py
```diff
@@ -462,6 +461,55 @@ def _layer(self):
...
+            )
+
+        dsk[_barrier_key] = (shuffle_barrier, token, transfer_keys)
+
+        # TODO: Decompose p2p Into transfer/barrier + unpack
+        name = self._name
+        for i, part_out in enumerate(parts_out):
```
hendrikmakait, line 502: Is there a specific reason for not reusing the partition name/ID in the task name? I feel like this will add one additional indirection to keep in mind when debugging in the future.

hendrikmakait (commented):
Overall, this looks good to me, with two open questions.

dask_expr/shuffle.py
rjzamora: Thanks for pointing this out. We can probably use a `set` if that's what `shuffle_transfer` expects.

dask_expr/shuffle.py
rjzamora: I'm not certain that I fully understand what you are saying here, but here is what I *think* you are looking for:

**Regarding the name**: We need the name of these final `shuffle_unpack` tasks to match the name of the expression (i.e. `self._name`).

**Regarding the index**: Since `P2PShuffle` is technically a `PartitionsFiltered` instance, we also need to make sure that the output indices are always a contiguous range starting with `0` (e.g. [0, 1, 2, 3]). That is, if `self._partitions` is `[1, 3, 4]`, it means we must remap what would **usually** be partitions `[1, 3, 4]` to have the indices `[0, 1, 2]`. This `PartitionsFiltered` mechanism is absolutely an unfortunate level of indirection. However, it essentially allows us to avoid low-level culling in dask-expr.

hendrikmakait: Sorry, I referred to the use of the `i` as the index in the task key (not name). That explanation clears it up, I wasn't aware that a contiguous range is expected. Thanks!

hendrikmakait (approved):
LGTM looking at it from the shuffling side

dask_expr/tests/test_distributed.py
rjzamora, line 1: @jrbourbeau - Does this seem like a reasonable starting point for distributed tests?

dask_expr/tests/test_distributed.py
```diff
@@ -0,0 +1,70 @@
...
+async def clean_scheduler(
+    scheduler: Scheduler, interval: float = 0.01, timeout: int | None = None
+) -> None:
+    """Assert that the scheduler has no shuffle state"""
+    deadline = Deadline.after(timeout)
+    extension = scheduler.extensions["shuffle"]
+    while extension.states and not deadline.expired:
+        await asyncio.sleep(interval)
+    assert not extension.states
+    assert not extension.heartbeats
```
mrocklin, line 43: I strongly recommend using `distributed.utils_tests.gen_cluster`

mrocklin: Oh wait, I see that we are.  Never mind, I was confused.  I realize now that this is for shuffling code.  Maybe not a big deal here?  This is already tested in dask/dask?

dask_expr/tests/test_distributed.py
mrocklin: I think that as long as we're using `gen_cluster` everything is fine.

dask_expr/tests/test_distributed.py
rjzamora: Right, I copied this from the shuffle tests in distributed and thought we may also want to make sure we are not leaving behind any shuffle state when our test finishes. Do you think we can just rely on `gen_cluster` and skip the `clean_worker` and `clean_scheduler` calls?

mrocklin: They do very different things.  gen_cluster sets up a Dask cluster.  clean_worker/scheduler verify that no shuffling state leaked out for p2p shuffling.  I think that the p2p shuffling is written and tested in dask/distributed.  I don't think that we need to also test that it functions well in this code base.  Instead, I think we need to verify that it runs, and that things like optimizations work well.

rjzamora: I think I'll assume "yes" for now, so we can keep `test_distributed.py` light.

rjzamora: I agree that the leakage is already tested for in dask/distributed and doesn't need to be tested here. I'm pretty sure I just misunderstood exactly what those utilities were doing.

I think this PR is now in good shape for initial p2p support.

PR #170, opened 2023-06-20
Change incorrect Projection test

phofl:
The last commit broke ci because of other changes on main since then. 

The test that I changed was incorrect, doing ``df["x"]`` will reduce to a Series, selecting ``x`` again would reduce to a scalar, so we can't just get rid of the Projection.


rjzamora (approved):
Make sense!

phofl:
Merging to get ci back to green

PR #171, opened 2023-06-21
Use list to represent columns on Expr

phofl:
cc @rjzamora

### DIFF
--- a/dask_expr/collection.py
+++ b/dask_expr/collection.py
@@ -96,6 +96,10 @@ class FrameBase(DaskMethodsMixin):
     def size(self):
         return new_collection(self.expr.size)
 
+    @property
+    def columns(self):
+        return pd.Index(self.expr.columns)
+
     def __len__(self):
         return new_collection(Len(self.expr)).compute()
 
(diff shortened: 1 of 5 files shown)


dask_expr/expr.py
```diff
@@ -491,9 +491,9 @@ def _name(self):
     @property
     def columns(self):
```
rjzamora, line 492:
```suggestion
    def columns(self) -> list:
```
Maybe we start being explicit about the expected type here? (Certainly not a blocker)

dask_expr/collection.py
```diff
@@ -96,6 +96,10 @@ def _meta(self):
     def size(self):
         return new_collection(self.expr.size)
 
+    @property
+    def columns(self):
```
rjzamora, line 100: My preference is to stick with exactly what you have here, but just a thought: Do we want to raise an error for `Series.columns` (to be consistent with pandas)?

rjzamora (approved):
Nice. Thanks @phofl!

dask_expr/expr.py
phofl: Sounds good to me

dask_expr/collection.py
rjzamora: >We can add this here, this wouldn't break any tests

Up to you. This PR is already a win "as is" in my mind.

phofl: Then let's merge as is, the columns change isn't urgent

PR #172, opened 2023-06-21
Add ``_simplify_up`` in ``Blockwise``

phofl:
I think whitelisting them here is the best approach

### DIFF
--- a/dask_expr/expr.py
+++ b/dask_expr/expr.py
@@ -810,6 +810,26 @@ class Blockwise(Expr):
         else:
             return (self.operation,) + tuple(args)
 
+    def _simplify_up(self, parent):
+        if isinstance(
+            self,
+            (
+                Replace,
+                Isin,
+                Clip,
+                ToTimestamp,
+                IsNa,
+                Round,
+                Abs,
+                Apply,
+                Map,
+            ),
+        ) and isinstance(parent, Projection):
+            if list(self.columns) == list(parent.columns):
+                return self
+
+            return type(self)(self.frame[parent.operand("columns")], *self.operands[1:])
+
 
 class MapPartitions(Blockwise):
     _parameters = [
--- a/dask_expr/tests/test_collection.py
+++ b/dask_expr/tests/test_collection.py
@@ -728,6 +728,29 @@ def test_dir(df):
     assert "sum" in dir(df.index)
 
 
+@pytest.mark.parametrize(
+    "func, args",
+    [
+        ("replace", (1, 2)),
+        ("isin", ([1, 2],)),
+        ("clip", (0, 5)),
+        ("isna", ()),
+        ("round", ()),
+        ("abs", ()),
+        ("apply", (lambda x: x + 1,)),
+        # ("map", (lambda x: x+1, )),  # add in when pandas 2.1 is out
+    ],
+)
+@pytest.mark.parametrize("indexer", ["x", ["x"]])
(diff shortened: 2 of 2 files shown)


dask_expr/expr.py
```diff
@@ -810,6 +810,26 @@ def _task(self, index: int):
...
+                Round,
+                Abs,
+                Apply,
+                Map,
+            ),
+        ) and isinstance(parent, Projection):
+            if list(self.columns) == list(parent.columns):
+                return self
+
+            return type(self)(self.frame[parent.operand("columns")], *self.operands[1:])
```
mrocklin, line 831: Dusting off my "OO Programmer" hat (a rarely used hat) I'd say that this breaks the idea of inheritance, where the superclass shouldn't have to think about the subclasses.

Instead, I'll suggest that we add a class attribute like "projection_passthrough=False` to `Blockwise` and then set it to True for all of these classes.  Then future classes can add that attribute and opt-in to this optimization.

phofl: Good point, added the variable

dask_expr/expr.py
```diff
@@ -1029,23 +1041,27 @@ class AsType(Elemwise):
...
 class Abs(Elemwise):
+    projection_passthrough = True
     _parameters = ["frame"]
     operation = M.abs
 
 
 class Apply(Elemwise):
     """A good example of writing a less-trivial blockwise operation"""
 
+    projection_passthrough = True
```
mrocklin, line 1064: This one seems suspect to me

dask_expr/expr.py
```diff
@@ -1068,6 +1084,7 @@ def _task(self, index: int):
 
 
 class Map(Elemwise):
+    projection_passthrough = True
```
mrocklin, line 1087: This one seems suspect to me

dask_expr/expr.py
phofl: Why? Both of them are column-wise as long as we don't support an axis keyword. So we can drop columns that aren't needed. 

Or is your concern with regards to side-effects of potential udfs?

dask_expr/expr.py
mrocklin: I just checked and saw that there is no `DataFrame.map` so I'll retract this comment.

phofl: map is applied to every value, there isn't much you can do with regards to structure

dask_expr/expr.py
```diff
@@ -810,6 +811,13 @@ def _task(self, index: int):
         else:
             return (self.operation,) + tuple(args)
 
+    def _simplify_up(self, parent):
+        if self.projection_passthrough and isinstance(parent, Projection):
+            if list(self.columns) == list(parent.columns):
+                return self
```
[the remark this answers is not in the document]
phofl: removed

dask_expr/expr.py
phofl: I have to correct myself: You can do whatever you want if your udf is given as a string (drop for example). I don't know why you would use it like that, but this means that we can't push projections through.

dask_expr/expr.py
```diff
@@ -731,6 +731,7 @@ class Blockwise(Expr):
 
     operation = None
     _keyword_only = []
+    projection_passthrough = False
```
mrocklin, line 734: Final comment (I promise): Should we make this private?

phofl: Fine by me. Makes probably more sense here

PR #173, opened 2023-06-21
Add ``_simplify_up`` for ``dropna``

phofl:


dask_expr/tests/test_collection.py
```diff
@@ -721,6 +721,16 @@ def test_find_operations(df):
     assert next(iter(adds))._name == df2._name
 
 
+@pytest.mark.parametrize("subset", ["x", ["x"]])
+def test_dropna_simplify(pdf, subset):
+    pdf["z"] = 1
+    df = from_pandas(pdf)
+    q = df.dropna(subset=subset)["y"]
+    result = optimize(q, fuse=False)
```
rjzamora, line 729: **Optional suggestion**: `result = q.simplify()`

Might also be good to add `assert_eq(result, pdf.dropna(subset=subset)["y"])`

rjzamora (approved):
Minor suggestion, but looks good as is.

dask_expr/tests/test_collection.py
phofl: Done, I added a utility in the other PR. Will rebase after one of them is merged

PR #175, opened 2023-06-21
Add ``simplify_up`` for ``astype``

phofl:


dask_expr/expr.py
```diff
@@ -1027,6 +1027,17 @@ class AsType(Elemwise):
...
 
+    def _simplify_up(self, parent):
+        if isinstance(parent, Projection):
+            dtypes = self.operand("dtypes")
+            if isinstance(dtypes, dict):
+                dtypes = {
+                    key: val for key, val in dtypes.items() if key in parent.columns
+                }
+                if not dtypes:
+                    return type(parent)(self.frame, *parent.operands[1:])
```
mrocklin, line 1038: Maybe this moves into `simplify_down` so that we get the same result, but in two steps, and a bit more generally.

Probably this doesn't matter because I'm not sure how else we would ever get an empty `dtypes` input, but in general I think it's good to identify when we're doing a couple of different optimizations at once, and try to pull them apart when we can.  

This comment is probably more pedantic than useful though.  Feel free to ignore.

dask_expr/tests/test_collection.py
```diff
@@ -681,6 +681,24 @@ def test_len(df, pdf):
...
+    assert result._name == expected._name
+    assert_eq(q, pdf.astype({"x": "float64", "y": "float64"})["x"])
+
+    q = df.astype({"y": "float64"})["x"]
+    result = optimize(q, fuse=False)
+    expected = df["x"]
+    assert result._name == expected._name
+
+    q = df.astype("float64")["x"]
+    result = optimize(q, fuse=False)
```
mrocklin, line 697: If you ever want, this is the same as `q.simplify()`

Today `optimize` calls `simplify` then `fuse`.  Sometimes just calling simplify can be simpler than `optimize(..., fuse=False)`

mrocklin (approved):

dask_expr/expr.py
phofl: I thought about this as well, but I don't want to remove empty dicts in general since passing an empty dict should raise.

dask_expr/tests/test_collection.py
phofl: I wasn't sure whether this might change at some point, so I decided to go with optimize hoping that this might be more stable (not sure if this is ever relevant though)

mrocklin: I could also imagine the `fuse=` keyword going away.  In general I don't think that much is stable :)

phofl: :)

PR #176, opened 2023-06-21
Implement ``simplify_up`` for ``explode``

phofl:


dask_expr/expr.py
```diff
@@ -1092,6 +1092,18 @@ class ExplodeSeries(Blockwise):
...
+        if isinstance(parent, Projection):
+            columns = set(parent.columns).union(self.column)
+            if columns == set(self.frame.columns):
+                # Don't add unnecessary Projections, protects against loops
+                return
+
+            return type(parent)(
+                type(self)(self.frame[sorted(columns)], *self.operands[1:]),
+                *parent.operands[1:],
+            )
```
mrocklin, line 1105: Parent is a Projection here right?  Maybe `type(parent)` isn't necessary, but we can just use `[...]` (implicitly, I'm not expecting any relevant subclasses of Projection).

Also, does explode add columns?  If not, then maybe we don't need the outer projection at all?

phofl: We have the type thing in a couple of places, I can clean them up all at once when the simplify prs are in.

No it does not, but you might have something like the following:

```
df.explode(column="a")["b"]
```

You have to keep "a" and "b" before applying explode, that's why we need both projections unfortunately.

dask_expr/collection.py
```diff
@@ -671,6 +671,12 @@ def rename(self, columns):
         return new_collection(expr.RenameFrame(self.expr, columns=columns))
 
     def explode(self, column):
+        if (
+            column is not None
+            and not isinstance(column, list)
+            and not hasattr(column, "dtype")
+        ):
+            column = [column]
```
rjzamora, line 679: **Just an observation**: This pretty much matches the `subset` normalization used for #173. Will this logic be repeated in more places? If so, maybe we should add/use a simple utility to always convert to `list`? Could information ever be lost by converting `pd.Index` -> `list`?

dask_expr/tests/test_collection.py
```diff
@@ -53,6 +53,15 @@ def test_explode():
     assert_eq(pdf.a.explode(), df.a.explode())
 
 
+def test_explode_simplify(pdf):
```
rjzamora, line 56: Same (optional) suggestions as https://github.com/dask-contrib/dask-expr/pull/173/files#r1240350384 I suppose.

rjzamora (approved):

dask_expr/collection.py
phofl: Yes this is the same, adding a utility is on my todo list, can also do it in this PR if you prefer.

Yeah, but nothing that's relevant here fortunately.

dask_expr/tests/test_collection.py
phofl: Updated

dask_expr/collection.py
phofl: Updated to create a utility function

dask_expr/_utils.py
rjzamora, line 1: I'm not sure how I feel about prepending an underscore to the file name like this (it seems to be a convention in pandas, but not really in dask).

dask_expr/_utils.py
```diff
@@ -0,0 +1,8 @@
+def _convert_to_list(column) -> list:
```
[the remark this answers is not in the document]
rjzamora: Just a note that I'm fine with merging the `_simplify_up` changes first, and visiting the utility question separately.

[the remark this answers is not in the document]
phofl: reverted

dask_expr/_utils.py
phofl: Yes that makes sense! But I think it's easier to go the private -> non-private route than the other way around (We (pandas team) are thinking about making pandas.core private for quite some time now, but it's such a hassle...)

rjzamora: >it's easier to go the private -> non-private route

Yeah, good point.

dask_expr/_utils.py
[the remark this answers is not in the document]
phofl: Thx

PR #177, opened 2023-06-21
Implement ``select_dtypes``

phofl:


dask_expr/collection.py
```diff
@@ -678,6 +678,10 @@ def to_parquet(self, path, **kwargs):
 
         return to_parquet(self, path, **kwargs)
 
+    def select_dtypes(self, include=None, exclude=None):
+        columns = self._meta.select_dtypes(include=include, exclude=exclude).columns
+        return new_collection(expr.Projection(self.expr, columns=columns))
```
rjzamora, line 683: **(Slight) Style Preference**
```suggestion
        return new_collection(self.expr[columns])
```

rjzamora (approved):

PR #179, opened 2023-06-22
Implement ``_simplify_up`` for ``concat``

phofl:
This handles the concat cases we already have. Might have to adjust later

### DIFF
--- a/dask_expr/concat.py
+++ b/dask_expr/concat.py
@@ -6,7 +6,7 @@ from dask.dataframe.dispatch import make_meta, meta_nonempty
 from dask.dataframe.utils import check_meta, strip_unknown_categories
 from dask.utils import apply, is_dataframe_like, is_series_like
 
-from dask_expr.expr import AsType, Expr
+from dask_expr.expr import AsType, Expr, Projection
 
 
 class Concat(Expr):
@@ -84,8 +84,23 @@ class Concat(Expr):
         )
 
     def _simplify_up(self, parent):
-        # TODO: implement passing projections through
-        return
+        if isinstance(parent, Projection):
+            columns = parent.columns
+            columns_frame = [
+                sorted(set(frame.columns).intersection(columns))
+                for frame in self._frames
+            ]
+            if all(
+                cols == frame.columns
+                for frame, cols in zip(self._frames, columns_frame)
+            ):
+                return
+
+            frames = [frame[cols] for frame, cols in zip(self._frames, columns_frame)]
+            return type(parent)(
+                type(self)(self.join, self.ignore_order, self._kwargs, *frames),
+                *parent.operands[1:],
+            )
 
 
 class StackPartition(Concat):
@@ -115,7 +130,4 @@ class StackPartition(Concat):
         return dsk
 
     def _simplify_down(self):
-        return
-
-    def _simplify_up(self, parent):
         return
--- a/dask_expr/concat.py
+++ b/dask_expr/concat.py
@@ -97,9 +97,8 @@ class Concat(Expr):
                 return
 
             frames = [
-                frame[cols]
+                frame[cols] if cols != sorted(frame.columns) else frame
                 for frame, cols in zip(self._frames, columns_frame)
-                if cols != sorted(frame.columns)
             ]
             return type(parent)(
                 type(self)(self.join, self.ignore_order, self._kwargs, *frames),
--- a/dask_expr/concat.py
+++ b/dask_expr/concat.py
@@ -91,12 +91,16 @@ class Concat(Expr):
                 for frame in self._frames
             ]
             if all(
-                cols == frame.columns
+                cols == sorted(frame.columns)
                 for frame, cols in zip(self._frames, columns_frame)
             ):
                 return
 
-            frames = [frame[cols] for frame, cols in zip(self._frames, columns_frame)]
+            frames = [
+                frame[cols]
+                for frame, cols in zip(self._frames, columns_frame)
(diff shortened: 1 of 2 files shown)


dask_expr/concat.py
```diff
@@ -84,8 +84,23 @@ def _simplify_down(self):
         )
 
     def _simplify_up(self, parent):
-        # TODO: implement passing projections through
-        return
+        if isinstance(parent, Projection):
+            columns = parent.columns
+            columns_frame = [
+                sorted(set(frame.columns).intersection(columns))
```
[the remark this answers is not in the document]
rjzamora: Can you explain the reason for sorting? Is this related at all to how pandas will order the columns after `concat`?

dask_expr/concat.py
```diff
@@ -84,8 +84,23 @@ def _simplify_down(self):
...
+                for frame in self._frames
+            ]
+            if all(
+                cols == frame.columns
```
rjzamora, line 94: can we also avoid the projection pushdown if the columns are the same, but in a different order?

dask_expr/concat.py
```diff
@@ -84,8 +84,23 @@ def _simplify_down(self):
...
+                for frame, cols in zip(self._frames, columns_frame)
+            ):
+                return
+
+            frames = [frame[cols] for frame, cols in zip(self._frames, columns_frame)]
```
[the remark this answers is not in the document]
rjzamora: Note that I also have the same question about whether the order of the columns really matters.

dask_expr/tests/test_concat.py
```diff
@@ -65,3 +65,15 @@ def test_concat_one_no_columns(df, pdf):
...
     assert_eq(result, expected)
+
+
+def test_concat_simplify(pdf, df):
+    pdf2 = pdf.copy()
+    pdf2["z"] = 1
+    df2 = from_pandas(pdf2)
+    q = concat([df, df2])[["z", "x"]]
+    result = q.simplify()
+    expected = concat([df[["x"]], df2[["x", "z"]]])._simplify_down()[["z", "x"]]
```
rjzamora, line 76: Does this not work with `simplify` instead of `_simplify_down`?

rjzamora (commented):
Thanks @phofl

Just have a few questions, but the code looks mostly good to me.

dask_expr/concat.py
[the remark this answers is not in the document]
phofl: We have to compare somehow to frame.columns, which is easier when they are sorted but not strictly necessary if we compare both lists after sorting

dask_expr/concat.py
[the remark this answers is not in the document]
phofl: Yep this should be doable, good point

dask_expr/tests/test_concat.py
phofl: Simplify would remove the projection in the end to, I just want to push it down into StackPartition and keep everything else as is

dask_expr/concat.py
phofl: Yep I’ll adjust these conditions

dask_expr/concat.py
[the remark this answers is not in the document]
phofl: Testing the expression structure also gets harder if these can occur in any order

dask_expr/tests/test_concat.py
rjzamora: Okay, so you are saying `concat([df[["x"]], df2[["x", "z"]]])._simplify_down()` is **not** the same as `concat([df[["x"]], df2[["x", "z"]]]).simplify()`? Interesting.

phofl: No sorry my comment was wrong, this should yield the same result, otherwise something else is buggy.

Initially, I had it at then end, which would have produced something different, but doing it in the middle should work. Sorry for the confusion

dask_expr/concat.py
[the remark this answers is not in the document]
phofl: Added it, this keeps the cycles shorter

dask_expr/concat.py
phofl: Added as well

dask_expr/concat.py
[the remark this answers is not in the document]
phofl: I'd prefer keeping the sorted condition if there is no requirement from cudf that would break this. Using a set would make the order non-deterministic, which restricts our ability to test the expression structure. So this is just an internal thing, not pandas related.

dask_expr/tests/test_concat.py
rjzamora: Okay cool. That makes me feel better :)

dask_expr/concat.py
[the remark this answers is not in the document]
rjzamora: Okay, that seems fair to me.

dask_expr/concat.py
```diff
@@ -84,8 +84,27 @@ def _simplify_down(self):
...
+                for frame in self._frames
+            ]
+            if all(
+                cols == sorted(frame.columns)
+                for frame, cols in zip(self._frames, columns_frame)
+            ):
+                return
+
+            frames = [
+                frame[cols]
```
rjzamora, line 100: Can't this give us the wrong number of frames now?

phofl: Yes sorry, to reliant on tests. Fixed now

rjzamora (approved):

PR #180, opened 2023-06-23
Implement ``__str__`` for ``concat``

phofl:


mrocklin:
My sense is that maybe we should stop implementing `__str__` and just rely on `pprint`.  The str implementation is weird.  Thoughts?

phofl:
I kind of prefer the __str__ representation for short expression.

```
print(df.a.sum())
```

```
<dask_expr.expr.Scalar: expr=df['a'].sum()>
```

This is a bit closer to what I would expect as a user compared to pprint:

```
Sum:
  Projection: columns='a'
    FromPandas: frame='<pandas>' npartitions=5
```

mrocklin:
i agree that short is good, but our implementation of str is haphazard at
best, and sometimes wrong.  There's a quality issue here I think

phofl:
>  There's a quality issue here I think

Yep I agree with this

PR #185, opened 2023-06-23
Add ``drop``

phofl:


dask_expr/collection.py
```diff
@@ -677,6 +677,13 @@ def rename(self, columns):
     def explode(self, column):
         return new_collection(expr.ExplodeFrame(self.expr, column=column))
 
+    def drop(self, labels=None, columns=None, errors="raise"):
```
rjzamora, line 680: This logic feels slightly different than [`dask.dataframe`](https://github.com/dask/dask/blob/d00183ce91013303735eb28209c6a09b0d57a54b/dask/dataframe/core.py#L5575). Does `dask.dataframe` need updating, or should we handle `axis` to better capture the usage of `labels` and `columns`?

phofl: We haven't added the axis keyword for other methods until now, that's why I didn't do it here either. dask.dataframe is fine logic-wise

PR #187, opened 2023-06-26
Change ``optimize`` to ``simplify``

phofl:
cc @rjzamora follow-up here


rjzamora (approved):
Thanks! I do think this makes the tests slightly clearer.

PR #188, opened 2023-06-26
Implement utility to check if we can align two objects

phofl:
@mrocklin this is a minimal implementation of what we discussed on Friday. This is only a "happy" path that assumes that both objects are dask-expr collections for now.

### DIFF
--- a/dask_expr/collection.py
+++ b/dask_expr/collection.py
@@ -406,7 +406,7 @@ class FrameBase(DaskMethodsMixin):
         df = self.optimize(**optimize_kwargs) if optimize else self
         return new_dd_object(df.dask, df._name, df._meta, df.divisions)
 
-    def _can_align_without_repartition(self, other):
+    def _is_co_aligned(self, other):
         names = set()
 
         left = self.expr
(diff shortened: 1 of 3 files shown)


dask_expr/collection.py
```diff
@@ -406,6 +406,25 @@ def to_dask_dataframe(self, optimize: bool = True, **optimize_kwargs) -> _Frame:
         df = self.optimize(**optimize_kwargs) if optimize else self
         return new_dd_object(df.dask, df._name, df._meta, df.divisions)
 
+    def _is_co_aligned(self, other):
+        names = set()
+
+        left = self.expr
+        while isinstance(left, expr.Blockwise):
+            names.add(left._name)
+            left = left.frame
```
mrocklin, line 415: I think that this will break if you have an operation that has multiple inputs, like the following:

```python
df = read_parquet(...)
z = df.x + df.y
z[df.a > 0]
```

Instead, my suggestion would be to make a function like the following:

```python
def non_blockwise_ancestors(expr):
    """ Traverse through tree to find ancestors that are not blockwise """
    stack = [expr]
    while stack:
        e = stack.pop()
        if isinstance(e, Blockwise):
            stack.extend(e.dependencies())
        else:
            yield e

def are_co_aligned(*exprs):
    """ Do inputs come from different parents, modulo blockwise? """
    return len(set(ancestors)) == 1 
```

Maybe there is some nuance here, like maybe we don't care about ancestors that are broadcastable.  My guess though is that some structure like the above will be what you'll get to.  In general I think that we're all going to get better at writing graph/tree traversals 🙂

phofl:
Excluding broadcastable expressions as well now

PR #189, opened 2023-06-26
Add utility to convert to list

phofl:
This is strict now

cc @rjzamora

### DIFF
--- a/dask_expr/_util.py
+++ b/dask_expr/_util.py
@@ -1,3 +1,6 @@
+from __future__ import annotations
+
+
 def _convert_to_list(column) -> list | None:
     if column is None or isinstance(column, list):
         pass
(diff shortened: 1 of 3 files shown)


dask_expr/_util.py
```diff
@@ -0,0 +1,10 @@
+def _convert_to_list(column):
```
rjzamora, line 1:
```suggestion
def _convert_to_list(column) -> list | None:
```

Also, do we want a name  (`_columns_as_list`) and description (?) clarifying that the purpose of this utility is to normalize a column/columns  argument to `list`?  Or, do we want to keep it more general? I'm on the fence about this.

phofl: I generally prefer the ``_convert_to_foo`` pattern. We don't really have column-specific logic in there.

rjzamora (approved):

PR #190, opened 2023-06-26
Test against dev ``distributed`` too

jrbourbeau:
xref https://github.com/dask-contrib/dask-expr/pull/157#discussion_r1240422572

cc @rjzamora


rjzamora (approved):
Thanks!

PR #192, opened 2023-06-27
Add lower optimization layer

phofl:
I think I am leaning more and more to do this in a more general way. But this should be a good starting point. A simple new layer ``lower`` that can be used for repartitioning for example (I have a PR that builds on this soonish)

### DIFF
--- a/dask_expr/collection.py
+++ b/dask_expr/collection.py
@@ -122,32 +122,35 @@ class FrameBase(DaskMethodsMixin):
 
     def __dask_graph__(self):
         out = self.expr
-        out = out.simplify().lower()
+        out = out.optimize(fuse=False)
         return out.__dask_graph__()
 
     def __dask_keys__(self):
         out = self.expr
-        out = out.simplify().lower()
+        out = out.optimize(fuse=False)
         return out.__dask_keys__()
 
     def simplify(self):
         return new_collection(self.expr.simplify())
 
-    def lower(self):
-        return new_collection(self.expr.lower())
+    def lower_once(self):
+        return new_collection(self.expr.lower_once())
+
+    def optimize(self, fuse: bool = True):
+        return new_collection(self.expr.optimize(fuse=fuse))
 
     @property
     def dask(self):
         return self.__dask_graph__()
 
     def __dask_postcompute__(self):
-        state = self.simplify().lower()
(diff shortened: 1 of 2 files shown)


phofl:
I've incorporated @rjzamora's comment https://github.com/dask-contrib/dask-expr/pull/195#issuecomment-1615114189

This would function as a short term solution

dask_expr/collection.py
```diff
@@ -122,29 +122,35 @@ def __getitem__(self, other):
...
     def __dask_graph__(self):
         out = self.expr
-        out = out.simplify()
+        out = out.optimize(fuse=False)
         return out.__dask_graph__()
 
     def __dask_keys__(self):
         out = self.expr
-        out = out.simplify()
+        out = out.optimize(fuse=False)
```
rjzamora, line 130: **Just a note** (**not** a suggestion for this PR):  We added the `simplify` step to `__dask_graph__`/`__dask_keys__` to deal with the fact that there may be "abstract" nodes in the expression graph. It **may** make sense to only call `lower` here (instead of `optimize`). Where lower would do something like:

```python
    def lower(self):
        expr = self
        while True:
            out = expr.lower_once()
            if out._name == expr._name:
                break
            expr = out
        return expr
```

Such a change would probably require a lot of tests to change, but may make the most sense (not sure about this yet).

rjzamora (commented):
I do think this approach makes sense. It is not technically "extensible", but I'm not convinced that it needs to be for now (especially since I still feel that "lower" is a fundamentally different kind of optimization than everything else we want to do in `simplify`.

This PR doesn't seen to add any `_lower` implementations or tests. Correct? After moving some `_simplify_down` logic over to `_lower`, it seems like many of the tests that are just using `simplify` will begin to fail (sorry - I know I was the one who suggested we use `simplify` in many places).

phofl:
Yeah I wanted to keep the diff small here. That's why I didn't move anything yet. And yes, we will probably have to change a bunch of tests for this :(

rjzamora:
>Yeah I wanted to keep the diff small here

Got it, that's fair.

rjzamora (approved):

phofl:
lets get this in, can always refactor. I'll take a stab at the tests now

PR #193, opened 2023-06-27
Implement AlignDivisions

phofl:
This is on top of #192 and #188 and basically an alternative to #166

@mrocklin I hope this is roughly what you had in mind for an AlignDivisions class that can be used for such tasks. I am assuming that we will have to push more stuff up here, but this should be a good starting point.

### DIFF
--- a/dask_expr/align.py
+++ b/dask_expr/align.py
@@ -2,7 +2,7 @@ import functools
 
 from tlz import merge_sorted, unique
 
-from dask_expr.expr import Expr, Projection
+from dask_expr.expr import Expr, Projection, is_broadcastable
 from dask_expr.repartition import RepartitionDivisions
 
 
@@ -49,22 +49,3 @@ class AlignDivisions(Expr):
         return RepartitionDivisions(
             self.frame, new_divisions=self.divisions, force=True
         )
-
-
-def is_broadcastable(dfs, s):
-    """
-    This Series is broadcastable against another dataframe in the sequence
-    """
-
-    def compare(s, df):
-        try:
-            return s.divisions == (min(df.columns), max(df.columns))
-        except TypeError:
-            return False
-
-    return (
-        s.ndim <= 1
-        and s.npartitions == 1
-        and s.known_divisions
-        and any(compare(s, df) for df in dfs if df.ndim == 2)
-    )
--- a/dask_expr/align.py
+++ b/dask_expr/align.py
@@ -15,11 +15,8 @@ class AlignDivisions(Expr):
 
     @functools.cached_property
     def dfs(self):
-        _is_broadcastable = functools.partial(is_broadcastable, self.dependencies())
         return [
-            df
-            for df in self.dependencies()
-            if df.ndim > 0 and not _is_broadcastable(df)
+            df for df in self.dependencies() if df.ndim > 0 and not is_broadcastable(df)
         ]
 
     def _divisions(self):
--- a/dask_expr/align.py
+++ b/dask_expr/align.py
@@ -7,13 +7,11 @@ from dask_expr.repartition import RepartitionDivisions
 
 
 class AlignDivisions(Expr):
-    @property
-    def _meta(self):
-        return self._frame._meta
+    _parameters = ["frame"]
 
     @property
-    def _frame(self):
-        return self.operands[0]
+    def _meta(self):
+        return self.frame._meta
 
     @functools.cached_property
     def dfs(self):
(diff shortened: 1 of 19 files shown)


dask_expr/expr.py
```diff
@@ -1671,14 +1708,33 @@ def optimize(expr: Expr, fuse: bool = True) -> Expr:
     simplify
     optimize_blockwise_fusion
     """
-    expr = expr.simplify()
+    expr = expr.simplify().lower()
```
rjzamora, line 1711: Related PR/discussion: https://github.com/dask-contrib/dask-expr/issues/194#issuecomment-1609848203

mrocklin:
I haven't gone through this in depth.  I think that it would be good to agree on some tests around expression tree structure.  For example I would expect that after optimization expressions like `df.x + df.y` are just `Add(Projection(df, "x"), Projection(df, "y"))`.  Things like that would make me feel more reassured that we're not injecting odd things into the expression.

dask_expr/align.py
```diff
@@ -0,0 +1,74 @@
+import functools
+
+from tlz import merge_sorted, unique
+
+from dask_expr.expr import Expr, Projection
+from dask_expr.repartition import RepartitionDivisions
+
+
+class AlignDivisions(Expr):
```
mrocklin, line 9: You can still add `_parameters` here, even if they're incomplete:

```suggestion
class AlignDivisions(Expr):
    _parameters = ["frame"]
```

dask_expr/align.py
```diff
@@ -0,0 +1,74 @@
...
+    @property
+    def _meta(self):
+        return self._frame._meta
+
+    @property
+    def _frame(self):
+        return self.operands[0]
```
mrocklin, line 16:
```suggestion
```

Maybe remove this and use `self.frame`?

dask_expr/align.py
```diff
@@ -0,0 +1,74 @@
...
+
+    def _simplify_up(self, parent):
+        if isinstance(parent, Projection):
+            return type(self)(
+                self._frame[parent.operand("columns")], *self.operands[1:]
+            )
+
+    def _lower(self):
+        if not self.dfs:
+            return self._frame
```
mrocklin, line 41: Is this possible?  Isn't`self._frame in self.dfs`

dask_expr/collection.py
```diff
@@ -118,29 +129,32 @@ def __getitem__(self, other):
...
     def __dask_postcompute__(self):
-        state = self.simplify()
+        state = self.simplify().lower()
         if type(self) != type(state):
             return state.__dask_postcompute__()
         return _concat, ()
 
     def __dask_postpersist__(self):
-        state = self.simplify()
+        state = self.simplify().lower()
```
mrocklin, line 157: I think there was some conversation around simplifying after lowering.  Maybe we want to replace these with `.optimize()` which then can choose to alternate between the two if it wants?

This gets a little weird because of fusion, but maybe we start controlling that with config instead of a keyword.

dask_expr/expr.py
```diff
@@ -276,6 +276,43 @@ def _simplify_down(self):
...
+            out = expr._lower()
+            if out is None:
+                out = expr
+            if not isinstance(out, Expr):
+                return out
+            if out._name != expr._name:
+                expr = out
+                continue
+
+            # Lower all children
```
mrocklin, line 293: I think I like the fact that `lower_up` doesn't exist

dask_expr/expr.py
```diff
@@ -1671,14 +1708,33 @@ def optimize(expr: Expr, fuse: bool = True) -> Expr:
...
+    """Traverse through tree to find ancestors that are not blockwise or are IO"""
+    stack = [expr]
+    while stack:
+        e = stack.pop()
+        if isinstance(e, IO):
+            yield e
+        elif isinstance(e, Blockwise):
+            stack.extend(e.dependencies())
+        else:
+            yield e
```
mrocklin, line 1729: Maybe broadcastable expressions should also be excluded here

`df.x + a.sum()` and `df.y + b.sum()` should still be co-aligned even if `a` and `b` are quite different.  (this would make a good test somewhere probably).

dask_expr/align.py
phofl: All of them could be broadcastable

dask_expr/collection.py
phofl: We had some discussion about this in #194 and agreed that a more general mechanism probably makes more sense. Adding lower felt not extensible enough (starting with the simplify -> lower -> simplify things)

dask_expr/expr.py
phofl: That's a good point, better excluding them here than in AlignDivisions. Added this to #188

phofl:
I'll add some explicit tests tomorrow. We have implicit ones already.

phofl:
Updated now after lower is in

dask_expr/align.py
```diff
@@ -0,0 +1,51 @@
+import functools
+
+from tlz import merge_sorted, unique
+
+from dask_expr.expr import Expr, Projection, is_broadcastable
+from dask_expr.repartition import RepartitionDivisions
+
+
+class AlignDivisions(Expr):
```
mrocklin, line 9: Thoughts on renaming to `AlignPartitions`?  

It feels like we're moving away from the term divisions in general.  Maybe partitions has more longevity.

phofl: Renamed

phofl:
Fixed the failing test

mrocklin (approved):

Issue #194, opened 2023-06-27
[Proposal] Add "phases" argument to `Expr.simplify`


rjzamora:
I spent some time thinking about #166, #181 and #186 yesterday, and experimented a bit with optimization ordering. I played with  some relatively complex approaches, but ultimately decided to propose something pretty simple.

**General Idea**: We add an optional `phases: tuple` argument to `simplify`, move the graph-traversal logic into a `_simplify` method, and add an `allow_group` argument to `_simplify`, `_simplify_up`, and `simplify_down`.

**Rough Code Outline**:

```python
class AbstractExpr(Expr):
    # "Example" Expr class
    
    _simplified = None  # Use attribute for hashing (can be done in separate work)

    def _simpify_down(self, allow_group: tuple):
        # Hypothetical example
        if "repartition" in allow_group:
            # Hold off on "lowering" until "repartition" phase
            return RepartitionDivisions(...)
    
    def _simpify_up(self, parent, allow_group: tuple):
        # Probably don't need to check allow_group for projection
        if isinstance(parent, Projection):
            ...
                
def simplify(self, phases=("abstract", "lower", "repartition")):
    # NOTE: Could always add/change phases later on
    
    # Loop over all "phases"
    expr = self
    for p in range(len(phases)):
        allow_group = phases[:p+1]
        
        # For each phase, we do the usual "simplify" loop
        # calling _simplify_down, _simplify_up and _simplify
        # until nothing changes. We pass in `allow_group`
        # to enforce rough ordering of optimizations.
        ...
        
    return expr

def _simplify(self, allow_group: tuple):
    
    # Use chached result if available
    if (self._simplified or {}).get(allow_group) is not None:
        return self._simplified
    
    # Usual "simplify" loop here, but `allow_group`
    # is passed to `_simplify_up` and `_simplify_down`
    expr = self
    ...
    
    # Cache result
    self._simplified = {allow_group: expr}
    return expr
```

rjzamora:
@phofl - This certainly overlaps with #193

Note that one of my earlier experiments did something like `simplify().lower()`, but I remembered discussing with @mrocklin a while back that we will probably end up wanting more than two simplify-like "phases" in the long run. Also, we will probably want to run the more-general simplify optimizations *after* lowering as well.

phofl:
Yes I think that I agree with you. This became more and more obvious while I was playing with the ``lower`` implementation. Mentioned something similar in #192 as well. 

[...]

PR #195, opened 2023-06-27
Add phases to ``Expr.simplify``

rjzamora:
Rough implementation of the proposal in #194

For now, I am only using two kinds of simplify categories (`"general"` and `"lower"`). In the future, we could always add categories like `"projection"` and `"predicate-pushdown"`.

Note that, in contrast to the plan in #194, this PR does not require "higher-priority" optimizations to be considered in every phase. Instead, the user is allowed to specify a specific tuple of simplify categories for each phase (e.g. `phases=["general", "lower", ("general", "lower")]`.


mrocklin:
Hrm, the allow_group thing feels off to me.  I'm trying to think about why I dislike it and I think it's because the kind of optimization is included deeply within the optimization code itself.  I think that it would be good to keep things separate if possible.  This might mean that we have many instances of the same method name, maybe something like the following:

```python
@optimize("general")
def _simplify_down(...):
    ...

@optimize("lower")
def _simplify_down(...):
   ...
```

But if we're going to do this, then we might as well just use method names:

```python
def optimize_simplify(...):
    ...

def optimize_lower(...):
    ...
```

The benefit of naming optimization passes the way you do (or with the decorators) is that it makes things really extensible.  It's obvious how to add more kinds of passes.  

Mixing the names and the optimization code itself though definitely makes me nervous.  Maybe that's ok though, this is more of a gut reaction than anything.

rjzamora:
Yes, I think I shared the same gut reaction (and still do a bit). I played with the registration concept and also felt that we might as well as distinct methods.

I ultimately decided to lean hard in the direction of “extensible” for
Now. However, I was expecting that we may collectively decide to shift :)

dask_expr/collection.py
```diff
@@ -126,8 +126,8 @@ def __dask_keys__(self):
         out = out.simplify()
         return out.__dask_keys__()
 
-    def simplify(self):
-        return new_collection(self.expr.simplify())
+    def simplify(self, **kwargs):
```
phofl, line 129: Can we get rid of the kwargs here?

dask_expr/expr.py
```diff
@@ -208,39 +209,77 @@ def _layer(self) -> dict:
...
+        self,
+        phases: list
+        | tuple = (
+            "general",
+            (
+                "general",
+                "lower",
+            ),
+        ),
+    ):
```
phofl, line 222: I would prefer using

```
(("general", ), ("general, "lower"))
```

and being more strict below

dask_expr/expr.py
```diff
@@ -208,39 +209,77 @@ def _layer(self) -> dict:
...
-        changed:
-            whether or not any change occured
         """
+
+        # Loop over all phases
+        expr = self
+        for phase in phases:
+            allow_group = (phase,) if isinstance(phase, str) else phase
+            if not isinstance(allow_group, tuple):
+                raise TypeError("Each element of `phases` must be str or tuple")
```
phofl, line 245: I would be more strict here instead of introducing some magic (from a user pov). Raise if the phase isn't a tuple

phofl (commented):
FWIW using different methods felt too restrictive when working on the lower PR. It's only a gut feeling as well though. I can see where @mrockling is coming from, putting the checks inside the methods isn't ideal. 

I don't think we lock ourselves in no matter in which direction we decide to go now. We can always reiterate later.

mrocklin:
> FWIW using different methods felt too restrictive when working on the lower PR

From my perspective I kind of want each individual optimization step to be an isolated method.  This is probably mostly an aesthetic preference that comes from writing other optimization systems, where being able to inject or remove or track or measure each step has been helpful.  Pulling things apart into small units seems attractive to me.  Pushing everything together and having arbitrary Python code be the only control system feels off. 

That might not matter at all in practice here (we may not get that sophisticated) so maybe I'm over-thinking things.

rjzamora:
>Pushing everything together and having arbitrary Python code be the only control system feels off.

I definitely understand this feeling. I would also like for different flavors of "simplify"-based logic to be separated in a more systematic way.  To be honest, I'm still hoping to find a better compromise.

Some details that are currently challenging me:

- In the current state of this PR, the `_simplify_up`/`_simplify_down` implementations can make intelligent decisions about optimization ordering within the categories included in the `allow_group` tuple.
- The most natural way of organizing distinct optimization types by distinct methods is to split them into proper optimization passes, and to add the `phases` argument to `optimize` instead of `simplify` (e.g. `expr.optimize(phases=("simplify", "lower", "simplify", "fuse")`). However, this means:
  - We can no longer leverage the coupling behavior above (i.e. an expression cannot coordinate multiple types of optimizations in the same phase/pass).
  - We need to "fully" lower the expression graph during the "lower" phase, rather than incrementally performing general optimizations as abstract nodes like `Merge` are lowered to other abstract nodes like `Shuffle`. (Not sure if this matters in practice, but our ability to do this disappears).
  - Non-"lower" optimizations will no longer be allowed to return abstract nodes (e.g. `Shuffle`, `Repartition`, etc..), because there is no guarantee that a "lower" phase will always follow a "simplify" phase.

rjzamora:
Given the challenges listed above, I'm leaning towards a design that preserves the `expr.simplify(phases=)` approach, but requires `Expr` definitions to register distinct `_simplify_up` and `_simplify_down` methods for each optimization type (i.e. "general" or "lower" for now).  This means the `_simplify_up`/`_simplify_down` logic cannot coordinate multiple optimization types (something I don't believe to be very important anyway), but we can still perform multiple optimization types within the same phase.

dask_expr/collection.py
rjzamora: Yeah, I didn't want to set a `phases` default in two places, but we can use a `phases=None` default to avoid this.

dask_expr/expr.py
rjzamora: Yeah, sounds good.

dask_expr/expr.py
```diff
@@ -268,13 +303,45 @@ def simplify(self):
...
+        return None
+
+    def _simplify_down_lower(self):
+        return None
+
+    def _simplify_up_general(self, parent: Expr):
+        return None
+
+    def _simplify_up_lower(self, parent: Expr):
+        return None
```
rjzamora, line 344: We technically don't *need* these methods to exist, since `Expr._simplify_down`/`up` will return `None` when an `AttributeError` is encountered. However, the code is **much** faster when we don't need to handle missing-attribute errors often (which also means a dispatch-based approach would also be slower).

I still like the idea of handling the `AttributeError` case, because this allows users to add/use custom optimization types.

mrocklin:
If I were to do this without an extensible phases system, using just simplify and lower, then I would probably write it as follows.

```python
def optimize(expr):
    old = None
    while old and old._name != expr._name:
        old = expr
        expr = expr.simplify()
        expr = expr.lower()
    return expr
```

We keep alternating between the two (first doing simplify) until they converge.

mrocklin:
I'd like something extensible.  Mostly because I don't want us to paint ourselves into a corner.  In my view phases are good, but not if they force us to mix everything.  That's painting ourselves into a corner.  Separating things into separate simplify/lower methods sounds fine to me, even if the orchestration function (like my `optimize` function above) is not yet very sophisticated.  We can make it more sophisticated in the future without changing a bunch of optimization code.

rjzamora:
I also like [this](https://github.com/dask-contrib/dask-expr/pull/195#issuecomment-1613213048) approach on its face. (which is why it was one of the original plans)

Corners we *may* be painting ourselves into if we do this:

- `lower` will need to keep lowering nodes until everything is fully lowered and nothing is changing. This means that if intermediate lowering steps produce filters/projections, the fully-lowered state may block "general" optimizations like predicate-pushdown and column projection.
- In the near future, I would like to be able to toggle optimizations like predicate-pushdown and column projection (for both performance analysis and debugging). Therefore, we will either need additional methods like `project_columns()`, or we will still need to add the phases concept within `simplify`.

mrocklin:
Good point about partial lowering.  

I'm also ok with something more sophisticated.  I just want the sophistication to be *outside* of the optimization function code.

mrocklin:
> Therefore, we will either need additional methods like project_columns(), or we will still need to add the phases concept within simplify

There is a wide world of options at our disposal :)

I think that you're ok with this, but I'll say it anyway.  Rather than have "phases" we can also have a soup of different optimizations, each tagged in different ways or part of different families.  We can then combine these optimizations in different ways.  In general when thinking of optimization I think it's good to separate out ...

-  The legal transfomations that can occur
- The control flow that applies those transformations in some order

Building the transformations is what we've mostly done.  This is in the `_simplify_*` methods today.  The control flow is entirely in the `simplify` method.  Separating these allows us to evolve the optimization system without having to rewrite all of the transformations (really annoying to do!).  I'm going to get very nervous if we start putting control flow stuff in the transformation code.  That way leads (historically in my experience) to either great pain, for a system that is hard to evolve.  

If we're expecting to have more phases/families/... to optimization then I think we probably want to start separating out transformations and tagging them like the following:

```python
@optimization("simplify", "column-projection")
def _transform_down(expr):
    ...

@optimization("lower")
def _transform_down(expr):
    ...
```

Then we can screw around with control flow questions like "do we want to simplify every time after we lower?" separately from modifying all of the transformation code.

rjzamora:
Thanks for following up @mrocklin - I'm pretty sure we agree on the kinds of things we want to do. I also agree that the ideal solution will separate the complexity/sophistication from node-local optimization logic.

I'm in no rush to merge anything here until it is clear that we are on a promising path.

I like the idea of tagging methods very much. That said, I am still struggling to think of a clean way to to configure how distinguish different optimization flavors **and** different graph traversal/re-write approaches. It feels as though `simplify` is more of a distinct approach for traversing and rewriting the graph, and cannot be categorized in the same was as the local logic defined in `_simplify_up` and `_simplify_down`.

rjzamora:
Since this PR still falls short of what we want, perhaps it makes sense to pick up on @phofl 's work in https://github.com/dask-contrib/dask-expr/pull/192 as our short-term approach to separating general "simplify" and "lowering" logic.  The more I think about the "lower" optimization, the more I feel like it is conceptually distinct from the other flavors of "simplify" (and deserves a distinct pass over the expression graph).

I think my only preference would be for optimize to look something like:

```python
def optimize(expr, fuse=True):

    result = expr
    while True:
        out = result.simplify().lower_once()
        if out._name == result._name:
            break
        result = out

    if fuse:
        result = optimize_blockwise_fusion(result)

    return result
```

(note the use of `lower_once` instead of `lower`) to capture iterative partial lowering. In the near future, we can always work out a system to tag different flavors of simplify/lower.

mrocklin:
> It feels as though simplify is more of a distinct approach for traversing and rewriting the graph, and cannot be categorized in the same was as the local logic defined in _simplify_up and _simplify_down.

To be clear, `simplify` is very different than `_simplify_up` and `_simplify_down` to me.  I think that `simplify` is control flow code (something that is likely to change in the future as we figure out more about optimization while the `_simplify_up` and `_simplify_down` implementations are transformations that are likely to remain valid for a long time.  

 > I am still struggling to think of a clean way to to configure how distinguish different optimization flavors and different graph traversal/re-write approaches

I think that as long as we label the transformations well (simplify vs lower vs other future names) then we can explore the problem of applying transformations as an interesting problem without a lot of code burden.  

I'm not sure I fully undestood your last comment though.  I'd be happy to chat live if you're free early next week.

rjzamora:
>I think that simplify is control flow code (something that is likely to change in the future as we figure out more about optimization while the _simplify_up and _simplify_down implementations are transformations that are likely to remain valid for a long time.

Yes. I think this is exactly what I am attempting to say. It is easy to write different methods for different kinds of transformation logic (i.e. different flavors of `_simplify_up` and `_simplify_down`). My only point was that we probably want different kinds of control flow for different kinds of transformation logic as well.

Given my concerns about distinguishing control flow from transformation code, #192 seems like a reasonable near-term solution.

PR #198, opened 2023-06-29
Ensure that df is immutable after from_pandas

phofl:
This bit me earlier...

Copy-on-Write would make our lives easier here

### DIFF
--- a/dask_expr/collection.py
+++ b/dask_expr/collection.py
@@ -795,10 +795,10 @@ def optimize(collection, fuse=True):
     return new_collection(expr.optimize(collection.expr, fuse=fuse))
 
 
-def from_pandas(*args, **kwargs):
+def from_pandas(data, *args, **kwargs):
     from dask_expr.io.io import FromPandas
 
-    return new_collection(FromPandas(*args, **kwargs))
+    return new_collection(FromPandas(data.copy(), *args, **kwargs))
 
 
 def from_graph(*args, **kwargs):
--- a/dask_expr/io/tests/test_io.py
+++ b/dask_expr/io/tests/test_io.py
@@ -198,6 +198,14 @@ def test_from_pandas(sort):
     assert_eq(df, pdf)
 
 
+def test_from_pandas_immutable():
+    pdf = pd.DataFrame({"x": [1, 2, 3, 4]})
+    expected = pdf.copy()
+    df = from_pandas(pdf)
+    pdf["z"] = 100
+    assert_eq(df, expected)
+
+
 def test_parquet_complex_filters(tmpdir):
     df = read_parquet(_make_file(tmpdir))
     pdf = df.compute()


dask_expr/collection.py
```diff
@@ -795,10 +795,10 @@ def optimize(collection, fuse=True):
     return new_collection(expr.optimize(collection.expr, fuse=fuse))
 
 
-def from_pandas(*args, **kwargs):
+def from_pandas(data, *args, **kwargs):
     from dask_expr.io.io import FromPandas
 
-    return new_collection(FromPandas(*args, **kwargs))
+    return new_collection(FromPandas(data.copy(), *args, **kwargs))
```
rjzamora, line 801: This feels unfortunate, but also makes sense.

We don't need to do this explicitly in `dask.dataframe`, because we generate the graph and `_meta` right away.  The unfortunate part is that this means we will ultimately need to copy the data twice: first we copy here, then we copy the data when we use slices to build the graph.

All the more reason to avoid `from_pandas` for performance-sensitive applications.

rjzamora (approved):

dask_expr/collection.py
phofl: That' actually not completely correct. The DataFrame is only copied if you have to sort it. Horizontal slicing creates a view, which is mutable. This was a bug in dask/dask as well, see https://github.com/dask/dask/pull/10383. The difference is that we have a new object which ensures that a regular setitem like in my test doesn't propagate. 

We could get by with a shallow copy if we decide to enable Copy-on-Write by default. I think it's a bit early for this though.

rjzamora: >Horizontal slicing creates a view, which is mutable

Are you saying it's a view in general, or just when all columns have the same dtype?

```python
In [1]: import pandas as pd

In [2]: pdf = pd.DataFrame({"x": range(10), "y": ["a"]*10})

In [3]: pdf.iloc[:2]._is_view
Out[3]: False

In [4]: pdf[["x"]].iloc[:2]._is_view
Out[4]: True
```

>This was a bug in dask/dask as well, see https://github.com/dask/dask/pull/10383

Interesting, thanks for linking.

phofl: ``_is_view`` only works properly if it's only one Block, but this doesn't mean that your object isn't a view (generally, I wouldn't count on ``_is_view``, as far as I know it's only correct if you slice columns (but I am guessing here, never bothered much with it. Joris and I  implemented a significantly more precise mechanism for Copy-on-Write, but it's not really accessible for users))

The easiest way to test whether a new object is a view or not is the following:

```
pdf = pd.DataFrame({"x": range(10), "y": ["a"]*10})
subset = pdf.iloc[:2]
subset.iloc[0,0] = 100
pdf
```

PR #201, opened 2023-07-03
Implement PoC set_index

phofl:
This is a trimmed down implementation of set_index. I'll add support for the remaining keywords in follow-ups. 

This will be moved to ``lower`` or a similar optimization step as soon as we have merged something that implements it. Hence, I put it into ``simplify_down`` for now.

### DIFF
--- a/dask_expr/collection.py
+++ b/dask_expr/collection.py
@@ -710,6 +710,8 @@ class DataFrame(FrameBase):
         return new_collection(Eval(self.expr, _expr=expr, expr_kwargs=kwargs))
 
     def set_index(self, other, drop=True, sorted=False, divisions=None):
+        if isinstance(other, DataFrame):
+            raise TypeError("other can't be of type DataFrame")
         if isinstance(other, Series):
             if other._name == self.index._name:
                 return self
--- a/dask_expr/collection.py
+++ b/dask_expr/collection.py
@@ -10,6 +10,7 @@ from dask.base import DaskMethodsMixin, is_dask_collection, named_schedulers
 from dask.dataframe.core import (
     _concat,
     _Frame,
+    check_divisions,
     is_dataframe_like,
     is_index_like,
     is_series_like,
@@ -37,6 +38,7 @@ from dask_expr.reductions import (
     ValueCounts,
 )
 from dask_expr.repartition import Repartition
+from dask_expr.shuffle import SetIndex
 
 #
 # Utilities to wrap Expr API
@@ -687,6 +689,21 @@ class DataFrame(FrameBase):
 
     def eval(self, expr, **kwargs):
         return new_collection(Eval(self.expr, _expr=expr, expr_kwargs=kwargs))
+
+    def set_index(self, other, drop=True, sorted=False, divisions=None):
+        if isinstance(other, Series):
+            if other._name == self.index._name:
+                return self
+        else:
+            if other == self.index.name:
+                return self
+
+        if divisions is not None:
+            check_divisions(divisions)
+        other = other.expr if isinstance(other, Series) else other
+        return new_collection(
+            SetIndex(self.expr, other, drop, sorted, user_divisions=divisions)
+        )
 
 
 class Series(FrameBase):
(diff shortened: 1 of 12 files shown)


dask_expr/shuffle.py
```diff
@@ -609,3 +623,67 @@ def operation(cls, df, index, name: str, npartitions: int):
...
+            other = self._other
+        return self.frame._meta.set_index(other, drop=self.drop)
+
+    @property
+    def other(self):
+        if isinstance(self._other, Expr):
+            return self._other
+        return self.frame[self._other]
+
+    @functools.lru_cache  # noqa: B019
```
phofl, line 651: Are we ok with caching here?

dask_expr/collection.py
```diff
@@ -707,6 +709,21 @@ def select_dtypes(self, include=None, exclude=None):
     def eval(self, expr, **kwargs):
         return new_collection(Eval(self.expr, _expr=expr, expr_kwargs=kwargs))
 
+    def set_index(self, other, drop=True, sorted=False, divisions=None):
+        if isinstance(other, Series):
+            if other._name == self.index._name:
+                return self
+        else:
+            if other == self.index.name:
+                return self
```
mrocklin, line 718:
```suggestion
        elif other == self.index.name:
            return self
```

Not a big deal, but maybe simpler

dask_expr/collection.py
```diff
@@ -707,6 +709,21 @@ def select_dtypes(self, include=None, exclude=None):
...
+
+        if divisions is not None:
+            check_divisions(divisions)
+        other = other.expr if isinstance(other, Series) else other
```
mrocklin, line 722: Might a user do `df.set_index(some_dataframe)`?

dask_expr/expr.py
```diff
@@ -1173,6 +1174,38 @@ class Drop(Elemwise):
...
+        df2 = df.set_index("_index", drop=True)
+        df2.index.name = index_name
+        return df2
+
+
+class SortIndexBlockwise(Blockwise):
+    _parameters = ["frame"]
+    operation = M.sort_index
+
+
```
mrocklin, line 1232: These should maybe stay close to the set_index implementation in `shuffle.py`?

dask_expr/shuffle.py
mrocklin: Fine with me.  Thoughts on merging this method with `_divisions` above?  Will this ever be used separately?

dask_expr/collection.py
phofl: Yep potentially, we should disallow this though

dask_expr/shuffle.py
phofl: We will have to compute more statistics ("presorted" for example") in this method when adding support for other keywords and code-paths, which will make this method different from ``_divisions``.

dask_expr/expr.py
phofl: Oh that makes sense, thx.

dask_expr/collection.py
phofl: Added a check

dask_expr/shuffle.py
mrocklin: I'm not sure I understand, will there be times when we call this method outside of `_divisions`

phofl: Yes!

mrocklin: Then happy to retract my comment.  Thanks for clarifying.

dask_expr/expr.py
mrocklin: Is there a reason why this one survived?  I could imagine two reasons:
1.  At some point we're thinking about `sort=False`.
2. Copy-paste mistake

In either case I think it probably makes sense to move.  It isn't really related to any other code in that this file.

mrocklin (approved):
Fine to merge whenever from my perspective.

dask_expr/expr.py
phofl: This is more broad for me, like has other applications than ``set_index`` only. It's probably better to move this for now as well and move back in case we need it at some point

PR #205, opened 2023-07-04
Implement more imputs for assign

phofl:


dask_expr/collection.py
```diff
@@ -451,12 +452,21 @@ class DataFrame(FrameBase):
...
+        data = self.copy()
         for k, v in pairs.items():
-            if not isinstance(v, Series):
-                raise TypeError(f"Column assignment doesn't support type {type(v)}")
             if not isinstance(k, str):
                 raise TypeError(f"Column name cannot be type {type(k)}")
-            result = new_collection(expr.Assign(result.expr, k, v.expr))
+
+            if callable(v):
+                v = v(data)
```
mrocklin, line 461: Maybe this should be data.map_partitions(v)?  That might be safer for more kinds of callables.

phofl: I am a bit on the fence here, this was my first thought as well when I looked up the dask/dask implementation. 

Being too loose might cause more troubles than it's worth. I think I'd rather tell the user to do this himself if he really wants to.

mrocklin: OK, happy to defer to you

PR #206, opened 2023-07-05
Implement presorted for set_index

phofl:


dask_expr/tests/test_shuffle.py
```diff
@@ -136,3 +136,10 @@ def test_set_index(df, pdf):
 
     with pytest.raises(TypeError, match="can't be of type DataFrame"):
         df.set_index(df)
+
+
+def test_set_index_pre_sorted(pdf):
+    pdf = pdf.sort_values(by="y", ignore_index=True)
+    df = from_pandas(pdf, npartitions=10)
+    assert_eq(df.set_index("y"), pdf.set_index("y"))
+    assert_eq(df.set_index(df.y), pdf.set_index(pdf.y))
```
mrocklin, line 145: It seems like more testing might be warranted here.  Maybe we assert that all expressions in the graph are blockwise?

phofl: Oh that's a good call, will add tests

phofl: This should do the trick

dask_expr/tests/test_shuffle.py
```diff
@@ -136,3 +138,23 @@ def test_set_index(df, pdf):
...
+
+def test_set_index_pre_sorted(pdf):
+    pdf = pdf.sort_values(by="y", ignore_index=True)
+    df = from_pandas(pdf, npartitions=10)
+    q = df.set_index("y")
+    assert_eq(q, pdf.set_index("y"))
+    check_expression_tree_is_blockwise(q.simplify().expr)
+    q = df.set_index(df.y)
+    assert_eq(q, pdf.set_index(pdf.y))
+    check_expression_tree_is_blockwise(q.simplify().expr)
```
mrocklin, line 151:
```suggestion
    assert all(isinstance(expr, Blockwise) for expr in q.simplify().find_operations(Expr))
```

This could probably be improved actually, see https://github.com/dask-contrib/dask-expr/pull/208

dask_expr/tests/test_shuffle.py
```diff
@@ -136,3 +138,28 @@ def test_set_index(df, pdf):
...
+    q = df.set_index(df.y)
+    assert_eq(q, pdf.set_index(pdf.y))
+    result = q.simplify().expr
+    assert (
+        len(
+            set(result.find_operations(Expr))
+            - set(result.find_operations((Blockwise, FromPandas)))
+        )
+        == 0
+    )
```
mrocklin, line 165: Just a nit, but you could probably just compare the sets directly `assert set(...) == set(...)`

mrocklin: Or `assert all(isinstance(expr, ...) for expr in result.walk())`

mrocklin (approved):
I haven't really reviewed the logic here, but fine to merge by me

Issue #186, opened 2023-06-23
Some thoughts on high-level reasoning


mrocklin:
I had two conversations recently with @fjetter and @phofl about reasoning about dataframes.  Often during different stages (expression building, optimization, task graph building) we want to know various attributes of our dataframe.  Here are a few things people have asked for in the past:

1.  Is it sorted along this column?
2. How many partitions do we have?
3. How many rows are in each partition?
4. What are the min/max values of each partition in a sorted dataframe so that we can do joins better?

Historically we relied on `divisions` to answer many of these questions, and we're still used to doing that.  However, this isn't great, because divisions can change as we optimize (maybe we decide to repartition based on worker size, for example) and because divisions doesn't hold all of the infomation we want (like lengths of partitions).  

Instead, I think that we need to get used to asking questions of expressions by either inspecting the current tree, or by using other expressions.  Two examples:

1.  @rjzamora  added a `Lengths` expression, which computes the length of each partition, and knows how to optimize itself through Elemwise and ReadParquet calls so that, oftentimes, we can convert a `Lengths(df)` expression into a `Literal`.  

    This is an example of using an expression to answer a question about an existing expression

2.  We could ask if two dataframes share the same partition alignment by asking if they have the same ancestors modulo `Blockwise`.  For example the following are co-aligned:

    ```python
    df = read_parquet(...)
    a = df.x[df.z > 1] + 10
    b = df.y - df.z.sum()
    assert are_co_aligned(a, b)
    ```

    This is an example of looking at an expression tree to answer a question.  

3.  Similar to above we might infer that a column is sorted if it had previously been the target of a `set_index` or `sort_values` call, and was only operated on by monotonic expressions (we'd have to add this to known monotonic expressions)

General principles: 

1.  rely on divsiions less
2. try to only capture user inputs in Expression operands
3. think about asking questions using expression structure, or with new expressions with fun optimizations

PR #207, opened 2023-07-05
Ensure that ``divisions`` are available after ``set_index``

phofl:
I missed that initially. We should make them available if we know them like here

### DIFF
--- a/dask_expr/shuffle.py
+++ b/dask_expr/shuffle.py
@@ -671,10 +671,15 @@ class SetPartition(SetIndex):
         )
 
         if isinstance(self._other, Expr):
-            index_set = _SetIndexPostSeries(shuffled, self.other._meta.name)
+            index_set = _SetIndexPostSeries(
+                shuffled, self.other._meta.name, new_divisions=self._divisions()
+            )
         else:
             index_set = _SetIndexPostScalar(
-                shuffled, self.other._meta.name, drop=self.drop
+                shuffled,
+                self.other._meta.name,
+                drop=self.drop,
+                new_divisions=self._divisions(),
             )
 
         return SortIndexBlockwise(index_set)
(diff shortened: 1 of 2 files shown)


dask_expr/shuffle.py
```diff
@@ -691,17 +696,23 @@ def _meta(self):
 
 
 class _SetIndexPostScalar(Blockwise):
-    _parameters = ["frame", "index_name", "drop"]
+    _parameters = ["frame", "index_name", "drop", "new_divisions"]
```
mrocklin, line 699: I'm uncomfortable putting divisions explicitly into operands.  It makes things somewhat brittle and hard to change I think.

Can this class not look up the expression tree to determine these values?

https://github.com/dask-contrib/dask-expr/issues/186

phofl: Only ``SetPartition`` and ``SetIndex`` are aware of the divisions, since the calculation happens there. But both classes aren't part of the final expression tree after the simplify steps since they are removed in ``_simplify_down``. I think we have to pass this information along in ``_simplify_down`` in some way.

A related note: This will move into ``_lower`` as soon as it is available, so we would lock that in later on during the optimization process.

phofl: Will put up a doc PR shortly.

``SetIndex`` will make the decision if we have to shuffle our DataFrame or if we are already sorted so that a simple blockwise ``set_index`` followed by a blockwise ``sort_index`` is sufficient. So we have to know the divisions to make that decision (for now). This is implemented in #206

We could (theoretically) move this to compute time, but this would significantly restrict follow up operations right now (e.g. set_index followed by a merge, merge wouldn't have a clue that a blockwise merge might be sufficient).

We could also go in a different direction and use a "``set_index`` is dumb" approach, but this would place a bigger burden on our users to give us the necessary information

I completely agree that this should come as late as possible, but I think we have to do this in ``optimize`` (for now at least). 

This describes the current situation. I think there are some options available that we can look into. For example, ``SetIndex`` could inspect the expression tree to check whether knowing the divisions right now makes sense or if we can defer to compute time. But that's not what I am aiming for right now. I'd like to get ``set_index`` running as good as possible without going down any rabbit holes.

PR #208, opened 2023-07-05
Add walk method

mrocklin:


dask_expr/expr.py
```diff
@@ -703,8 +696,24 @@ def find_operations(self, operation: type) -> Generator[Expr]:
             for dep in node.dependencies():
                 stack.append(dep)
 
-            if isinstance(node, operation):
-                yield node
+            yield node
+
+    def find_operations(self, operation: type) -> Generator[Expr]:
```
phofl, line 701: I think that using multiple operation types here would be helpful

mrocklin: Sure.  Pushed.

phofl (approved):
I think using multiple operation types here would be helpful

Edit: Sorry some caching issue

PR #209, opened 2023-07-05
Add small docstrings to SetIndex

phofl:
I hope that this helps a bit

### DIFF
--- a/dask_expr/shuffle.py
+++ b/dask_expr/shuffle.py
@@ -617,6 +617,25 @@ class AssignPartitioningIndex(Blockwise):
 
 
 class SetIndex(Expr):
+    """Abstract ``set_index`` class.
+
+    Simplifies (later lowers) either to Blockwise ops if we are already sorted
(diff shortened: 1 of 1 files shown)


dask_expr/shuffle.py
```diff
@@ -653,6 +672,23 @@ def _simplify_down(self):
 
 
 class SetPartition(SetIndex):
+    """Shuffles the DataFrame according to its new divisions.
+
+    Simplifies the Expression to blockwise pre-processing, shuffle and
+    blockwise post-processing expressions.
```
mrocklin, line 678: Nothing to do here, just a thought: "should we have multiple expressions for this, or just one expression and a more complex `_layer` method?"

The benefit of one expression is that it's probably simpler to reason about and maintain.

The benefit of multiple expressions is that it probably optimizes better (blockwises can be fused, for example)

phofl: We will reuse SetPartition for sort values, that’s why I made 2

PR #210, opened 2023-07-05
Use lower instead of simplify_down for some expressions

phofl:
cc @rjzamora 

Yikes the failures were caused by a typo...

This showed that we need better test coverage for the lower steps though, will try to look into it in the coming days.

### DIFF
--- a/dask_expr/concat.py
+++ b/dask_expr/concat.py
@@ -59,7 +59,7 @@ class Concat(Expr):
 
         return [None] * (sum(df.npartitions for df in dfs) + 1)
 
-    def _simplify_down(self):
+    def _lower(self):
         dfs = self._frames
         cast_dfs = []
         for df in dfs:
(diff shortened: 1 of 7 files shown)


rjzamora (approved):
Thanks @phofl - This PR seems fine as it only claims to use lower for "some" expressions.

Some other expressions that still need `_lower` methods:

- `Head` (need to split some of the `_simplify_down` logic)
- `Tail` (need to split some of the `_simplify_down` logic)
- `reductions.Mean`
- `groupby.Mean`
- `groupby.Std`
- `JoinRecursive`

Note that I found these "missing" cases by adding `Expr.lower` and using that method instead of `optimize` in `__dask_graph__`, `__dask_keys__`, `__dask_postcompute__`, and `__dask_postpersist__`. Although I think that I *like* the fact that we automatically optimize the expression before a task graph is materialized, we should probably only **require** that the `lower` pass takes place.

PR #212, opened 2023-07-05
Support cudf as a DataFrame backend

rjzamora:
- Includes the necessary changes for cudf to be used as an alternative to pandas as the DataFrame backend
- Uses the `DASK_DATAFRAME__BACKEND` environment variable (with default of `"pandas"`) to set the backend library in `test_io.py` and `test_collection.py`. This approach can be expanded incrementally throughout the test suite without changing much code. I added skip/xfail statements for cudf in a few places, but we don't need to do this for all tests that fail with cudf (it is valuable to be able to "run" tests with a cudf backend either way).
- Does not work with latest release of cudf, nor the cudf nightly, because neither of these cudf versions support pandas>=2
  - I am currently working with a variation of the [pandas_2.0_feature_branch](https://github.com/rapidsai/cudf/tree/pandas_2.0_feature_branch) development branch


mrocklin:
Oh cool.  I'm curious to see what comes here.

My sense (correct me if I'm wrong) is that we probably don't want to annotate every test in order to support GPUs.  I'm thinking that this is just a temporary thing as you're playing around.  If not then let's chat.

rjzamora:
>My sense (correct me if I'm wrong) is that we probably don't want to annotate every test in order to support GPUs. I'm thinking that this is just a temporary thing as you're playing around. If not then let's chat.

Right, I'd rather not annotate everything, and I'm definitely still exploring here.

It isn't difficult to parameterize the `df` fixture to use both pandas and cudf (when available). However, it turns out that many tests fail for cudf. In some cases, the failures are from bugs (e.g. https://github.com/rapidsai/cudf/pull/13666). In other cases it's because of limits in cudf's API coverage. For example, `explode` doesn't work because we assume that the backend supports a list of `columns` (and [cudf doesn't](https://github.com/rapidsai/cudf/issues/10271)), and `idxmin`/`max` doesn't work, because cudf only has this operation implemented for `groupby`.

At the moment, I'm thinking that I should just configure `pdf`/`df` with something like:

```python
@pytest.fixture(
    params=[
        "pandas",
        pytest.param("cudf", marks=pytest.mark.skipif(cudf is None, reason="cudf not found.")),
    ]
)
def backend(request):
    yield request.param
```

I'm thinking this would allow us to skip (maybe `xfail`?) specific tests for the "cudf" backend instead of annotating every test. That said, I'm still feeling open to completely different approaches. I'm certainly preoccupied with finding/fixing the actual bugs that I'm finding.

rjzamora:
**Update**:

- This PR is currently blocked by https://github.com/dask-contrib/dask-expr/pull/223 and https://github.com/rapidsai/cudf/pull/13692 (related to deterministic tokenization)
- In order to reduce the test complexity here, I am thinking that I may roll back the pandas/cudf parameterization, and instead use an environment variable (e.g. `TEST_DASK_EXPR_BACKEND`) to define the backend library to import/use. This would allow one to run tests with the cudf backend by executing something like: `TEST_DASK_EXPR_BACKEND=cudf py.test -v dask_expr`

dask_expr/io/tests/test_io.py
```diff
@@ -1,20 +1,25 @@
...
 from dask.dataframe.utils import assert_eq
 
 from dask_expr import from_dask_dataframe, from_pandas, optimize, read_csv, read_parquet
 from dask_expr._expr import Expr, Lengths, Literal, Replace
 from dask_expr._reductions import Len
 from dask_expr.io import ReadParquet
 
+# Import backend DataFrame library to test
+BACKEND = config.get("dataframe.backend", "pandas")
+lib = importlib.import_module(BACKEND)
```
rjzamora, line 16: @phofl - Any opinions on using this approach as a "switch" to test the `cudf` backend?

dask_expr/tests/test_collection.py
```diff
@@ -15,30 +17,34 @@
...
 
 @pytest.fixture
-def pdf():
-    pdf = pd.DataFrame({"x": range(100)})
-    pdf["y"] = pdf.x * 10.0
-    yield pdf
+def bdf():
+    bdf = lib.DataFrame({"x": range(100)})
+    bdf["y"] = bdf.x * 10.0
+    yield bdf
```
rjzamora, line 29: Note that I renamed `pdf` to `bdf` for "backend" DataFrame, but also happy to continue using the `pdf` convention for both backends.

mrocklin: I have a preference to stick with `pdf` and a general default for pandas naming.

rjzamora: Sounds good to me. Thanks

dask_expr/io/tests/test_io.py
rjzamora: Certainly interested in your thoughts as well @mrocklin

phofl: I think this is ok, although it will take me some time to get used to it😂

traveling today, will take a closer look at the whole pr when I am back

dask_expr/tests/test_collection.py
```diff
@@ -348,6 +379,8 @@ def test_repr(df):
 
 
 def test_combine_first_simplify(pdf):
+    if BACKEND == "cudf":
```
phofl, line 382: Can we do this as a decorator? 

```
@pytest.mark.xfail(BACKEND_IS_CUDF)
```

dask_expr/tests/test_collection.py
```diff
@@ -799,8 +837,9 @@ def test_drop_duplicates(df, pdf):
     assert_eq(df.drop_duplicates(subset=["x"]), pdf.drop_duplicates(subset=["x"]))
     assert_eq(df.x.drop_duplicates(), pdf.x.drop_duplicates())
 
-    with pytest.raises(KeyError, match=re.escape("Index(['a'], dtype='object')")):
-        df.drop_duplicates(subset=["a"])
+    if BACKEND == "pandas":
+        with pytest.raises(KeyError, match=re.escape("Index(['a'], dtype='object')")):
```
phofl, line 841: Why is this not failing for cudf?

dask_expr/io/parquet.py
```diff
@@ -681,6 +686,20 @@ def _update_length_statistics(self):
 #
 
 
+def _set_engine(engine=None, meta=None):
```
phofl, line 689: Can we rename to ``_set_parquet_engine``? This confused me a bit in ``_collections.py``

phofl (commented):
Couple of comments

dask_expr/tests/test_collection.py
rjzamora: Does it seem okay to just catch the KeyError, and to check that `"'a'"` is in the message?

dask_expr/tests/test_collection.py
rjzamora: Yeah, can/should definitely do this for the general case, but I *think* we still need to add an `if` statement for a few parameter-specific cases.

dask_expr/tests/test_collection.py
phofl: Yeah that's fine by me

dask_expr/_expr.py
```diff
@@ -54,6 +54,18 @@ def __init__(self, *args, **kwargs):
...
+            if not hasattr(dep, self._required_attribute):
+                # Raise a ValueError instead of AttributeError to
+                # avoid infinite recursion
+                raise ValueError(f"{dep} has no attribute {self._required_attribute}")
+
+    @property
+    def _required_attribute(self) -> str:
+        # Specify if the first `dependency` must support
+        # a specific attribute for valid behavior.
+        return None
```
rjzamora, line 68: @phofl - I ran into quite a few cases where the cudf backend was hanging (rather than quickly failing) for cases where a `cudf.DataFrame`/`Series` did not support the same attribute as `pd.DataFrame`/`Series` (e.g. `nbytes`, `align`, etc).

What do you think about doing something like this so that we can more-quickly detect that the dependency is missing a necessary attribute?

dask_expr/_reductions.py
```diff
@@ -42,6 +42,15 @@ class ApplyConcatApply(Expr):
...
 
+    # def __init__(self, *args, **kwargs):
+    #     super().__init__(*args, **kwargs)
+    #     if self._required_attribute:
+    #         dep = next(iter(self.dependencies()))._meta
+    #         if not hasattr(dep, self._required_attribute):
+    #             # Raise a ValueError instead of AttributeError to
+    #             # avoid infinite recursion
+    #             raise ValueError(f"{dep} has no attribute {self._required_attribute}")
+
```
rjzamora, line 53:
```suggestion
```

dask_expr/_expr.py
phofl: Oh yes I like this.

phofl (approved):

phofl:
thx!

PR #213, opened 2023-07-06
Implement simplify for Drop

phofl:
I like this approach since it reduces complexity in our other _simplify_up steps


mrocklin:
Right, I think we're learning here that if we can reduce operations to a core common subset then optimization is good.

phofl:
Yep that was exactly my thought as well

PR #219, opened 2023-07-10
Make filenames private

phofl:


mrocklin:
Fine by me.  This defends against a thing I don't really care about, but I don't care either way, so I'm happy to defer.

jrbourbeau (commented):
+1

Issue #218, opened 2023-07-07
We tend to read data multiple times from storage


phofl:
I exchange a couple of messages with Rick on this topic yesterday.

Our optimization logic has a weakness when it comes to reading data (the issue extends to other methods as well, but let's focus on I/O to keep it a bit simpler). We tend to read data multiple times when we have different branches in our tree from the same root. One example:

```
df = read_parquet("taxi.parquet")
df["tipped"] = df.tips != 0
df["bpf_valid"] = df.base_passenger_fare != 0

q = df[["tipped", "bpf_valid", "tips"]].sum()
```

``tips`` is read twice in this case. Once for the ``NE`` expression that is used for assignment and once more for the sum aggregation at the end. 

This is a problem that get's worse when the query gets more complicated. It's triggered if one operation operates on 2 DataFrames from the same root.  

Ideally, we would only read the data once.

```
read_parquet(path, columns=["base_passenger_fare", "tips"], ...)["tips"]
```

There might be reasons not to do this, e.g. memory pressure, but this is probably only an option when we are not reading from remote storage. Our current optimization logic does not provide a mechanism to address this.

This isn't first-order priority but we should address this soonish.

This extends to multiple operations as well:

```
df = read_parquet("taxi.parquet")
df = df.replace(1, 100).fillna(100)
df["tipped"] = df.tips != 0
df["bpf_valid"] = df.base_passenger_fare != 0

q = df[["tipped", "bpf_valid", "tips"]].sum()
```

We will apply ``replace`` and ``fillna`` to ``tips`` twice as well.

cc @rjzamora

mrocklin:
Maybe we can handle this in a final global optimization pass (like blockwise fusion) that happens after simplify/lower.  We would look at all leaves of the graph, identify leaves that can be merged, and then replace those nodes by projections on some common read.

mrocklin:
I'm hopeful that that isn't too hard

phofl:
Same here, follow up operations could get a little tricky. Sometimes we remove whole operations if they are referencing columns that are dropped later on. That _might_ make things a little tricky

mrocklin:
I think that after we do this pass there's no more simplification/lowering.  We've got the final read_parquet calls the rest of the tree will want.  We produce these with a single read_parquet call followed by some projections.  I think that everything should be smooth past then.  Anything I'm missing?

phofl:
The following is something that I had in mind:

```
df = read_parquet(...)
[...]

PR #220, opened 2023-07-10
Add global optimization to limit redundant work

rjzamora:
Rough exploration of a possible solution to #218

This PR is very far from being "ready for review", but I wanted to start experimenting in public. I don't like the current naming, API, or organization yet.


rjzamora:
@phofl @mrocklin - I'd be interested to get feedback on the general direction of this PR.

As far as I can tell, the logic in this PR should address the problematic examples described in #218. The basic idea is that we add a new optimization pass within `Expr.optimize` - *after* the `simplify` step and *before* the `optimize_blockwise_fusion` step. This new `simplify_global` optimization pass is similar to `Expr.simplify`, except (1) we pass along a reference to the final "root" expression, and (2) we always rewrite/modify children **before** parents. While (1) allows the local `_simplify_global` logic to compare the current `Expr` object with similar nodes in the global graph, (2) makes it easier for us to effectively "push back up" problematic `Projection` operations.

What I'm not sure about:

- Are there other (better) ways to get the same result? If not...
- What should we call this optimization pass? I used `simplify_global` for now to be as general as possible, but I cannot say that I'm a fan (next bullet point is related).
- Should the code for this optimization pass be more specific and isolated? I originally assumed that the `simplify_global` "control flow" could be used for other optimizations that need access to `root`. However, I am now thinking that we should be more explicit that the purpose of this logic is to "pull back" problematic column-projection state.

dask_expr/io/parquet.py
```diff
@@ -427,6 +427,46 @@ def columns(self):
...
+        # access by aggregating multiple operations with different
+        # column projections into the same operation.
+        alike = self._find_similar_operations(root, ignore=["columns", "_series"])
+        if alike:
+            # We have other ReadParquet operations in the expression
+            # graph that can be combined with this one, because
+            # the only difference is the column projection
+
+            # Find the column-projection union needed to combine
+            # the qulified ReadParquet operations
```
phofl, line 441: typo

phofl (commented):
This generally looks good to me. I think I agree with your assessment that we should limit this particular step to combining similar expressions for now. Maybe something like this as the name?

rjzamora:
>I think I agree with your assessment that we should limit this particular step to combining similar expressions for now. Maybe something like this as the name?

How about `combine_similar`?

phofl:
Sounds good!

phofl (approved):

phofl:
Ready to merge from your PoV?

rjzamora:
Thanks for the review @phofl !

>Ready to merge from your PoV?

I feel that the actual optimization logic is fine for now. It is completely separate from `simplify`, and can always be improved/removed as needed.

My only concern is that we may want to add some simple tests for the new `_substitute_parameters` and `_find_similar_operations` methods. Also, I wonder if those "methods" should be moved over to `_utils` for now? Wdyt?

phofl:
Yeah I think moving them over makes sense.
Not against tests, but I think this would be fine as is too

rjzamora:
>Yeah I think moving them over makes sense.
Not against tests, but I think this would be fine as is too

Okay - I think I'll merge so that this is out of the way.

PR #221, opened 2023-07-11
Implement _simplify_up for groupby agg

phofl:


dask_expr/tests/test_groupby.py
```diff
@@ -34,9 +34,17 @@ def test_groupby_numeric(pdf, df, api, numeric_only):
...
+    agg = getattr(g, api)(numeric_only=numeric_only)["y"]
+
+    expect = getattr(pdf.groupby("x"), api)(numeric_only=numeric_only)["y"]
+    assert_eq(agg, expect)
+
 
 @pytest.mark.parametrize("func", ["count", "value_counts", "size"])
-def test_groupby_no_numeric_only(pdf, df, func):
+def test_groupby_no_numeric_only(pdf, func):
+    pdf = pdf.drop(columns="z")
```
rjzamora, line 46: Curious about this?

dask_expr/tests/test_groupby.py
```diff
@@ -34,9 +34,17 @@ def test_groupby_numeric(pdf, df, api, numeric_only):
     expect = getattr(pdf.groupby("x"), api)(numeric_only=numeric_only)
     assert_eq(agg, expect)
 
+    g = df.groupby("x")
+    agg = getattr(g, api)(numeric_only=numeric_only)["y"]
+
+    expect = getattr(pdf.groupby("x"), api)(numeric_only=numeric_only)["y"]
+    assert_eq(agg, expect)
```
rjzamora, line 41: Ah - so this was actually failing before, because we were falling back to the "incorrect" `Reduction._simplify_up` logic. Correct?

dask_expr/tests/test_groupby.py
rjzamora: Never mind - I take it that this is just another variation on testing a projection after a groupby.

dask_expr/groupby.py
```diff
@@ -297,6 +307,16 @@ def combine_kwargs(self):
...
 
+    def _simplify_up(self, parent):
+        if isinstance(parent, Projection):
+            columns = sorted(set(parent.columns + self.by))
+            if columns == self.frame.columns:
+                return
+            return type(parent)(
+                type(self)(self.frame[columns], *self.operands[1:]),
+                *parent.operands[1:],
+            )
```
rjzamora, line 318: Logic looks fine here. Just leaving a note that this could also be `return type(self)(self.frame[columns], *self.operands[1:])[parent.operand("columns")]` (though I'm not sure that is any more "readable" - so, I don't have much of a personal preference)

rjzamora (approved):

dask_expr/tests/test_groupby.py
phofl: Failing for var, the others just didn't push anything through

PR #222, opened 2023-07-11
Add npartitions and sort as arguments to from_pandas

phofl:


dask_expr/collection.py
```diff
@@ -871,10 +871,12 @@ def optimize(collection, fuse=True):
     return new_collection(expr.optimize(collection.expr, fuse=fuse))
 
 
-def from_pandas(data, *args, **kwargs):
+def from_pandas(data, npartitions=1, sort=True, **kwargs):
```
rjzamora, line 874: Do we need `kwargs` now that we are already exposing all "public" options explicitly.

rjzamora (approved):
Seems fine to me. Did you run into a bug/problem that motivated this? Something we should test?

Also, do we still want the `npartitions` and `sort` defaults to be defined in `FromPandas`?

dask_expr/collection.py
phofl: Probably not, removed

phofl:
I got annoyed that autocompletion does not work. We used the pattern do define defaults in both to avoid seeing them in pprint and friends

rjzamora:
>to avoid seeing them in pprint and friends

Ah, great point. Forgot about that.

PR #231, opened 2023-07-18
Implement sort=False for set_index

phofl:


rjzamora (commented):
Looks good. Note that I can resolve merge conflicts later if your busy.

**Side note (personal opinion)**:

This PR makes perfect sense, because it is consistent with `dask.dataframe`. With that said, I'm not a fan of the `sort=True` default. For small-data demo code, it's sometimes nice that we automatically sort by the new index. However, this behavior is **not** consistent with pandas, and I have seen the dramatic performance impact cause problems and confusion for many new users.

I'm not totally convinced that this is the right place or time to diverge with `dask.dataframe` convention. However, it does seem reasonable to expect a user to explicitly opt into the sorting behavior (or to use code like `df.set_index(...).sort_index()`).

phofl:
I mostly agree with your opinion, but want to suggest another alternative that we should at least think about:

We could be smart about actually sorting the Index or about calculating the divisions during optimization time. We can look at the expression tree to figure out whether or not sorting the Index would actually help. This has it's own set of complications, but would reduce the difference between dask/dask and us. That said I am generally more careful about breaking and changing stuff than others.

rjzamora:
>We can look at the expression tree to figure out whether or not sorting the Index would actually help

I would be very open to this, but agree it could get complicated.

>That said I am generally more careful about breaking and changing stuff than others.

I am also pretty careful about avoiding "breaking changes" in general. However, I also think we should take advantage of dask-expr's lack of API inertia while we can :)

dask_expr/_collection.py
```diff
@@ -828,6 +829,9 @@ def set_index(
         elif other == self.index.name:
             return self
 
+        if not sort:
+            return new_collection(SetIndexBlockwise(self.expr, other, drop, None))
+
```
rjzamora, line 834: I think I missed this detail from #230, but don't we need to satisfy `npartitions` when `sort=False` and when `sorted=True`?

phofl: I think the idea behind that was to have a fastpath. I’d rather raise if sort is false and partitions is given instead of doing a repartition

phofl: There is no performance benefit over doing an explicit repartition

rjzamora: Yeah, raising an error is fine with me, I just don't want to silently ignore the `npartitions` argument. Does that make sense?

phofl: Yep, added a check

dask_expr/_collection.py
```diff
@@ -828,6 +829,13 @@ def set_index(
         elif other == self.index.name:
             return self
 
+        if not sort:
+            if npartitions is not None:
+                raise ValueError(
+                    "Specifying npartitions with sort=False is not supported. Call `repartition` afterwards."
```
rjzamora, line 835: May also make sense to raise an error when `sorted` is `True` (below).

rjzamora (approved):
One more optional suggestion, but looks good. Thanks!

dask_expr/_collection.py
phofl: lets rather move it after sorted, makes a bit more sense, since performance is the same, but specifying sorted enables setting the divisions

rjzamora: Yeah, you can order things however you feel is best. I'm only noting that it is problematic (perhaps wrong) for the user to specify `npartitions` when `sort=False` **or** when `sorted=True` (since we ignore `npartitions` for both of these cases).

phofl: Ah sorry, misunderstood you earlier. Updated this now.

PR #232, opened 2023-07-19
Lower ``ApplyConcatApply`` into ``Chunk`` and ``TreeReduce``

rjzamora:
In preparation for shuffle-based groupby support, this PR breaks the general `ApplyConcatApply` expression into a blockwise `Chunk`, followed by a `TreeReduce`. The immediate result of this work is that the "chunk" component of a groupby aggregation can be fused with other `Blockwise` operations:

```python
import dask_expr as dx
import pandas as pd

pdf = pd.DataFrame({"x": list(range(10)) * 10, "y": range(100), "z": 1})
df = dx.from_pandas(pdf, npartitions=4)

agg = df.groupby("x").y.sum()
agg.optimize().pprint()
```
```
Sum(TreeReduce):  split_every=0
  Fused(32411):
  | Sum(Chunk):  chunk=<methodcaller: sum> columns=y by=['x'] numeric_only=False
  |   Projection: columns=['x', 'y']
  |     FromPandas: frame='<pandas>' npartitions=4
```

In the future, this change also makes it easier to implement a shuffle-based reduction for `split_out > 1`.

### DIFF
--- a/dask_expr/_expr.py
+++ b/dask_expr/_expr.py
@@ -2055,8 +2055,7 @@ def optimize_blockwise_fusion(expr):
             if v == set() or all(not isinstance(_expr, Blockwise) for _expr in v)
         ]
         while roots:
-            # Need to pop roots in FIFO order
-            root = roots.pop(0)
+            root = roots.pop()
             seen = set()
             stack = [root]
             group = []
@@ -2080,7 +2079,9 @@ def optimize_blockwise_fusion(expr):
                         # of partitions, since broadcasting within
                         # a group is not allowed.
                         stack.append(dep)
-                    elif dep not in roots and dependencies[dep]:
+                    elif dependencies[dep] and dep._name not in [
+                        r._name for r in roots
+                    ]:
                         # Couldn't fuse dep, but we may be able to
                         # use it as a new root on the next pass
                         roots.append(dep)
--- a/dask_expr/_expr.py
+++ b/dask_expr/_expr.py
@@ -2055,7 +2055,8 @@ def optimize_blockwise_fusion(expr):
             if v == set() or all(not isinstance(_expr, Blockwise) for _expr in v)
         ]
         while roots:
-            root = roots.pop()
+            # Need to pop roots in FIFO order
+            root = roots.pop(0)
             seen = set()
             stack = [root]
             group = []
--- a/dask_expr/_reductions.py
+++ b/dask_expr/_reductions.py
@@ -101,7 +101,7 @@ class ApplyConcatApply(Expr):
 class Chunk(Blockwise):
     """Partition-wise component of `ApplyConcatApply`
 
-    This class is used within `ApplyConcatApply._simplify_down`.
+    This class is used within `ApplyConcatApply._lower`.
 
     See Also
     --------
@@ -141,6 +141,15 @@ class Chunk(Blockwise):
 
 
 class TreeReduce(Expr):
+    """Tree-reduction component of `ApplyConcatApply`
+
+    This class is used within `ApplyConcatApply._lower`.
+
+    See Also
+    --------
+    ApplyConcatApply
+    """
+
     _parameters = [
         "frame",
         "kind",
--- a/dask_expr/_reductions.py
+++ b/dask_expr/_reductions.py
@@ -108,7 +108,7 @@ class Chunk(Blockwise):
     ApplyConcatApply
     """
 
-    _parameters = ["frame", "type", "chunk", "chunk_kwargs"]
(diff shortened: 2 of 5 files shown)


dask_expr/_expr.py
```diff
@@ -2079,7 +2079,9 @@ def _fusion_pass(expr):
                         # of partitions, since broadcasting within
                         # a group is not allowed.
                         stack.append(dep)
-                    elif dep not in roots and dependencies[dep]:
+                    elif dependencies[dep] and dep._name not in [
+                        r._name for r in roots
+                    ]:
```
rjzamora, line 2084: It turns out `dep not in roots` doesn't actually work the way we want it to, and so we were previously missing some opportunities for fusion.

**Reproducer**:
```python
import dask_expr as dx
import pandas as pd

pdf = pd.DataFrame({"x": list(range(10)) * 10, "y": range(100), "z": 1})
df = dx.from_pandas(pdf)

assert df.expr not in [df[["x"]].expr]
```

dask_expr/io/tests/test_io.py
```diff
@@ -62,9 +62,9 @@ def df_bc(fn):
         ),
     ],
 )
-def test_optimize(tmpdir, input, expected):
+def test_simplify(tmpdir, input, expected):
     fn = _make_file(tmpdir, format="parquet")
-    result = optimize(input(fn), fuse=False)
+    result = input(fn).simplify()
     assert str(result.expr) == str(expected(fn).expr)
```
rjzamora, line 68: These tests were only checking the behavior of `simpllify` anyway.

dask_expr/tests/test_collection.py
```diff
@@ -649,7 +649,7 @@ def test_tree_repr(df, fuse):
     assert str(df.seed) in s.lower()
     if fuse:
         assert "Fused" in s
-        assert s.count("|") == 9
+        assert s.count("|") == 10
```
rjzamora, line 652: We can also just test that `"|"` is in the string to avoid fragile checks like this.

dask_expr/_reductions.py
```diff
@@ -65,58 +85,140 @@ def _layer(self):
...
 
     def _divisions(self):
         return (None, None)
 
+    def __str__(self):
+        chunked = str(self.frame)
+        split_every = getattr(self, "split_every", 0)
+        return f"{type(self).__name__}({chunked}, kind={funcname(self.kind)}, split_every={split_every})"
+
+    def _tree_repr_lines(self, indent=0, recursive=True):
```
phofl, line 208: Does it make sense to add tests for both ``_tree_repr_lines``?

phofl (approved):
Small comments, looks good generally

dask_expr/_reductions.py
```diff
@@ -65,58 +85,140 @@ def _layer(self):
...
+        header = f"{funcname(self.kind)}({funcname(type(self))}): "
+        lines = []
+        if recursive:
+            for dep in self.dependencies():
+                lines.extend(dep._tree_repr_lines(2))
+
+        split_every = getattr(self, "split_every", 0)
+        header += f" split_every={split_every}"
```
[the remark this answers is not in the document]
rjzamora: Yeah, my thought was that the most important parameter for the `TreeReduce` component is `split_every`, because everything else will just match the `Chunk` dependency. I'm very open to other solutions here.

dask_expr/_reductions.py
rjzamora: Hmm. I do agree that we don't have the best coverage here, but I'm also a bit uncomfortable with the "fragility" of testing specific details of the `pprint` output, or private `_tree_repr_lines` results.

That said, I'll be happy to revise if you have something in mind.

rjzamora:
Merged so I can start looking into `split_out>1`. Feel free to comment here if anything needs to be revisited.

dask_expr/_reductions.py
phofl: I was more concerned that we don't seem to hit this at all (I might be wrong). Having a simple test that checks that this renders at all would be sufficient for now. But not urgent, this isn't critical

rjzamora: [test_tree_repr](https://github.com/phofl/dask-expr/blob/0f577c6829472f9c02a655998fcf64202ee27886/dask_expr/tests/test_collection.py#L578) should be checking that these methods are producing `"Sum(Chunk):"` and `"Sum(TreeReduce): split_every=0"` for the `Chunk` and `TreeReduce` cases, respectively.

Issue #233, opened 2023-07-20
AttributeError: 'Series' object has no attribute 'std'


skrawcz:
Is this expected behavior? 

Repro:

```python
import dask_expr as dx
df = dx.datasets.timeseries(dtypes={"name": "string", "id": int, "quantity": int, "spend": float})
df["quantity"].std()
```

**Stacktrace**: 
```python
Traceback (most recent call last):
  File "lib/python3.9/site-packages/dask_expr/_collection.py", line 175, in __getattr__
    val = getattr(self.expr, key)
  File "lib/python3.9/site-packages/dask_expr/_expr.py", line 148, in __getattr__
    raise err
  File "lib/python3.9/site-packages/dask_expr/_expr.py", line 137, in __getattr__
    return object.__getattribute__(self, key)
AttributeError: 'Projection' object has no attribute 'std'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "lib/python3.9/site-packages/dask_expr/_collection.py", line 181, in __getattr__
    raise err
  File "lib/python3.9/site-packages/dask_expr/_collection.py", line 170, in __getattr__
    return object.__getattribute__(self, key)
AttributeError: 'Series' object has no attribute 'std'
```

phofl:
Hi, thanks for your report and thanks for trying dask-expr!

``std`` isn't implemented yet since it goes through dask.array code that we don't support yet. So for now this is the expected behaviour. Sorry.

mrocklin:
Maybe we should implement the numerically unstable algorithm instead as a
stopgap?

skrawcz:
No worries. Is there a list of supported APIs somewhere?

rjzamora:
>No worries. Is there a list of supported APIs somewhere?

Good call. Now that people are using dask-expr, we definitely need this kind of documentation (I opened https://github.com/dask-contrib/dask-expr/issues/234 to track this)

skrawcz:
> > No worries. Is there a list of supported APIs somewhere?
> 
> Good call. Now that people are using dask-expr, we definitely need this kind of documentation (I opened #234 to track this)

Sweet, thanks. It'd be nice if the error experience could discern/guess the situation and point people to that list :)

phofl:
Pushed out a new release that contains std and var among other things.

PR #236, opened 2023-07-20
Add ``var`` and ``std`` implementations

rjzamora:
Closes https://github.com/dask-contrib/dask-expr/issues/233

Adds simple ``var`` and ``std`` implementations using Welford's online algorithm.


phofl (approved):
thx!

PR #237, opened 2023-07-21
Implement sort_values

phofl:
There is some duplication in there, will refactor in a follow up to keep this easier to review.

### DIFF
--- a/dask_expr/_collection.py
+++ b/dask_expr/_collection.py
@@ -2,8 +2,9 @@ from __future__ import annotations
 
 import functools
 import warnings
-from collections.abc import Hashable
+from collections.abc import Callable, Hashable, Mapping
 from numbers import Number
+from typing import Any, Literal
 
 import numpy as np
 import pandas as pd
@@ -39,7 +40,7 @@ from dask_expr._reductions import (
     ValueCounts,
 )
 from dask_expr._repartition import Repartition
-from dask_expr._shuffle import SetIndex, SetIndexBlockwise
+from dask_expr._shuffle import SetIndex, SetIndexBlockwise, SortValues
 from dask_expr._util import _convert_to_list
 
 #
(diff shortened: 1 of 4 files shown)


dask_expr/_collection.py
```diff
@@ -864,6 +865,54 @@ def set_index(
...
+    def sort_values(
+        self,
+        by: str | list[str],
+        npartitions: int | None = None,
+        ascending: bool | list[bool] = True,
+        na_position: Literal["first"] | Literal["last"] = "last",
+        partition_size: float = 128e6,
+        sort_function: Callable[[pd.DataFrame], pd.DataFrame] | None = None,
+        sort_function_kwargs: Mapping[str, Any] | None = None,
+    ):
```
rjzamora, line 877: I don't consider `upsample` to be a critical key-word argument, but was there a specific reason you left it out?

It looks like you would only need to pass it through to `_calculate_divisions`.

dask_expr/_shuffle.py
```diff
@@ -618,7 +618,25 @@ def operation(cls, df, index, name: str, npartitions: int):
         return df.assign(**{name: index})
 
 
-class SetIndex(Expr):
+class BaseSetIndexSortValues(Expr):
+    def _divisions(self):
+        if self.user_divisions is not None:
+            return self.user_divisions
+        divisions, mins, maxes, presorted = _calculate_divisions(
```
rjzamora, line 625: This function call can be expensive, and we are calling it in both `BaseSetIndexSortValues._divisions` and `SortValues._lower`. I know you are planning on doing some refactoring in a follow-up, but I would consider it a high priority to leverage some kind of caching here.

rjzamora: **Another Note**: `_calculate_divisions` also requires the eager execution of a task graph, but doesn't leave us an opportunity to apply fusion first. We probably want to find a way to allow the user to configure the optimization options.

dask_expr/_shuffle.py
```diff
@@ -618,7 +618,25 @@ def operation(cls, df, index, name: str, npartitions: int):
...
+        )
+        if presorted:
+            divisions = mins.copy() + [maxes[-1]]
+        return divisions
+
+    @property
+    def npartitions(self):
+        if self.operand("npartitions") is not None:
+            return self.operand("npartitions")
+        return self.frame.npartitions
```
rjzamora, line 636:
```suggestion
        return self.operand("npartitions") or self.frame.npartitions
```
Minor preference, but happy either way.

dask_expr/_shuffle.py
```diff
@@ -716,6 +734,80 @@ def _simplify_up(self, parent):
...
+            sort_kwargs.update(self.operand("sort_function_kwargs"))
+        return sort_kwargs
+
+    @property
+    def _meta(self):
+        return self.frame._meta
+
+    def _lower(self):
+        by = self.frame[self.by[0]]
+        divisions, _, _, presorted = _calculate_divisions(
```
rjzamora, line 780: As noted above: We may have already executed this calculation earlier (if we happend to query `divisions`), so it would be nice to avoid another `compute` call here.

dask_expr/_shuffle.py
```diff
@@ -716,6 +734,80 @@ def _simplify_up(self, parent):
...
+            npartitions_out=len(divisions) - 1,
+            ignore_index=True,
+        )
+        return SortValuesBlockwise(
+            shuffled, self.sort_function, self.sort_function_kwargs
+        )
+
+    def _simplify_up(self, parent):
+        if isinstance(parent, Projection):
+            columns = parent.columns + self.by
```
rjzamora, line 802: Couldn't this include the `by` column twice?

rjzamora (commented):
I'm excited to see this!  I left a few initial thoughts/questions.

dask_expr/_shuffle.py
phofl: The function is already cached. This shouldn’t be executed more than once 

yeah I agree this needs some refactoring. I want to avoid executing this in general it it’s not necessary when looking at the expression tree

dask_expr/_collection.py
phofl: Yeah I left it out of set_index as well, will do as a follow up. Should probably have done this before doing sort values

dask_expr/_shuffle.py
phofl: Cached, see above

dask_expr/_shuffle.py
phofl: Oh fair, will add a test

dask_expr/_shuffle.py
rjzamora: >_calculate_division

Ah okay, I didn't see the `lru_cache`. I look forward to the refactor, but agree we can do it in a follow-up as long as we are caching.

phofl: Yeah caching on the class level didn't really work, that's why it's in a method

dask_expr/_shuffle.py
phofl: Updated

rjzamora (approved):

PR #239, opened 2023-07-24
Adapt new variable name from dask

phofl:


phofl:
Merging to get ci running again

phofl:
@rjzamora things got tricky since we started releasing. Users won't be able to use us now until the next Dask release is out. maybe we have to start special casing now. Thoughts?

rjzamora:
>Users won't be able to use us now until the next Dask release is out. maybe we have to start special casing now. Thoughts?

Yeah, I suppose we have two main options:

1. **We support both `main` and the latest official release.** After each release we will need to submit/merge PRs to cleanup any temporary "band-aids" that were added before the previous release.
2. **We only support `main`, and time our releases to always match with `dask` releases.** Any user interested in working with the `dask-expr:main` will also need to install `dask:main`/`distributed:main` from source.
 
I can think of obvious advantages/disadvantages either way, so I don't really have a preference.

phofl:
I think I'd prefer 1, I would like to keep the option to release more often than dask in the beginning

rjzamora:
>I think I'd prefer 1, I would like to keep the option to release more often than dask in the beginning

That works for me. Then perhaps we should:

1. Add a new **"band-aid"** label so that we can raise clear issues to track temporary changes that should be cleaned up after each release.
2. Start running CI on main **and** on the latest release.

PR #241, opened 2023-07-24
Implement concat axis=1 for unknown divisions

phofl:


dask_expr/_concat.py
```diff
@@ -114,7 +147,14 @@ def _simplify_up(self, parent):
                 for frame, cols in zip(self._frames, columns_frame)
             ]
             return type(parent)(
```
rjzamora, line 149: Seems like we are now much more likely to get empty column projections in a subset of the concatenated `DataFrame` objects. I *think* this will work just fine, but it might be good to explicitly drop items for the list in that case?

phofl: Yep makes sense, added

dask_expr/tests/test_concat.py
```diff
@@ -95,3 +95,33 @@ def test_concat_simplify_projection_not_added(pdf, df):
...
     assert_eq(q, pd.concat([pdf, pdf2])[["y", "x"]])
+
+
+def test_concat_axis_one_co_aligned(pdf, df):
+    df2 = df.rename({"x": "x_2", "y": "y_2"})
+    pdf2 = pdf.add_suffix("_2")
+    assert_eq(concat([df, df2], axis=1), pd.concat([pdf, pdf2], axis=1))
+
+
+def test_concat_axis_one_all_divisions_unknown(pdf, df):
```
rjzamora, line 106:
```suggestion
def test_concat_axis_one_all_divisions_unknown(pdf):
```
Looks like you aren't using the `df` fixture here.

phofl:
Thx for the review

PR #242, opened 2023-07-24
Implement add_suffix, add_prefix

phofl:


README.md
```diff
@@ -100,6 +100,8 @@ API Coverage
 - `partitions`
 - `prod`
 - `rename`
+- `add_prefix`
+- `add_sufix`
```
rjzamora, line 104: Could we keep this list in alphabetical order for now? :)

rjzamora (commented):
Just a suggestion for the `README`. Otherwise, looks good.

**Optional**: Doesn't need to be in this PR, but we probably want `_simplify_up` logic that is similar to `RenameFrame`. For both these cases, we could probably also push down filters as well.

README.md
phofl: Changed

phofl:
Added simplify steps as well

I'll have to look into filters more broadly for Blockwise stuff. So will keep this out of here

rjzamora (approved):
Nice. Thanks!

PR #243, opened 2023-07-24
Fix ``Binop`` column projection logic

rjzamora:
We currently push a `Projection` operation down to both operands to a `Binop` expression. However, there is no guarantee that both operands can be indexed with a `getitem` operation. This PR adds an `ndim` check for each operand.


phofl (approved):
thx!

PR #245, opened 2023-07-26
Implement simplify_down for Filter

phofl:
So this caused me more troubles than I'd like to admit. I think I found an issue with the ``combine_similar`` logic, it's too local.

If we end up pushing more than one operation through in simplify, we can not capture this with ``combine_similar``. Small reproducer (dropna blocks everything):

```
pdf = pd.DataFrame({"x": [1, 2, 3, 4, 5, 6], "y": [4, 5, 8, 6, 1, 4]})
df = from_pandas(pdf, npartitions=3)

df = df.dropna().replace(1, 5)
df = df[df.x > 3][["x", "y"]]
df.optimize(fuse=False)
```

We'll end up with the following operation after simplify:

```
df = df.dropna()[["x", "y"]][df.dropna()["x"].replace(1, 5) > 3].replace(1, 5)
```

``combine_similar`` is only looking one operation ahead, which means that it won't be able to capture that both branches have the same root with ``dropna``. Ideally, we would like to restructure this to:

```
df = df.dropna().replace(1, 5)
df[["x", "y"]][df.x > 3]
```

cc @rjzamora Thoughts here?

### DIFF
--- a/dask_expr/_expr.py
+++ b/dask_expr/_expr.py
@@ -952,6 +952,7 @@ class Blockwise(Expr):
     operation = None
     _keyword_only = []
     _projection_passthrough = False
+    _filter_passthrough = False
 
     @property
     def _required_attribute(self):
@@ -1044,7 +1045,12 @@ class Blockwise(Expr):
     def _combine_similar(self, root: Expr):
         # Push projections back up through `_projection_passthrough`
         # operations if it reduces the number of unique expression nodes.
-        if self._projection_passthrough and isinstance(self.frame, Projection):
+        if (
+            self._projection_passthrough
+            and isinstance(self.frame, Projection)
+            or self._filter_passthrough
+            and isinstance(self.frame, Filter)
+        ):
             try:
                 common = type(self)(self.frame.frame, *self.operands[1:])
             except ValueError:
@@ -1052,19 +1058,20 @@ class Blockwise(Expr):
                 # (There is no guarentee that the same method will exist for
                 # both a Series and DataFrame)
                 return None
-            projection = self.frame.operand("columns")
-            push_up_projection = False
+            operation = self.frame.operands[1]
+            push_up_op = False
             for op in self._find_similar_operations(root, ignore=self._parameters):
                 if (
-                    isinstance(op.frame, Projection)
+                    isinstance(op.frame, (Projection, Filter))
                     and (
                         common._name == type(op)(op.frame.frame, *op.operands[1:])._name
                     )
                 ) or common._name == op._name:
-                    push_up_projection = True
+                    push_up_op = True
+                    break
 
-            if push_up_projection:
-                return common[projection]
+            if push_up_op:
+                return common[operation]
         return None
 
 
@@ -1210,27 +1217,6 @@ class CombineFirst(Blockwise):
             )
 
 
-class RenameFrame(Blockwise):
-    _parameters = ["frame", "columns"]
-    _keyword_only = ["columns"]
-    operation = M.rename
-
-    def _simplify_up(self, parent):
-        if isinstance(parent, Projection) and isinstance(
-            self.operand("columns"), Mapping
-        ):
-            reverse_mapping = {val: key for key, val in self.operand("columns").items()}
-            if is_series_like(parent._meta):
-                # Fill this out when Series.rename is implemented
-                return
-            else:
-                columns = [
-                    reverse_mapping[col] if col in reverse_mapping else col
-                    for col in parent.columns
-                ]
-            return type(self)(self.frame[columns], *self.operands[1:])
-
-
 class Sample(Blockwise):
     _parameters = ["frame", "state_data", "frac", "replace"]
     operation = staticmethod(methods.sample)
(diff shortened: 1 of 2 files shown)


dask_expr/_expr.py
```diff
@@ -1044,27 +1045,33 @@ def _simplify_up(self, parent):
...
         # operations if it reduces the number of unique expression nodes.
-        if self._projection_passthrough and isinstance(self.frame, Projection):
+        if (
+            self._projection_passthrough
+            and isinstance(self.frame, Projection)
+            or self._filter_passthrough
+            and isinstance(self.frame, Filter)
+        ):
             try:
                 common = type(self)(self.frame.frame, *self.operands[1:])
```
phofl, line 1055: This could look further than ``self.frame.frame`` if ``self.frame.frame`` is either a Projection or a Filter, but this would make the logic after here more complicated. Open to other suggestions.

phofl: One possible implementation:

```
    def _combine_similar(self, root: Expr):
        # Push projections back up through `_projection_passthrough`
        # operations if it reduces the number of unique expression nodes.
        if (
            self._projection_passthrough
            and isinstance(self.frame, Projection)
            or self._filter_passthrough
            and isinstance(self.frame, Filter)
        ):
            frame, operations = self.frame, []
            while isinstance(frame, (Filter, Projection)):
                operations.append(frame.operands[1])
                frame = frame.frame
            else:
                try:
                    common = type(self)(frame, *self.operands[1:])
                except ValueError:
                    # May have encountered a problem with `_required_attribute`.
                    # (There is no guarentee that the same method will exist for
                    # both a Series and DataFrame)
                    return None
            push_up_op = False
            for op in self._find_similar_operations(root, ignore=self._parameters):
                if (
                    isinstance(op.frame, (Projection, Filter))
                    and (
                        common._name == type(op)(op.frame.frame, *op.operands[1:])._name
                    )
                ) or common._name == op._name:
                    push_up_op = True
                    break

            if push_up_op:
                for op in reversed(operations):
                    common = common[op]
                return common
        return None
```

rjzamora (commented):
I'll spend some time thinking through this, but I'm not sure I understand the motivating problem here. As far as I can tell, the current optimization pass already produces what we need:

```python
import dask_expr as dx
import pandas as pd

pdf = pd.DataFrame({"x": [1, 2, 3, 4, 5, 6], "y": [4, 5, 8, 6, 1, 4]})
df = dx.from_pandas(pdf, npartitions=3)

df = df.dropna().replace(1, 5)
df = df[df.x > 3][["x", "y"]]
df.simplify().pprint()
```
```
Projection: columns=['x', 'y']
  Filter:
    Replace: to_replace=1 value=5
      DropnaFrame:
        FromPandas: frame='<pandas>' npartitions=3
    GT: right=3
      Projection: columns='x'
        Replace: to_replace=1 value=5
          DropnaFrame:
            FromPandas: frame='<pandas>' npartitions=3
```
```python
df.optimize().pprint()
```
```
Fused(c8122):
| Filter:
|   GT: right=3
|     Projection: columns='x'
|       Replace: to_replace=1 value=5
|         DropnaFrame:
|           FromPandas: frame='<pandas>' npartitions=3
```

There is a good chance I'm wrong and just misunderstanding. (sorry if that's the case)

rjzamora:
>Yes you have to checkout this branch, sorry for not mentioning

Right, sorry for missing that! Makes sense.

dask_expr/_expr.py
rjzamora: I'm pretty sure I like [your proposed alternative](https://github.com/dask-contrib/dask-expr/pull/245/files#r1274852097). Since it is *slightly* more complicated, can you add a brief comment on "why" we are looping until we get to the end of the `Filter`/`Projection` stack? (just to help other follow the logic).

One point I don't understand (yet) is why you push up the operations in `reversed` order?

phofl: Updated

dask_expr/_expr.py
```diff
@@ -1542,6 +1564,8 @@ def _simplify_up(self, parent):
 class Projection(Elemwise):
     """Column Selection"""
 
+    _filter_passthrough = False
+
     _parameters = ["frame", "columns"]
     operation = operator.getitem
```
rjzamora, line 1570:
```suggestion
    _parameters = ["frame", "columns"]
    operation = operator.getitem
    _filter_passthrough = False
```

rjzamora: very minor suggestion

dask_expr/_expr.py
```diff
@@ -1285,7 +1265,35 @@ class Elemwise(Blockwise):
...
+            reverse_mapping = {val: key for key, val in self.operand("columns").items()}
+            if is_series_like(parent._meta):
+                # Fill this out when Series.rename is implemented
+                return
+            else:
+                columns = [
+                    reverse_mapping[col] if col in reverse_mapping else col
+                    for col in parent.columns
+                ]
+            return type(self)(self.frame[columns], *self.operands[1:])
```
rjzamora, line 1296: Note that you can also do something like `return self.substitute_parameters({"frame": self.frame[columns]})` in cases like this (which is maybe more readable?).

rjzamora (approved):
Only some minor suggestions, but I like this. Thanks @phofl 

As a follow-up, it would be good to add the necessary "filter passthrough" to `DropnaFrame` (and to figure out how to "push back up" projections that are not absorbed by the IO layer). I'd like to I'll probably be thinking through this today.

PR #247, opened 2023-07-28
Absorb column projections in other ``BlockwiseIO`` expressions

rjzamora:
Follow up to #245, where I noticed that we were not quite capturing optimal column projection behavior for expressions originating from `FromPandas`. More specifically, since `FromPandas` does not "absorb" a `Projection` into it's own `"columns"` operand, it becomes difficult to produce a "combined" column projection at the optimal position in the expression graph.

For example:

```python
import dask_expr as dx
import pandas as pd

pdf = pd.DataFrame({"x": [1, 2, 3, 4, 5, 6], "y": [4, 5, 8, 6, 1, 4], "z": 1})
df = dx.from_pandas(pdf, npartitions=3)

df = df.dropna().replace(1, 5)
df = df[df.x > 3][["x", "y"]]

df.optimize(fuse=False).pprint()
```

**Before this PR**:

In `main`, we currently perform `DropnaFrame` and `Replace` on **all** columns of `pdf`, even though we know that we can drop column `"z"` immediately.

```
Filter:
  Projection: columns=['x', 'y']
    Replace: to_replace=1 value=5
      DropnaFrame:
        FromPandas: frame='<pandas>' npartitions=3
  GT: right=3
    Projection: columns='x'
      Replace: to_replace=1 value=5
        DropnaFrame:
          FromPandas: frame='<pandas>' npartitions=3
```

**After this PR**:

By allowing `FromPandas` (and `ReadCSV`) to absorb column projections, we can produce much more optimal behavior.

```
Filter:
  Replace: to_replace=1 value=5
    DropnaFrame:
      FromPandas: frame='<pandas>' npartitions=3 columns=['x', 'y']
  GT: right=3
    Projection: columns='x'
      Replace: to_replace=1 value=5
        DropnaFrame:
          FromPandas: frame='<pandas>' npartitions=3 columns=['x', 'y']
```

### DIFF
--- a/dask_expr/_expr.py
+++ b/dask_expr/_expr.py
@@ -28,7 +28,7 @@ from dask.dataframe.utils import clear_known_categories, drop_by_shallow_copy
 from dask.utils import M, apply, funcname, import_required, is_arraylike
 from tlz import merge_sorted, unique
 
-from dask_expr._util import _tokenize_deterministic
+from dask_expr._util import _tokenize_deterministic, _tokenize_partial
 
 replacement_rules = []
 
@@ -768,20 +768,9 @@ class Expr:
             # No other operations of the same type. Early return
             return []
 
-        def _tokenize(rp):
-            # Helper function to "tokenize" the operands
-            # that are not in the `ignore` list
-            return _tokenize_deterministic(
-                *[
-                    op
-                    for i, op in enumerate(rp.operands)
-                    if i >= len(rp._parameters) or rp._parameters[i] not in ignore
-                ]
-            )
-
         # Return subset of `alike` with the same "token"
-        token = _tokenize(self)
-        return [item for item in alike if _tokenize(item) == token]
+        token = _tokenize_partial(self, ignore)
+        return [item for item in alike if _tokenize_partial(item, ignore) == token]
 
     def _node_label_args(self):
         """Operands to include in the node label by `visualize`"""
@@ -2148,7 +2137,12 @@ def are_co_aligned(*exprs):
     """Do inputs come from different parents, modulo blockwise?"""
     exprs = [expr for expr in exprs if not is_broadcastable(expr)]
     ancestors = [set(non_blockwise_ancestors(e)) for e in exprs]
-    return len(set(flatten(ancestors, container=set))) == 1
+    unique_ancestors = {
+        # Account for column projection within IO expressions
+        _tokenize_partial(item, ["columns", "_series"])
+        for item in flatten(ancestors, container=set)
+    }
+    return len(unique_ancestors) == 1
 
 
 ## Utilites for Expr fusion
--- a/dask_expr/_expr.py
+++ b/dask_expr/_expr.py
@@ -1077,6 +1077,7 @@ class Blockwise(Expr):
                 for i, op in enumerate(reversed(operations)):
                     common = common[op]
                     if i > 0:
+                        # Combine stacked projections
                         common = common._simplify_down() or common
                 return common
         return None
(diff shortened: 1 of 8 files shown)


dask_expr/io/io.py
```diff
@@ -39,19 +48,109 @@ def _layer(self):
 
 
 class BlockwiseIO(Blockwise, IO):
-    pass
+    _absorb_projections = False
```
rjzamora, line 51: Note that it probably makes sense to work with a `BlockwiseIO` subclass here (instead of adding `_absorb_projections`). However, I used the attribute to explore and haven't bothered to modify the approach yet. I'll be happy to revise.

dask_expr/_expr.py
```diff
@@ -1202,6 +1198,14 @@ def _simplify_up(self, parent):
                 type(self)(self.frame[sorted(columns)], *self.operands[1:]),
                 *parent.operands[1:],
             )
+        elif isinstance(parent, Projection):
```
phofl, line 1201: We can't do this. We drop rows if any column contains a missing value. The operation will change if we remove columns that could potentially contain NA values

dask_expr/_expr.py
```diff
@@ -1191,6 +1183,10 @@ class DropnaFrame(Blockwise):
     _keyword_only = ["how", "subset", "thresh"]
     operation = M.dropna
 
+    @property
+    def _projection_passthrough(self):
```
phofl, line 1187: Same here.

dask_expr/io/csv.py
```diff
@@ -1,16 +1,27 @@
...
+    ]
     _defaults = {
-        "usecols": None,
+        "columns": None,
         "header": "infer",
         "_partitions": None,
         "storage_options": None,
+        "_series": False,
     }
+    _absorb_projections = False
```
phofl, line 24: This is a TODO for the future?

dask_expr/io/io.py
phofl: I think that's fine for now

dask_expr/_expr.py
```diff
@@ -1615,6 +1619,9 @@ def _simplify_down(self):
         if (
             str(self.frame.columns) == str(self.columns)
             and self._meta.ndim == self.frame._meta.ndim
+            and not (
```
phofl, line 1622: I am not sure I understand why this is necessary. Can you elaborate?

Why can't we return self.frame if the columns are the same anyway?

phofl (commented):
Some comments

phofl:
@rjzamora I've removed the dropna changes, but the rest is good as is for now. We can address my remaining  comments after you are back

dask_expr/_expr.py
rjzamora: Ah, Good catch! I don't know what I was thinking :)

dask_expr/_expr.py
rjzamora: My original thinking here was that if `self.frame` is a `BlockwiseIO` expression that can "absorb" projections, then we want to make sure `self.frame` actually absorbs the projection (by applying its own `simplify_up` logic). 

With that said, you are correct that "absorbing" the projection should not really change anything if the first two criteria of this `if` statement are `True`. Therefore, we probably can/should revert this change.  Note that I submitted https://github.com/dask-contrib/dask-expr/pull/267 to do this (where I included a necessary bug fix).

dask_expr/io/csv.py
rjzamora: Yeah, forgot about this. Just submitted https://github.com/dask-contrib/dask-expr/pull/268

Issue #159, opened 2023-06-19
Benchmark failures - bugs/missing operations


phofl:
- [x] GroupBy.value_counts
- [ ] iloc
- [x] select_dtypes
- [x] Repartition misses keyword-argument ``partition_size``
- [ ] GroupBy.apply
- [x] ``set_index``
- [x] grouping by a Series object
- [ ] grouping by an Index object
- [x] ``assign`` with callable
- [ ] ``GroupBy.agg`` ``shuffle`` keyword
- [x] missing align step in add/sub, ...
- [x] missing keywords in ``read_csv``
- [x] "p2p" shuffle

Not planning on fixing all of them immediately, just collecting failures and fixing what was on my agenda anyway or is a very small fix.

Reference build: https://github.com/coiled/benchmarks/actions/runs/5310108176/jobs/9611644148?pr=837

phofl:
@mrocklin thoughts on ``partition_size`` in ``repartition``? This triggers an immediate computation in dask/dask, but I think we could reasonably defer this to compute time

mrocklin:
I think that anything we can defer we should defer.  

I think that for this value in particular we'll likely have to compute it if anything asks us for divisions.

phofl:
closing, those all work

PR #249, opened 2023-07-31
Implement repartition_size

phofl:
xref #159

Went down a bit of a rabbit hole, but I don't think that this is doable without computing the npartitions with our setup right now


mrocklin (approved):
Not reviewing the technical bits here, but it's well isolated enough.  From my perspective feel free to merge.

PR #250, opened 2023-08-01
Optimize len for other methods

phofl:
None of those actually change the shape. Ran into this while doing some tests


mrocklin:
Fine by me.  

An alternative direction (not pushing for this) would be that we start including predicates on classes like the following:

```python
class Repartition(Expr):
    is_length_preserving = True
```

And then `Len` would just check this attribute.  This would be a bit more extensible (classes can choose to publish this fact without Len having to know about the class) but does muddy up the namespace a bit.

Again, not pushing for this.  I think that it's the kind of thing that could make sense if there are a few other predicates that would all make sense.  

Thoughts?

phofl:
I thought a little bit about this. Extensibility is what convinced me in the end. If users bring their own Expression, having a predicate enables them to specify their own things that will get taken into account by other expressions.

mrocklin:
I tried resolving merge conflicts with the GitHub UI but that didn't work.  In general I'm +1.  Feel free to merge as you like.  (although hopefully after fixing the linting issue, sorry about that).

PR #254, opened 2023-08-02
Fix combine_similar for read_parquet if one branch does not project columns

phofl:
…olumns


phofl:
Merging this one, want to run the benchmarks again.

cc @rjzamora when you get back

rjzamora (commented):
Ah, makes sense. Thanks for fixing this @phofl

PR #255, opened 2023-08-02
Avoid huge graphs in make_timeseries

phofl:
We can't create the state_data on the client, this results in huge_graphs (one of the reasons that the coiled benchmarks are failing right now)

### DIFF
--- a/dask_expr/datasets.py
+++ b/dask_expr/datasets.py
@@ -33,7 +33,7 @@ class Timeseries(PartitionsFiltered, BlockwiseIO):
         "_partitions": None,
     }
 
-    @property
+    @functools.cached_property
     def _meta(self):
         dtypes = self.operand("dtypes")
         states = [0] * len(dtypes)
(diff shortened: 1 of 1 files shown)


dask_expr/datasets.py
```diff
@@ -51,7 +51,7 @@ def random_state(self):
             k: (
                 np.random.randint(2e9, size=npartitions)
                 if self.seed is None
-                else random_state_data(npartitions, self.seed)
+                else [self.seed] * npartitions
```
mrocklin, line 54: Does this mean that every partition gets the same seed and might generate the same data?

phofl: Oh, I missunderstood something, sorry. Thx!

Updated to avoid serializing all of them but keep the behaviour..

phofl:
Same dtypes have the same values (as in dask/dask) but different partitions have different values. Going ahead and merge to get a Benchmark run. Happy to adjust in a follow up If there is a better way for this

PR #257, opened 2023-08-04
Avoid holding onto cached references for set_index

phofl:
This caused me some headaches


phofl:
We are currently duplicating the DataFrame, so I'll go ahead and merge this so that I can release it today

PR #258, opened 2023-08-04
Fix empty timeseries creation

phofl:


phofl:
We didn't have test coverage for this when I changed it initially

PR #260, opened 2023-08-07
Implement combine_similar for timeseries

phofl:
@rjzamora I'd appreciate thoughts here when you get back

This isn't super important for users, but not having combine_similar here makes most of our benchmarks meaningless since we'll end up doing a lot of unnecessary work.

### DIFF
--- a/dask_expr/datasets.py
+++ b/dask_expr/datasets.py
@@ -283,6 +283,8 @@ def timeseries(
     if seed is None:
         seed = np.random.randint(2e9)
 
+    # Add a token for dtypes to be able to identify operations that where the
+    # same before going through optimizations (needed in combine_similar)
     token_dtypes = _tokenize_deterministic(dtypes)
 
     expr = Timeseries(
--- a/dask_expr/datasets.py
+++ b/dask_expr/datasets.py
@@ -3,11 +3,10 @@ import operator
 
 import numpy as np
 import pandas as pd
-from dask.utils import is_dataframe_like, random_state_data
+from dask.utils import random_state_data
 
 from dask_expr._collection import new_collection
-from dask_expr._expr import Projection
-from dask_expr._util import _tokenize_deterministic, is_scalar
+from dask_expr._util import _tokenize_deterministic
 from dask_expr.io import BlockwiseIO, PartitionsFiltered
 
 __all__ = ["timeseries"]
@@ -24,10 +23,10 @@ class Timeseries(PartitionsFiltered, BlockwiseIO):
         "partition_freq",
         "seed",
         "kwargs",
+        "columns",
         "_partitions",
         "_token_dtypes",
         "_series",
-        "columns",
     ]
     _defaults = {
         "start": "2000-01-01",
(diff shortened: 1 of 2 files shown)


rjzamora:
Planning to look carefully at this today. My initial question is "can we use `_absorb_projections = True` to avoid adding new `_combine_similar` logic?". Perhaps `Timeseries` is somehow different from the other `BlockwiseIO` cases?

rjzamora:
>Timeseries changes dtypes instead of columns if a projection is pushed down

Yeah, that makes sense. We do want the user to pass `dtypes` to `timeseries` (and not specify `columns`). However, I don't really see the harm in (privately) supporting `columns` within `Timeseries`. In other words, we can probably avoid code duplication if we just add a `dtypes` property that simply checks `self.operand("columns")` to apply a projection.

phofl:
Yep you are correct, that makes more sense. I've adjusted this pr accordingly

dask_expr/datasets.py
```diff
@@ -53,42 +70,45 @@ def random_state(self):
...
-        if isinstance(parent, Projection) and len(self.dtypes) > 1:
-            dtypes = {col: self.operand("dtypes")[col] for col in parent.columns}
-            out = Timeseries(
+        if isinstance(parent, Projection) and is_dataframe_like(self._meta):
+            make_series = is_scalar(parent.operand("columns")) and not self._series
+            return Timeseries(
                 self.start,
                 self.end,
-                dtypes=dtypes,
+                dtypes=self.operand("dtypes"),
```
rjzamora, line 102: Now that we aren't changing `dtypes` here, then we can remove `Timeseries._simplify_up` altogether. We also don't need the `_token_dtypes` operand anymore.

dask_expr/datasets.py
```diff
@@ -31,19 +38,29 @@ class Timeseries(PartitionsFiltered, BlockwiseIO):
...
+            return result[result.columns[0]]
+        return result
 
     def _divisions(self):
         return pd.date_range(start=self.start, end=self.end, freq=self.partition_freq)
 
+    @property
+    def _dtypes(self):
+        dtypes = self.operand("dtypes")
+        return {col: dtypes[col] for col in self.operand("columns")}
```
rjzamora, line 62: What if `self.operand("columns")` is `None`?

dask_expr/datasets.py
phofl: Yeah that's fair.

We still need the token. We have to be able to distinguish between different time series calls

```
df = timeseries()
df2 = timeseries()

(df + df2) != 2*df
```

dask_expr/datasets.py
phofl: Good point, should not happen. Adjusted the signature

dask_expr/datasets.py
rjzamora: >We have to be able to distinguish between different time series calls

Wont the seeds be different (making them distinguishable anyway)?

phofl: I have another PR to remove the seed, specifying a seed makes serialising the graph very slow (should probably optimise this, but didn't have time yet) and is inconsistent with dask/dask, which is what makes benchmarking hard

rjzamora (approved):

dask_expr/datasets.py
phofl: We should be good here though, the randint call is only triggered when constructing the graph, so not when we go through optimize

PR #262, opened 2023-08-07
Pass split_every through to TreeReduce

### LINKED ISSUE #261: Aggregations in groupby are broken
They somehow pull a lot of data onto one worker. 

test_q3 from coiled benchmarks is a reproducer, simpler:

```
uri = "s3://coiled-datasets/h2o-benchmark/N_1e8_K_1e2_parquet/*.parquet"
ddf = read_parquet(uri, engine="pyarrow", storage_options={"anon": True})
ddf = ddf[["id3", "v1", "v3"]]
(
    ddf.groupby("id3", dropna=False, observed=True)[["v1"]].sum()
    #.agg({"v1": "sum", "v3": "mean"})
    .compute()
)
```

column selection and aggregation works but takes forever, using agg kills the worker. 30 workers (which is way more than we'd need normally)

phofl:
This was just not passed through

closes #261

### DIFF
--- a/dask_expr/_reductions.py
+++ b/dask_expr/_reductions.py
@@ -51,7 +51,7 @@ class ApplyConcatApply(Expr):
         # This is an abstract expression
         raise NotImplementedError()
 
-    @property
+    @functools.cached_property
     def _meta(self):
         meta = meta_nonempty(self.frame._meta)
         meta = self.chunk(meta, **self.chunk_kwargs)
@@ -98,6 +98,7 @@ class ApplyConcatApply(Expr):
             aggregate,
             combine_kwargs,
             aggregate_kwargs,
+            split_every=getattr(self, "split_every", 0),
         )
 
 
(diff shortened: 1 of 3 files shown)


dask_expr/_reductions.py
```diff
@@ -161,7 +162,9 @@ class TreeReduce(Expr):
         "aggregate",
         "combine_kwargs",
         "aggregate_kwargs",
+        "split_every",
     ]
+    _defaults = {"split_every": 0}
```
fjetter, line 167: `0` appears to be odd. dask apply_concat_apply is setting this to 8 and is raising if this is smaller than 2

https://github.com/dask/dask/blob/bdb827b7e2d0183449b967e9e9941703cdb711c5/dask/dataframe/core.py#L6734-L6738

I am probably missing something here, am I not? is this a sentinel for no splits? Is 1 the same then?

phofl: Yep 0 means no splits. 

The regular reductions have split_every=False in dask/dask, which has the same meaning as 0 here. 
This pr sets the default to 8 for groupby reductions, we should be consistent with dask/dask after this PR is in

fjetter (approved):

PR #263, opened 2023-08-08
Remove automatic seed value

phofl:
Having a seed still creates a relatively large graph which makes benchmarks against dask/dask partially meaningless

### DIFF
--- a/dask_expr/datasets.py
+++ b/dask_expr/datasets.py
@@ -221,8 +221,5 @@ def timeseries(
     if dtypes is None:
         dtypes = {"name": "string", "id": int, "x": float, "y": float}
 
-    if seed is None:
-        seed = np.random.randint(2e9)
-
     expr = Timeseries(start, end, dtypes, freq, partition_freq, seed, kwargs)
     return new_collection(expr)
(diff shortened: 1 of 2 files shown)


dask_expr/datasets.py
```diff
@@ -221,8 +221,5 @@ def timeseries(
     if dtypes is None:
         dtypes = {"name": "string", "id": int, "x": float, "y": float}
 
-    if seed is None:
-        seed = np.random.randint(2e9)
```
rjzamora, line 225: It seems like this we are somehow loosing some determinism when we remove these lines. For example, I get failures for the following test when these lines are removed:

```python
@pytest.mark.parametrize("seed", [42, None])
def test_timeseries_deterministic_head(seed):
    df = timeseries(end="2000-01-02", seed=seed)
    assert_eq(df.head(), df.head())
```

rjzamora:
While reviewing this PR, I started to get the sense that the original `Timeseries.random_state` logic/usage is simply "wrong". Rather than avoiding a default random seed, let me know what you think of https://github.com/dask-contrib/dask-expr/pull/270 as an alternative fix.

phofl:
Yeah I was planning to look into this as well, closing here

PR #265, opened 2023-08-08
Add more efficient length computations to merge and concat

phofl:
Found these in benchmarks

### DIFF
--- a/dask_expr/_merge.py
+++ b/dask_expr/_merge.py
@@ -166,6 +166,7 @@ class Merge(Expr):
             # Reorder the column projection to
             # occur before the Merge
             if isinstance(parent, Index):
+                # Index creates an empty column projection
                 projection, parent_columns = [], None
             else:
                 projection, parent_columns = parent.operand("columns"), parent.operand(
--- a/dask_expr/_merge.py
+++ b/dask_expr/_merge.py
@@ -3,7 +3,7 @@ import functools
 from dask.dataframe.dispatch import make_meta, meta_nonempty
 from dask.utils import M, apply
 
-from dask_expr._expr import Blockwise, Expr, Projection
+from dask_expr._expr import Blockwise, Expr, Index, Projection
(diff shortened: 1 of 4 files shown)


dask_expr/_merge.py
```diff
@@ -162,12 +162,17 @@ def _lower(self):
...
     def _simplify_up(self, parent):
-        if isinstance(parent, Projection):
+        if isinstance(parent, (Projection, Index)):
             # Reorder the column projection to
             # occur before the Merge
-            projection = parent.operand("columns")
-            if isinstance(projection, (str, int)):
-                projection = [projection]
+            if isinstance(parent, Index):
+                projection, parent_columns = [], None
```
rjzamora, line 169: Just to make it easier for others to understand this code in the future, maybe we can we add a comment to clarify that `Index` makes an implicit "empty-column" projection?

dask_expr/tests/test_concat.py
```diff
@@ -128,3 +128,14 @@ def test_concat_axis_one_drop_dfs_not_selected(pdf, df):
...
+
+def test_concat_index(df, pdf):
+    df2 = from_pandas(pdf, npartitions=3)
+    result = concat([df, df2])
+    expected = lib.concat([pdf, pdf])
+    assert_eq(len(result), len(expected))
+
+    query = Len(result.expr).optimize(fuse=False)
+    expected = (0 + Len(df.expr) + Len(df2.expr)).optimize(fuse=False)
+    assert query._name == expected._name
```
rjzamora, line 141: The `0 +` looks slightly weird/fragile to me, but I suppose if it passes...

rjzamora (approved):
Minor suggestion, but lgtm. Thanks!

dask_expr/tests/test_concat.py
rjzamora: Yeah, I see. No worries :)

PR #267, opened 2023-08-15
Fix empty column projection in ``FromPandas``

rjzamora:
Addresses [this code-review comment](https://github.com/dask-contrib/dask-expr/pull/247/files#r1280308640) from #247, which exposed a bug for the case that we absorb an empty column projection into `FromPandas`.

**Notes**: Simply removing the `_absorb_projections` criteria exposed an infinite-recursion bug in `test_fusion.py::test_optimize_fusion_many` which all boiled down to the `.columns` attribute being incorrect for some expressions in the case that there is an empty column projection in `FromPandas`.


phofl:
thx!

PR #268, opened 2023-08-15
Absorb column projections in ``ReadCSV``

rjzamora:
Simple follow-up to https://github.com/dask-contrib/dask-expr/pull/247 for `ReadCSV`.

### DIFF
--- a/dask_expr/io/csv.py
+++ b/dask_expr/io/csv.py
@@ -37,10 +37,7 @@ class ReadCSV(PartitionsFiltered, BlockwiseIO):
 
     @property
     def _meta(self):
-        meta = self._ddf._meta
-        if self.columns is not None:
-            return meta[self.columns[0]] if self._series else meta[self.columns]
-        return meta
+        return self._ddf._meta
 
     @functools.cached_property
     def columns(self):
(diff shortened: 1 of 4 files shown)


dask_expr/io/csv.py
```diff
@@ -37,10 +37,7 @@ def _ddf(self):
 
     @property
     def _meta(self):
-        meta = self._ddf._meta
-        if self.columns:
-            return meta[self.columns[0]] if self._series else meta[self.columns]
-        return meta
+        return self._ddf._meta
```
rjzamora, line 40:
```suggestion
        meta = self._ddf._meta
        if self.operand("columns") is not None:
            return meta[self.columns[0]] if self._series else meta[self.columns]
        return meta
```

dask_expr/io/tests/test_io.py
```diff
@@ -282,12 +282,16 @@ def test_to_parquet(tmpdir, write_metadata_file):
         df2.to_parquet(tmpdir, overwrite=True)
 
 
-def test_combine_similar(tmpdir):
+@pytest.mark.parametrize(
+    "fmt,read_func,read_cls",
```
phofl, line 286: Personally, I like spaces in between these arguments, but no blocker

phofl (approved):

phofl:
thx

dask_expr/io/tests/test_io.py
rjzamora: Didn't even realize I was allowed to leave spaces here, but that makes sense :)

PR #269, opened 2023-08-16
Implement groupby over index

phofl:
This addresses a couple of benchmark failures

### DIFF
--- a/dask_expr/_groupby.py
+++ b/dask_expr/_groupby.py
@@ -206,7 +206,6 @@ class GroupbyAggregation(ApplyConcatApply):
         return {
             "funcs": self.spec["chunk_funcs"],
             "sort": False,
-            "by": self.by,
             **_as_dict("observed", self.observed),
             **_as_dict("dropna", self.dropna),
         }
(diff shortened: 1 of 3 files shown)


dask_expr/_groupby.py
```diff
@@ -188,6 +193,8 @@ def spec(self):
 
     @classmethod
     def chunk(cls, df, by=None, **kwargs):
+        if hasattr(by, "dtype"):
```
phofl, line 196: Not too happy about this, open for suggestions

rjzamora (commented):
I took a quick glance at this, and plan to look more carefully later today.

One immediate thought I have is whether there is any way to avoid adding groupby-specific attributes/properties (like "by") to the classes defined in `_reductions.py`. At a high level, I understand why you took this approach, but I'm still hopeful that we can keep the "general" classes in `_reductions` a bit cleaner. (though, a closer look may convince me otherwise).

rjzamora:
>We have to pack this into args somehow because it can be an expression, open to other approaches. Not happy with this myself...

Right, that makes sense. However, it *may* make the most sense (in the long run) to add some kind of "base" `ApplyConcatApplyBy` and `ChunkBy` classes in `_groupby.py`, and to take the time to keep groupby details out of the general reduction classes.

phofl:
I tried this, but it does a lot of copy-paste things that I liked even less. Ideally I agree with you though

rjzamora:
Just leaving a note that this PR was superseded by https://github.com/dask-contrib/dask-expr/pull/271

PR #272, opened 2023-08-21
Implement HashJoinP2P for merge

### LINKED ISSUE #266: DataFrame merge should use HashJoinLayer instead of two separate P2Ps
```python
import dask_expr
from distributed import Client
c = Client()
df = dask_expr.datasets.timeseries()
df2 = dask_expr.datasets.timeseries()
sum_x = (
    df
    .merge(df2, on=["name", "id"])
    .x_x.sum()
)
print(sum_x.optimize().tree_repr())
```

```
Sum(TreeReduce): split_every=0
  Fused(8cf79):
  | Sum(Chunk):
  |   Projection: columns='x_x'
  |     BlockwiseMerge: left_on=['name', 'id'] right_on=['name', 'id']
  |       Projection: columns=['x', 'name', 'id']
            P2PShuffle: partitioning_index='_partitions' npartitions_out=30 ignore_index=False options=None
              Fused(d0cfd):
              | AssignPartitioningIndex: partitioning_index=['name', 'id'] index_name='_partitions' npartitions_out=30
              |   Timeseries: dtypes={'x': <class 'float'>, 'name': 'string', 'id': <class 'int'>} seed=1132422555
  |       Projection: columns=['name', 'id', 'x']
            P2PShuffle: partitioning_index='_partitions' npartitions_out=30 ignore_index=False options=None
              Fused(f9ad4):
              | AssignPartitioningIndex: partitioning_index=['name', 'id'] index_name='_partitions' npartitions_out=30
              |   Timeseries: dtypes={'name': 'string', 'id': <class 'int'>, 'x': <class 'float'>} seed=946678847
```

is using a `BlockwiseMerge` that acts on two `P2PShuffle` but instead it should be using a `HashJoin`, see https://github.com/dask/distributed/pull/7514 https://github.com/dask/distributed/issues/7496
(shortcoming of scheduler heuristics will not schedule two p2ps efficiently; other binary ops can be affected similarly)

phofl:
The memory footprint of the blockwise version was terrible

- closes #266

### DIFF
--- a/dask_expr/_merge.py
+++ b/dask_expr/_merge.py
@@ -159,10 +159,12 @@ class Merge(Expr):
             return HashJoinP2P(
                 left,
                 right,
-                left_on=shuffle_left_on,
-                right_on=shuffle_right_on,
+                left_on=left_on,
+                right_on=right_on,
                 suffixes=self.suffixes,
                 indicator=self.indicator,
+                left_index=left_index,
+                right_index=right_index,
             )
 
         if shuffle_left_on:
@@ -245,6 +247,8 @@ class HashJoinP2P(Merge, PartitionsFiltered):
         "how",
         "left_on",
         "right_on",
+        "left_index",
+        "right_index",
         "suffixes",
         "indicator",
         "_partitions",
@@ -253,6 +257,8 @@ class HashJoinP2P(Merge, PartitionsFiltered):
         "how": "inner",
         "left_on": None,
         "right_on": None,
+        "left_index": None,
+        "right_index": None,
         "suffixes": ("_x", "_y"),
         "indicator": False,
         "_partitions": None,
@@ -271,6 +277,8 @@ class HashJoinP2P(Merge, PartitionsFiltered):
             right_on=self.right_on,
             indicator=self.indicator,
             suffixes=self.suffixes,
+            left_index=self.left_index,
+            right_index=self.right_index,
         )
 
     def _layer(self) -> dict:
@@ -324,6 +332,8 @@ class HashJoinP2P(Merge, PartitionsFiltered):
                 self.right._meta.drop(columns=_HASH_COLUMN_NAME),
                 self._meta,
                 self.suffixes,
+                self.left_index,
+                self.right_index,
             )
         return dsk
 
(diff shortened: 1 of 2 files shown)


dask_expr/_merge.py
```diff
@@ -1,11 +1,16 @@
...
+from distributed.shuffle._merge import merge_transfer, merge_unpack
+from distributed.shuffle._shuffle import shuffle_barrier
 
-from dask_expr._expr import Blockwise, Expr, Index, Projection
+from dask_expr._expr import Blockwise, Expr, Index, PartitionsFiltered, Projection
 from dask_expr._repartition import Repartition
-from dask_expr._shuffle import Shuffle, _contains_index_name
+from dask_expr._shuffle import AssignPartitioningIndex, Shuffle, _contains_index_name
+
+_HASH_COLUMN_NAME = "__hash_partition"
```
rjzamora, line 13: I take it this is a "magic" string used by `distributed`'s p2p merge?

dask_expr/_merge.py
```diff
@@ -213,6 +238,96 @@ def _simplify_up(self, parent):
...
+                merge_unpack,
+                self.left._name,
+                self.right._name,
+                part_out,
+                _barrier_key_left,
+                _barrier_key_right,
+                self.how,
+                self.left_on,
+                self.right_on,
+                self.left._meta.drop(columns=_HASH_COLUMN_NAME),
```
rjzamora, line 323: I don't love how the p2p merge depends on `_HASH_COLUMN_NAME` and drops this column, but isn't responsible for generating this column. Unfortunately, the main issue is that `merge_unpack` drops the shuffle column, so there is not much we can do here to make the algorithm more intuitive.

To clarify, it would be much clearer if we could just lower the merge to "add partitioning columns" + "p2p merge" + "drop partitioning column"

rjzamora (approved):
Thanks @phofl !

dask_expr/_merge.py
phofl: Yeah I mostly agree but I fixed something else there today. This is on my medium term todo list

dask_expr/_merge.py
phofl: Yep

phofl:
Merging this

PR #273, opened 2023-08-21
from_pandas raising for incompatible npartitions and not respecting sort

phofl:
This wasn't tested well enough... very confusing error message :)


dask_expr/io/io.py
```diff
@@ -199,12 +205,18 @@ def _simplify_up(self, parent):
...
+            return super().npartitions
+        return len(self._divisions_and_locations[0]) - 1
+
     def _locations(self):
         return self._divisions_and_locations[1]
 
     def _filtered_task(self, index: int):
         start, stop = self._locations()[index : index + 2]
-        part = self.frame.iloc[start:stop]
+        part = self._sorted_data.iloc[start:stop]
```
rjzamora, line 219: I don't think we want to do this unless `self.sort` is `True`?

phofl: Oh yes, thx for catching. Added a test as well...

dask_expr/io/io.py
```diff
@@ -199,12 +205,19 @@ def _simplify_up(self, parent):
     def _divisions(self):
         return self._divisions_and_locations[0]
 
+    @functools.cached_property
+    def npartitions(self):
+        if self._filtered:
+            return super().npartitions
+        return len(self._divisions_and_locations[0]) - 1
+
```
rjzamora, line 213: Maybe we can avoid adding this (which I find slightly confusion), and instead fix the source of the actual problem. As far as I understand it, the underlying problem is that [PartitionsFiltered.npartitions is incorrectly calling `super().npartitions`](https://github.com/dask-contrib/dask-expr/blob/4eced7384c09126800df429a6040e926477d2503/dask_expr/_expr.py#L2050) in a case where the `"npartitions"` parameter can/should not be trusted.

Perhaps we just want to change the `return super().npartitions` to `return len(self.divisions) - 1` in `PartitionsFiltered`?

rjzamora: Actually, the problem is probably not in `PartitionsFiltered`, but all the way down in `Expr`! As far as I can tell, it is not really safe to assume the "npartitions" operand will always be satisfied. Therefore, we should probably simplify `Expr.npartitions`

phofl: I don't know about that. This is a pretty good edge case and most likely only annoying when testing. I don't see this as a real world use case, I'd feel more comfortable keeping this isolated in here for now

rjzamora (approved):
Thanks @phofl 

>I don't know about that. This is a pretty good edge case and most likely only annoying when testing. I don't see this as a real world use case, I'd feel more comfortable keeping this isolated in here for now

I don't feel strongly enough to block this fix.

That said, I do disagree with this. I think it is generally a bad idea to assume that an `"npartitions"` operand corresponds to the number of partitions the expression will actually produce. In practice, there are many cases where an `npartitions` parameter cannot be satisfied, and so it will continue being dangerous for a base class like `Expr` to make that promise.

phofl:
merging this since it screws me over in the dask ci

PR #274, opened 2023-08-23
Enable fusion when computing results

phofl:
This feels a bit hacky, Not sure if there's a better place to put this

### DIFF
--- a/dask_expr/_collection.py
+++ b/dask_expr/_collection.py
@@ -145,6 +145,14 @@ class FrameBase(DaskMethodsMixin):
             return new_collection(self.expr.__getitem__(other.expr))
         return new_collection(self.expr.__getitem__(other))
 
+    def persist(self, **kwargs):
+        out = self.optimize()
+        return DaskMethodsMixin.persist(out, **kwargs)
+
+    def compute(self, **kwargs):
+        out = self.optimize()
+        return DaskMethodsMixin.compute(out, **kwargs)
+
     def __dask_graph__(self):
         out = self.expr
         out = out.optimize(fuse=False)
--- a/dask_expr/_collection.py
+++ b/dask_expr/_collection.py
@@ -145,13 +145,17 @@ class FrameBase(DaskMethodsMixin):
             return new_collection(self.expr.__getitem__(other.expr))
         return new_collection(self.expr.__getitem__(other))
 
-    def persist(self, **kwargs):
-        out = self.optimize()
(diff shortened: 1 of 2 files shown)


dask_expr/_collection.py
```diff
@@ -145,6 +145,18 @@ def __getitem__(self, other):
             return new_collection(self.expr.__getitem__(other.expr))
         return new_collection(self.expr.__getitem__(other))
 
+    def persist(self, fuse=True, combine_similar=True, **kwargs):
+        out = self.optimize(combine_similar=combine_similar, fuse=fuse)
+        return DaskMethodsMixin.persist(
+            out, combine_similar=combine_similar, fuse=fuse, **kwargs
```
rjzamora, line 151: Why are `combine_similar` and `fuse` passed into `DaskMethodsMixin.persist` like this? Same question for `compute` below.

dask_expr/_collection.py
```diff
@@ -145,6 +145,18 @@ def __getitem__(self, other):
...
+
+    def compute(self, fuse=True, combine_similar=True, **kwargs):
+        out = self.optimize(combine_similar=combine_similar, fuse=fuse)
+        return DaskMethodsMixin.compute(
+            out, combine_similar=combine_similar, fuse=fuse, **kwargs
+        )
+
     def __dask_graph__(self):
         out = self.expr
         out = out.optimize(fuse=False)
```
rjzamora, line 162: It feels "wrong" that we are always optimizing the task graph here, even if it is in service of a `compute` call with `combine_similar=False`.

I think we need to change the logic in `__dask_graph__`/`__dask_keys__`/etc to only do the minimal "lowering" needed to construct a task graph.  Right now we only have `lower_once`, so we probably need to introduce a `lower` method to repeat `lower_once` until the exression is "fully lowered". What do you think?

rjzamora (commented):
Thanks for looking into this! I have some thoughts/questions

phofl:
@rjzamora I am merging this (need the fused fix somewhere else). Happy to reiterate if you have other comments

rjzamora:
>I am merging this (need the fused fix somewhere else). Happy to reiterate if you have other comments

No problem. However, I might raise an issue to track the fact that passing `combine_similar=False` argument is effectively meaningless, since we always run that optimization when we generate the graph (which is one reason why I feel that https://github.com/dask-contrib/dask-expr/pull/276 is quite important).

PR #275, opened 2023-08-23
Further reduce`` timeseries`` graph size

rjzamora:
Follow up to https://github.com/dask-contrib/dask-expr/pull/270

@phofl pointed out that the size of the task graph was still problematic. This PR should take us down to a size that is close to the graph in `dask.dataframe`.


phofl (approved):
Thx for tackling this

PR #276, opened 2023-08-23
Add ``lower_completely`` method

rjzamora:
Possible compliment to https://github.com/dask-contrib/dask-expr/pull/274

Adds `Expr.lower_completely` to run automatically before graph generation (instead of `optimize`). By avoiding optimization in `__dask_graph__`, we can configure optional optimization options in `compute`/`persist`.

### DIFF
--- a/dask_expr/_collection.py
+++ b/dask_expr/_collection.py
@@ -147,12 +147,12 @@ class FrameBase(DaskMethodsMixin):
 
     def __dask_graph__(self):
         out = self.expr
-        out = out.optimize(fuse=False)
+        out = out.lower_completely()
         return out.__dask_graph__()
 
     def __dask_keys__(self):
         out = self.expr
-        out = out.optimize(fuse=False)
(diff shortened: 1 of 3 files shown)


phofl:
I started chatting about integrating dask-expr better with the scheduler with @fjetter last week. I'd like to hold off on this until we can resume that discussion. Are you ok with this @rjzamora?

rjzamora:
>I started chatting about integrating dask-expr better with the scheduler with @fjetter last week. I'd like to hold off on this until we can resume that discussion. Are you ok with this @rjzamora?

Yes of course. This is not a high priority for me (yet). However, no matter how we choose for dask-expr to interact with the scheduler, I'm fairly confident that we don't want to require `optimize` to generate the graph when the only real requirement is to "lower completely."

phofl (approved):
This is good for now I think, better than optimising in there again. Although this should be a no-op in almost all cases

PR #278, opened 2023-08-28
Add shape and fix size

phofl:
Not having shape lets my debugger compute len, which is a PITA

Size was buggy as well

### DIFF
--- a/dask_expr/tests/test_collection.py
+++ b/dask_expr/tests/test_collection.py
@@ -1197,3 +1197,18 @@ def test_filter_pushdown(df, pdf):
     df_opt = df[["x", "y"]].simplify().replace(1, 5)
     expected = df_opt[df_opt.x > 5]
     assert result._name == expected._name
+
+
+def test_shape(df, pdf):
+    result = df.shape
+    assert result[0]._name == (df.size / 2)._name
+    assert assert_eq(result[0], 100)
+    assert result[1] == 2
+
+    result = df.x.shape
+    assert result[0]._name == (df.x.size)._name
+    assert assert_eq(result[0], 100)
+
+
+def test_size(df, pdf):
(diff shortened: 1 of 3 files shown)


dask_expr/tests/test_collection.py
```diff
@@ -1197,3 +1197,18 @@ def test_filter_pushdown(df, pdf):
     df_opt = df[["x", "y"]].simplify().replace(1, 5)
     expected = df_opt[df_opt.x > 5]
     assert result._name == expected._name
+
+
+def test_shape(df, pdf):
+    result = df.shape
+    assert result[0]._name == (df.size / 2)._name
```
rjzamora, line 1204: Do we really want to test that the implementation of `shape` is related to `size` in this way. If they both return a correct result when computed, then do we really "care" if this assertion fails?

Also can we use something like `pdf.shape[1]` instead of `2`? (or set `expect = pdf.shape` above, etc)

rjzamora (approved):
Thanks @phofl 

Changes look good to me. However, I have some (non-blocking) suggestions for `test_shape`

dask_expr/tests/test_collection.py
phofl: That's what we are doing in dask/dask as well, so I don't think that this hurts? I want to validate that the row-dimension is not eager, we could alternatively check for a Expression or so

rjzamora: >I don't think that this hurts?

No, it doesn't hurt. I'm just hoping to avoid white-box testing when we don't really need it, since I noticed such tests have led to tech debt in dask/dask. This case is not particularly problematic either way, so I'll leave it up to you.

Feel free to merge if/when you think this is ready :)

PR #282, opened 2023-08-29
Adjust for upstream changes in P2P shuffle

phofl:


rjzamora (commented):
Thanks for taking care of this! Do we need to support latest release **and** `main`, or is `main` sufficient?

mrocklin:
I suggest just main.  I'd like to see this project be inertia-free for a
while.  Maybe this means we have to time releases for a while as well.

phofl:
main is fine here, I am planning to do the next release on Friday after the dask release is out

PR #283, opened 2023-08-30
Merge projection selects too many columns

phofl:


dask_expr/_merge.py
```diff
@@ -203,7 +204,14 @@ def _simplify_up(self, parent):
                     projection = [projection]
 
             left, right = self.left, self.right
-            left_on, right_on = self.left_on, self.right_on
+            left_on = _convert_to_list(self.left_on)
+            if left_on is None:
+                left_on = []
```
rjzamora, line 209: Curious why you need this? Tests seem to pass fine without it.

rjzamora: If this is necessary, then we probably want to clean up the dead logic below that assumes `left_on` and `right_on` could be `None`.

phofl: Yeah sorry that got a bit too entangled, that's needed for #284 (which also cleans up the dead logic)

rjzamora (approved):

PR #284, opened 2023-08-30
Make meta calculcation for merge more efficient

phofl:
The non-empty nature of the meta creation is a bottleneck. Avoiding these repeated calculations cuts the optimization time of some of the tpch queries by 60 percent (excluding the read from remote storage part)

sits on top of #283

### DIFF
--- a/dask_expr/_merge.py
+++ b/dask_expr/_merge.py
@@ -337,6 +337,9 @@ class HashJoinP2P(Merge, PartitionsFiltered):
             )
         return dsk
 
+    def _simplify_up(self, parent):
+        return
+
 
 class BlockwiseMerge(Merge, Blockwise):
     """Merge two dataframes with aligned partitions
--- a/dask_expr/_merge.py
+++ b/dask_expr/_merge.py
@@ -37,7 +37,6 @@ class Merge(Expr):
         "suffixes",
         "indicator",
         "shuffle_backend",
-        "_meta",
     ]
     _defaults = {
         "how": "inner",
@@ -48,8 +47,11 @@ class Merge(Expr):
         "suffixes": ("_x", "_y"),
         "indicator": False,
         "shuffle_backend": None,
-        "_meta": None,
     }
+
+    def __init__(self, *args, _precomputed_meta=None, **kwargs):
+        super().__init__(*args, **kwargs)
+        self._precomputed_meta = _precomputed_meta
 
     def __str__(self):
         return f"Merge({self._name[-7:]})"
@@ -71,8 +73,8 @@ class Merge(Expr):
 
     @functools.cached_property
     def _meta(self):
-        if self.operand("_meta") is not None:
-            return self.operand("_meta")
+        if self._precomputed_meta is not None:
+            return self._precomputed_meta
         left = meta_nonempty(self.left._meta)
         right = meta_nonempty(self.right._meta)
         return make_meta(left.merge(right, **self.kwargs))
@@ -108,7 +110,9 @@ class Merge(Expr):
             or right.npartitions == 1
             and how in ("left", "inner")
         ):
-            return BlockwiseMerge(left, right, **self.kwargs, _meta=self._meta)
+            return BlockwiseMerge(
+                left, right, **self.kwargs, _precomputed_meta=self._meta
+            )
 
         # Check if we are merging on indices with known divisions
         merge_indexed_left = (
@@ -169,7 +173,7 @@ class Merge(Expr):
                 indicator=self.indicator,
                 left_index=left_index,
                 right_index=right_index,
-                _meta=self._meta,
+                _precomputed_meta=self._meta,
             )
 
         if shuffle_left_on:
@@ -191,7 +195,7 @@ class Merge(Expr):
             )
 
         # Blockwise merge
-        return BlockwiseMerge(left, right, **self.kwargs, _meta=self._meta)
+        return BlockwiseMerge(left, right, **self.kwargs, _precomputed_meta=self._meta)
 
     def _simplify_up(self, parent):
         if isinstance(parent, (Projection, Index)):
@@ -250,7 +254,7 @@ class Merge(Expr):
                     left[project_left],
                     right[project_right],
                     *self.operands[2:-1],
-                    _meta=self._meta[meta_cols],
+                    _precomputed_meta=self._meta[meta_cols],
                 )
                 if parent_columns is None:
                     return type(parent)(result)
@@ -288,7 +292,7 @@ class HashJoinP2P(Merge, PartitionsFiltered):
 
     @functools.cached_property
     def _meta(self):
-        return self.operand("_meta")
+        return self._precomputed_meta
 
     def _layer(self) -> dict:
         dsk = {}
--- a/dask_expr/_merge.py
+++ b/dask_expr/_merge.py
@@ -37,6 +37,7 @@ class Merge(Expr):
         "suffixes",
         "indicator",
         "shuffle_backend",
+        "_meta",
     ]
     _defaults = {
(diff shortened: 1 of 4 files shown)


dask_expr/_merge.py
```diff
@@ -37,6 +37,7 @@ class Merge(Expr):
         "suffixes",
         "indicator",
         "shuffle_backend",
+        "_meta",
```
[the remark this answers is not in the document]
mrocklin: > We genuinely change the object when we re-create it, which means that we will always trigger a fresh computation of meta

If the inputs to the pd.merge call are going to change then it seems like we need to recompute meta anyway.  If the inputs aren't changing but we're recomputing then maybe that is a sign that we're caching on the wrong things.  Maybe we should have a `staticfunction` or something instead.  

> We can simply not pass meta in this case which would trigger a fresh computation, this is only a fast path

Imagine an optimization which did something like the following:

```python
def subs(obj, old, new):
    operands = [
        new if operand == old else operand  # substitute old for new in all operands
        for operand in obj.operands
    ]
    return type(obj)(*operands)
```

This fails if we store derived state in operands, because `_meta` is in there and really we should choose to not include it any more. 

I'm ok keeping derived state on the class.  I just don't think that it should be in operands.  This probably requires custom constructors.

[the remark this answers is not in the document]
phofl: There are optimisations where meta changes but we compute the necessary information anyway, so we can simply adjust meta if we want to, Projections are a good example for this. 

Lower is an example where meta won't change in case of merge, but you can't properly cache it either. We might introduce a shuffle which means caching will fail for most cases.

I'll add a custom constructor here that should work as well. Then we can see how that looks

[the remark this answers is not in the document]
phofl: Sorry, my comment wasn't clear enough probably. That's how I understood what you were saying, but the problem is as follows:

- That won't work if we introduce a Projection (which is something we could accept, although that wouldn't make me happy in this particular case)
- Meta won't change when we lower the graph, but we will introduce a shuffle, so the inputs of the merge computation will change while lowering, which means that the cache wouldn't work anymore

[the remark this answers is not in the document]
mrocklin: Feels like we're maybe talking past each other here.  Happy to chat live if you're free later.

dask_expr/_merge.py
```diff
@@ -49,6 +49,10 @@ class Merge(Expr):
         "shuffle_backend": None,
     }
 
+    def __init__(self, *args, _precomputed_meta=None, **kwargs):
+        super().__init__(*args, **kwargs)
+        self._precomputed_meta = _precomputed_meta
```
rjzamora, line 54: I think I would strongly prefer if we use a global key-value cache in the same way we cache dataset info for parquet. In fact, we should probably formalize this caching approach to avoid repeating the same kind of logic in multiple places.

It seems like a unique meta depends on a token like...

```python
    @functools.cached_property
    def _meta_cache_token(self):
        return _tokenize_deterministic(
            self.left._meta,
            self.right._meta,
            self.how,
            self.left_on,
            self.right_on,
            self.left_index,
            self.right_index,
            self.suffixes,
            self.indicator,
        )
```

If `self.left._meta` or `self.right._meta` were to change (due to column projection), we would need to recalculate `meta` anyway. However, if the `Merge` object was responsible for pushing down the column projection, we could always update the cache within the simplify logic (since we would already know how the meta needs to change).

phofl: Fwiw I am also not too big of a fan of relying on meta in hashes, there are too many things in pandas that might mutate this unexpectedly, which would break this

rjzamora: To clarify, I don't really mind if we implement stand-alone caching logic in `_merge.py` for now. The thing I'm unsure about in this PR is that we are overriding `__init__` so that we can effectively cache `_meta` without adding `_meta` as an official operand.

It may be the case that this is exactly how we should be attacking this problem in `Merge` (and maybe everywhere). For example, maybe we will eventually have a special `known_meta=` kwarg in `Expr`, which **all** expression objects could leverage. However, since it is not a proper operand, this mechanism feels a bit confusing and fragile to me.

>The slowdown in merge goes back to the nonempty meta objects, not the actual computation on empty objects.

I don't think I understand your point here. Either way we are effectively caching the output of `_meta`, no?

>some of the operations in Lower have side effects, which makes adjusting the meta objects of left and right bothersome and complicated.

I don't see how this is any different for `_precomputed_meta`? In any case where you are confident defining `_precomputed_meta`, you could also just add the "lowered" object to the global cache before returning it.

rjzamora: >Fwiw I am also not too big of a fan of relying on meta in hashes, there are too many things in pandas that might mutate this unexpectedly, which would break this

Interesting. I'd say that should be a fundamental concern for dask-expr in then. What would be then most reliable way to cache the schema?

phofl: > I don't think I understand your point here. Either way we are effectively caching the output of _meta, no?

Yes but the nature of the slowdown makes me think that we need it only in merge and not in other places as of now. I am open to rewriting this here as well if this turns out differently.

> I don't see how this is any different for _precomputed_meta? In any case where you are confident defining _precomputed_meta, you could also just add the "lowered" object to the global cache before returning it.

pandas has some caveats that might change your dtype in meta but not on the actual df. Relying on the initial meta seems saver to me. But this might also be totally wrong.

> To clarify, I don't really mind if we implement stand-alone caching logic in _merge.py for now. The thing I'm unsure about in this PR is that we are overriding __init__ so that we can effectively cache _meta without adding _meta as an official operand.

@mrocklin and I chatted offline and landed on this solution. One motivating factor was the last part here: https://github.com/dask-contrib/dask-expr/pull/284#discussion_r1310193998

phofl: > Interesting. I'd say that should be a fundamental concern for dask-expr in then. What would be then most reliable way to hash the schema?

I don't have a good answer for that. Meta is still the best bet, but it has some limitations. This will get more stable in the future since we are deprecating all of these caveats at the moment.

rjzamora: >Yes but the nature of the slowdown makes me think that we need it only in merge and not in other places as of now. I am open to rewriting this here as well if this turns out differently.

Okay, I think you are are talking about the decision not to write general `Expr`-wide caching code here, which is all good with me. I was only thinking about the decision to use `_precomputed_meta` instead of a simple k/v cache.

Possible problems with the k/v cache approach:
- The meta-hashing issue you mentioned
- We would be keeping the cached meta in memory even after we need it (also a problem for parquet)

Possible problems with `_precomputed_meta`:
- I suppose we are breaking with convention a bit (seems okay)
- Any `substitute_parameters` call will drop the information, even if you aren't changing information that is relevant to meta

> One motivating factor was the last part here ...
Right, I agree that it would be a mistake to make `_precomputed_meta` a proper operand.

phofl: We would also have to patch meta of left and right in the HashJoin layer because that adds a column

``substitute_parameters`` parameters is annoying, we could override but that's not great either.

rjzamora: Sorry for adding all this discussion.  I do get that you are focusing on the `Merge`-specific `meta` issue at hand.

I'm just doing my best to keep the big picture in mind - It seems like we are going to keep running into cases where we would benefit from caching information outside the `Expr` object itself. Therefore, it would be nice if we could design a formal system where a collection of different caches can be managed in one place.

That said, I definitely don't think we need to do something like that right now.

>We would also have to patch meta of left and right in the HashJoin layer because that adds a column

Right, `HashJoinP2P._meta_cache_token` does need to drop the hash columns.

PR #286, opened 2023-08-30
Avoid clearing cached_plan for every dataset

phofl:
Performance impact is small, but debugging get's significantly easier with this


mrocklin (approved):
You might also consider `zict.LRU`.  It's already a dependency I think.  Maybe not worth the extra technology though.

phofl:
I'll look into what we want to use for caching when getting back to meta optimisations

PR #288, opened 2023-08-31
Fix bug in sort_values in distributed setting

phofl:
Our await logic doesn't play nicely with the intermediate computes, hence the different test logic here. dask/dask suffers the same issue

### DIFF
--- a/dask_expr/tests/test_distributed.py
+++ b/dask_expr/tests/test_distributed.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import pytest
+from distributed import Client, LocalCluster
 
 from dask_expr import from_pandas
 from dask_expr.tests._util import _backend_library
@@ -85,4 +86,17 @@ async def test_merge_p2p_shuffle(c, s, a, b):
     x = await x
     lib.testing.assert_frame_equal(
         x.reset_index(drop=True), df_left.merge(df_right)[["b", "c"]]
+    )
+
+
+def test_sort_values():
+    with LocalCluster() as cluster:
+        with Client(cluster) as client:  # noqa: F841
+            pdf = lib.DataFrame({"a": [5] + list(range(100)), "b": 2})
+            df = from_pandas(pdf, npartitions=10)
+
+            out = df.sort_values(by="a").compute()
+    lib.testing.assert_frame_equal(
+        out.reset_index(drop=True),
+        pdf.sort_values(by="a", ignore_index=True),
     )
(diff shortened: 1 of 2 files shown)


dask_expr/tests/test_distributed.py
```diff
@@ -86,3 +87,16 @@ async def test_merge_p2p_shuffle(c, s, a, b):
     lib.testing.assert_frame_equal(
         x.reset_index(drop=True), df_left.merge(df_right)[["b", "c"]]
     )
+
+
+def test_sort_values():
+    with LocalCluster() as cluster:
```
rjzamora, line 93: Any way to leverage `gen_cluster` here?

phofl: No, it's async

rjzamora: Can you explain more? Also, do we really want to allow a test to use all the available cores on our machine like this?

mrocklin: You maybe want these fixtures from distributed.utils_test

```python
@pytest.fixture
def cluster_fixture(loop):
    with cluster() as (scheduler, workers):
        yield (scheduler, workers)


@pytest.fixture
def s(cluster_fixture):
    scheduler, workers = cluster_fixture
    return scheduler


@pytest.fixture
def a(cluster_fixture):
    scheduler, workers = cluster_fixture
    return workers[0]


@pytest.fixture
def b(cluster_fixture):
    scheduler, workers = cluster_fixture
    return workers[1]


@pytest.fixture
def client(loop, cluster_fixture):
    scheduler, workers = cluster_fixture
    with Client(scheduler["address"], loop=loop) as client:
        yield client
```

They handle cleanup and various sanity checks. 

Alternatively, if you want things to be very fast, you might also want `with LocalCluster(processes=False)`

phofl: the gen_cluster decorator creates a async cluster and async tests will fail for sort_values/set_index because of intermediate computes. That's not a dask-expr problem, dask/dask has the same issue. 

Sure, we can restrict the number of workers

phofl: I think I prefer the fast version, this is only supposed to test that the function serialisation works as expected

rjzamora: >Sure, we can restrict the number of workers

Sorry for mentioning that - I care less about that detail (for now).

phofl: See https://github.com/dask/distributed/pull/8167 This one has the same issue in dask/dask

PR #289, opened 2023-08-31
Implement combine_similar for merge

phofl:
That was trickier than expected. We want to pull projections after the merge operation if we can avoid merging multiple times through this. I am not touching filters here since we are not pushing them through either, that might be something for later (but pushing filter through merges can be tricky). 

We performed the same 2 merge operation 3 times in one of the tpch queries, this PR cuts the runtime by 70%

### DIFF
--- a/dask_expr/_merge.py
+++ b/dask_expr/_merge.py
@@ -6,7 +6,14 @@ from distributed.shuffle._core import ShuffleId, barrier_key
 from distributed.shuffle._merge import merge_transfer, merge_unpack
 from distributed.shuffle._shuffle import shuffle_barrier
 
-from dask_expr._expr import Blockwise, Expr, Index, PartitionsFiltered, Projection
+from dask_expr._expr import (
+    Blockwise,
+    Expr,
+    Filter,
+    Index,
+    PartitionsFiltered,
+    Projection,
+)
 from dask_expr._repartition import Repartition
 from dask_expr._shuffle import AssignPartitioningIndex, Shuffle, _contains_index_name
 from dask_expr._util import _convert_to_list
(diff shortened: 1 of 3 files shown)


rjzamora:
>this PR cuts the runtime by 70%

Nice work! I'll take a look as soon as I can, but that might not be until tomorrow.

phofl:
Do you have time to take a look @rjzamora ?

dask_expr/_merge.py
```diff
@@ -247,6 +254,63 @@ def _simplify_up(self, parent):
                     return type(parent)(result)
                 return result[parent_columns]
 
+    def _combine_similar(self, root: Expr):
+        # Push projections back up to avoid performing the same merge multiple times
+
+        def _remove_projections(frame):
```
rjzamora, line 260: This feels complicated enough to be a stand-alone function/utility with a comment or doc-string explaining it's behavior.

dask_expr/_merge.py
```diff
@@ -247,6 +254,63 @@ def _simplify_up(self, parent):
...
+            operations, columns = [], []
+            frame_base = frame
+            while isinstance(
+                frame, (AssignPartitioningIndex, Shuffle, Filter, Projection)
+            ):
+                # Have to respect ops that were injected while lowering or filters
+                if isinstance(frame, Projection):
```
rjzamora, line 267: A lot of this logic feels similar to `Blockwise._combine_similar`. Have you considered the possibility of generalizing that code, and using it in other places?

rjzamora (commented):
Sorry - I still haven't found the time to look through this PR in detail (or to think carefully about possible alternatives).

I get the sense that this solution is probably fine, so no need to wait on my "full" review if I don't get to it soon.

dask_expr/_merge.py
phofl: refactored into a more general method

phofl:
I've refactored as suggested by you. Will merge this now (have something I want to build on top of that) but happy to iterate if necessary

PR #291, opened 2023-09-01
Implement StringAccessor

phofl:
That's one of the missing things for tpch

### DIFF
--- a/dask_expr/_str_accessor.py
+++ b/dask_expr/_str_accessor.py
@@ -1,0 +1,122 @@
+from dask_expr._accessor import Accessor
+from dask_expr._expr import Blockwise
+from dask_expr._reductions import Reduction
+
+
+class StringAccessor(Accessor):
+    """Accessor object for string properties of the Series values.
+
+    Examples
+    --------
+
+    >>> s.str.lower()  # doctest: +SKIP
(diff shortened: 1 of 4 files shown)


dask_expr/_str_accessor.py
```diff
@@ -0,0 +1,122 @@
...
+
+        return new_collection(CatBlockwise(self._series.expr, others, sep, na_rep))
+
+    def __getitem__(self, index):
+        return self._function_map("__getitem__", index)
+
+
+class CatBlockwise(Blockwise):
+    _parameters = ["frame", "others", "sep", "na_rep"]
+    operation = lambda _, ser, *args, **kwargs: ser.str.cat(*args, **kwargs)
```
mrocklin, line 113: I recommend avoiding lambdas if you can.  They can make serialization slower than desired.  This seems like a good place for a `staticfunction`

phofl: Thx.

I didn't know that the slowdown is big enough that it matters for us.

mrocklin: It used to be non-trivial to the point where I went through all of dask to remove lambdas.  I'm not sure about today.

PR #294, opened 2023-09-08
[WIP] Allow expressions to be shipped to the scheduler

fjetter:
Given the following two PRs, these minor changes would allow us to pass the expressions to the scheduler via pickle
- https://github.com/dask/dask/pull/10369
- https://github.com/dask/distributed/pull/7942

cc @phofl


dask_expr/_collection.py
```diff
@@ -93,16 +93,20 @@ def _wrap_unary_expr_op(self, op=None):
...
+class FrameBase(DaskMethodsMixin, NewDaskCollection):
     """Base class for Expr-backed Collections"""
 
     __dask_scheduler__ = staticmethod(
         named_schedulers.get("threads", named_schedulers["sync"])
     )
     __dask_optimize__ = staticmethod(lambda dsk, keys, **kwargs: dsk)
 
+    def __dask_tokenize__(self):
+        return self.expr._name
```
fjetter, line 108: This is just defined as part of the collections protocol. Not sure if it is actually required

dask_expr/_collection.py
```diff
@@ -154,14 +158,7 @@ def compute(self, fuse=True, combine_similar=True, **kwargs):
...
 
     def __dask_graph__(self):
-        out = self.expr
-        out = out.lower_completely()
-        return out.__dask_graph__()
-
-    def __dask_keys__(self):
-        out = self.expr
-        out = out.lower_completely()
-        return out.__dask_keys__()
```
fjetter, line 164: Having keys defined on the _collections_ level feels like an abstraction leak. Among other things this is what threw me off for a while when implementing this the first time. I had a hard time distinguishing collections from graphs in the existing code. I find this now a bit clearer in the above PRs

dask_expr/_collection.py
```diff
@@ -178,6 +175,15 @@ def optimize(self, combine_similar: bool = True, fuse: bool = True):
     def dask(self):
         return self.__dask_graph__()
 
+    def finalize_compute(self) -> FrameBase:
+        from ._repartition import RepartitionToFewer
+        if self.npartitions > 1:
+            return new_collection(RepartitionToFewer(self.expr, 1))
+        return self
```
fjetter, line 182: This is not really new but instead of doing the graph mutations ourselves by calling postcompute, etc. in the client or in dask, we are now effectively just modifying the expression to what the end result should look like. In this case, it's just a concat of all partitions

dask_expr/_collection.py
```diff
@@ -178,6 +175,15 @@ def optimize(self, combine_similar: bool = True, fuse: bool = True):
...
+
+    def postpersist(self, futures: dict) -> NewDaskCollection:
+        return from_graph(futures, self._meta, self.divisions, self._name)
```
fjetter, line 185: This is also not really new but I moved away from the "method that returns a callable [...]" API to just calling the callable. I haven't checked with any downstream projects, yet, if this complex callable construct is actually required. Not married to anything here.

dask_expr/_expr.py
```diff
@@ -699,8 +702,10 @@ def dtypes(self):
     def _meta(self):
         raise NotImplementedError()
 
-    def __dask_graph__(self):
+    def _materialize(self):
         """Traverse expression tree, collect layers"""
+        from distributed.scheduler import ensure_materialization_allowed
+        ensure_materialization_allowed()
```
fjetter, line 708: This is mostly for ad-hoc testing. If we find it valuable we can actually add this as a proper API

dask_expr/_expr.py
```diff
@@ -699,8 +702,10 @@ def dtypes(self):
     def _meta(self):
         raise NotImplementedError()
 
-    def __dask_graph__(self):
+    def _materialize(self):
```
fjetter, line 705: The underscore is probably a bit silly. Will likely change this again

dask_expr/_expr.py
```diff
@@ -35,7 +35,10 @@
 no_default = "__no_default__"
 
 
-class Expr:
+from dask.typing import DaskGraph
+# Note: subclassing isn't required. This is just for the prototype to have a
+# check for abstractmethods but the runtime checks for duck-typing/protocol only
+class Expr(DaskGraph):
```
fjetter, line 41: as the note says, there is no need for subclassing. this is just for prototyping (or we keep it, no strong preferences)

fjetter:
CI is obviously failing because I haven't set up CI to run against dask / distributed. I'll clean this up soon on the other PRs as well

fjetter:
There is only one caveat so far: Since we're delaying materialization until the graph is on the scheduler and since we don't have a concept of `merge(collection1, collection2)` in `dask-expr` yet (basically what `dask.base.collections_to_dsk` is doing) we have to submit expressions sequentially to the scheduler, i.e.

`dask.compute(collection1, collection2)` will generate two requests to the scheduler.

This has the shortcoming that we will not be performing a "multi-collections" optimization and we will not order the *entire* graph at once.  We will still deduplicate keys that are in both collections.

I find this approach a decent compromise to get started. If we notice this to be a severe problem we can think about a `merge` operation of some sorts.

dask_expr/_expr.py
phofl: This would add distributed as a hard dependency, wouldn't it? That's something I'd rather avoid

fjetter: 🔝 Draft ;)

fjetter: **If** we implemented an API like this, this would need to live in dask/dask

mrocklin:
> There is only one caveat so far: Since we're delaying materialization until the graph is on the scheduler and since we don't have a concept of merge(collection1, collection2) in dask-expr 

What if we made a `Tuple(Expr)` term?  Would that resolve this issue? (I haven't thought deeply about what's going on here yet, just throwing out ideas)

dask_expr/_collection.py
```diff
@@ -283,7 +280,18 @@ def optimize(self, fuse: bool = True):
 
     @property
     def dask(self):
-        return self.__dask_graph__()
+        # FIXME: This is highly problematic. Defining this as a property can
+        # cause very unfortunate materializations. Even a mere hasattr(obj,
+        # "dask") check already triggers this since it's a property, not even a
+        # method.
+        return self.__dask_graph_factory__().optimize().materialize()
```
crusaderky, line 287: What happens if we remove it completely? I guess some bits in xarray and pint will break but there will be work to be done in xarray/pint anyway so I don't see a problem.

dask_expr/_collection.py
```diff
@@ -283,7 +280,18 @@ def optimize(self, fuse: bool = True):
...
+
+    def finalize_compute(self):
+        return new_collection(Repartition(self.expr, 1))
```
crusaderky, line 290: This is very problematic in the _fairly common_ use case where the client mounts a lot more memory than a single worker. This forces the whole object to be unnecessarily collected onto a single worker and then sent to the client, whereas we could just have the client fetch separate partitions from separate workers (which may or may not happen all at once if it needs to transit through the scheduler).

This replicates the issue with the current finalizer methods in dask/dask, which are created by `dask.compute(df)` but are skipped by `df.compute()`.

fjetter (commented):
had some staged comments...

fjetter:
Brief update: This is mostly working. There are still failing tests I have to track down. Most of the things I had to fix so far are somehow coupled to me adding a `Repartition` expression as a postcompute which changes how the graph is handled compared to how it is done right now. As a reminder, right now it is materialized and this concat/repartition task is added to the low level graph.

This reveals stuff like https://github.com/dask/dask/pull/10739 

I will likely want to nuke the `dask` property. In the test suite this does all kinds of weird things like triggering quantile computations during an `hasattr` check. I don't think this is user friendly. This isn't a decision that has to be done on this PR, though and I suggest to postpone this to a follow up PR. This one with the siblings in dask and distributed will be big enough once everything is done

PR #295, opened 2023-09-08
Implement Groupby.apply

phofl:


phofl:
merging this, we should Get some support for apply

PR #297, opened 2023-09-11
Add optimization for Head(SortValues) -> nlargest/nsmallest

mrocklin:
It feels nice.  Good follow-on work would be to support set_index.

### DIFF
--- a/dask_expr/_shuffle.py
+++ b/dask_expr/_shuffle.py
@@ -820,6 +820,19 @@ class SortValues(BaseSetIndexSortValues):
         )
 
     def _simplify_up(self, parent):
+        from dask_expr._expr import Head, Tail
+        from dask_expr._reductions import NLargest, NSmallest
+
+        if isinstance(parent, Head):
+            if self.ascending:
+                return NSmallest(self.frame, n=parent.n, _columns=self.by)
+            else:
+                return NLargest(self.frame, n=parent.n, _columns=self.by)
+        if isinstance(parent, Tail):
+            if self.ascending:
+                return NLargest(self.frame, n=parent.n, _columns=self.by)
+            else:
+                return NSmallest(self.frame, n=parent.n, _columns=self.by)
         if isinstance(parent, Projection):
             parent_columns = parent.columns
             columns = parent_columns + [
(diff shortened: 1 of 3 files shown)


dask_expr/_shuffle.py
```diff
@@ -740,6 +740,29 @@ def _lower(self):
         return SetPartition(self.frame, self._other, self.drop, divisions)
 
     def _simplify_up(self, parent):
+        from dask_expr._expr import Head, Tail
+
+        # TODO, handle setting index with other frame
+        if (
+            isinstance(parent, Head)
+            and isinstance(self._other, (int, str))
```
phofl, line 748: Could we use the ``is_scalar`` check from dask_expr._util?

dask_expr/_shuffle.py
```diff
@@ -740,6 +740,29 @@ def _lower(self):
         return SetPartition(self.frame, self._other, self.drop, divisions)
 
     def _simplify_up(self, parent):
+        from dask_expr._expr import Head, Tail
+
+        # TODO, handle setting index with other frame
```
phofl, line 745: Injecting an Assign is probably the easiest way of getting there

dask_expr/_shuffle.py
```diff
@@ -740,6 +740,29 @@ def _lower(self):
...
+            and self._other in self.frame.columns
+        ):
+            return SetIndex(
+                NSmallest(self.frame, n=parent.n, _columns=self._other),
+                _other=self._other,
+            )
+
+        if (
+            isinstance(parent, Tail)
+            and isinstance(self._other, (int, str))
```
phofl, line 758: same here

dask_expr/_shuffle.py
mrocklin:
```
(Pdb) pp self._other
df['x']
(Pdb) pp is_scalar(self._other)
True
```

Something is wrong here.   See also https://github.com/dask-contrib/dask-expr/issues/298

dask_expr/_shuffle.py
mrocklin: Agreed.  I started doing this but it got complicated, so I stopped.

dask_expr/_shuffle.py
phofl: I'll take a look, this worries me a bit

mrocklin:
I've removed out the set_index stuff, which is more complicated.  I'm going to merge in the sort_values stuff and then raise another PR on top with set_index.  

This is mostly because I have a few other small things I want to raise and my branches on branches are getting weird.  

Sorry for the self merge.

PR #304, opened 2023-09-20
Fix equality comparison of scalars with expressions

phofl:
cc @mrocklin 

This is one potential approach that we could. We would have to remove every other binary op from Expr as well, since, ``2 < df.expr`` is still True.

The alternative would be to raise for ``bool(Expr)``, which might make more sense, although we couldn't use Expressions in Sets or as dictionary keys anymore. ``bool(DaskDataFrame)`` raises as well

### DIFF
--- a/dask_expr/_collection.py
+++ b/dask_expr/_collection.py
@@ -108,6 +108,11 @@ class FrameBase(DaskMethodsMixin):
     def __init__(self, expr):
         self._expr = expr
 
+    def __eq__(self, other):
+        if isinstance(other, FrameBase):
+            other = other.expr
+        return new_collection(expr.EQ(self.expr, other))
+
     @property
     def expr(self) -> expr.Expr:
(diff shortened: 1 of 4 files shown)


mrocklin:
I don't have a strong opinion here yet.  Do you?

phofl:
No not a strong opinion at this moment.

I think making ``bool(Expr)`` raise gives us more stability guarantees in the future but it will be annoying to deal with in the beginning

mrocklin:
I'm fine with having `bool(expr)` raise.  I guess this lets us continue to use binary relational operators.  

Are there cases today where we want to use expressions in sets and dicts?  That's probably relevant to help make this decision.

phofl:
There are cases where we are using them but I think that's not a good idea there. expr._name is a better choice in most cases I think

phofl:
#307

PR #307, opened 2023-09-21
Make bool(expr) raise

phofl:
fusing is based on names now, which means we have to keep track of the associated expression, everything else is an easy change.

I think this makes more sense long-term compared to bool(expr) always returning True

alternative to #304

### DIFF
--- a/dask_expr/_expr.py
+++ b/dask_expr/_expr.py
@@ -760,7 +760,7 @@ class Expr:
     def __dask_keys__(self):
         return [(self._name, i) for i in range(self.npartitions)]
 
-    def substitute(self, substitutions: dict) -> Expr:
+    def substitute(self, substitutions: list[tuple]) -> Expr:
         """Substitute specific `Expr` instances within `self`
 
         Parameters
@@ -778,6 +778,10 @@ class Expr:
         if not substitutions:
             return self
 
+        substitutions = {
+            k._name if isinstance(k, Expr) else k: v for k, v in substitutions
+        }
+
         if self._name in substitutions:
             return substitutions[self._name]
 
@@ -793,7 +797,7 @@ class Expr:
                 new.append(substitutions[operand])
                 update = True
             elif isinstance(operand, Expr):
-                val = operand.substitute(substitutions)
+                val = operand.substitute([(k, v) for k, v in substitutions.items()])
                 if operand._name != val._name:
                     update = True
                 new.append(val)
@@ -2293,7 +2297,7 @@ def optimize_blockwise_fusion(expr):
                         for operand in _expr.dependencies()
                         if operand._name not in local_names
                     ]
-                to_replace = {group[0]._name: Fused(group, *group_deps)}
+                to_replace = [(group[0], Fused(group, *group_deps))]
                 _ret = expr.substitute(to_replace)
                 return _ret, not roots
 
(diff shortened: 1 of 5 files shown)


dask_expr/_expr.py
```diff
@@ -496,6 +496,12 @@ def __getitem__(self, other):
         else:
             return Projection(self, other)  # df[["a", "b", "c"]]
 
+    def __bool__(self):
+        raise ValueError(
+            f"The truth value of a {self.__class__.__name__} is ambiguous. "
+            "Use a.any() or a.all()."
+        )
```
mrocklin, line 503: Could also call `bool(self._meta)`.  Not a big deal though.

dask_expr/_shuffle.py
```diff
@@ -857,7 +857,9 @@ def _simplify_up(self, parent):
                 return NSmallest(self.frame, n=parent.n, _columns=self.by)
         if isinstance(parent, Filter):
             return SortValues(
-                Filter(self.frame, parent.predicate.substitute({self: self.frame})),
+                Filter(
+                    self.frame, parent.predicate.substitute({self._name: self.frame})
```
mrocklin, line 861: This API seems unpleasant.  Maybe we should change the API to one of the following:

```python
def subs(self, old, new):

>>> self.subs(self, self.frame)
```

or 

```python
def subs(self, pairs):

>>> self.subs([(self, self.frame)])
```

mrocklin: I'll suggest the former until we need the latter.  (SymPy, for reference, does both).

mrocklin (commented):
I have a concern about subs, but that's it here.

dask_expr/_expr.py
phofl: that would raise as well, bool is not allowed for pandas objects

dask_expr/_shuffle.py
phofl: You want to get rid of the dict?

And you also want to convert from an expression to the name inside of substitute? 

Just checking that I am not misunderstanding anything

dask_expr/_expr.py
mrocklin: I should have been more explicit in my comment.  Calling `bool(meta)` would just defer our behavior to pandas', which might be nicer long-term (not that I expect pandas' behavior to change).

phofl: I think that would cause problems if meta is a scalar since that would work but potentially return incorrect results

dask_expr/_shuffle.py
mrocklin: Right now this method requires users to use the `._name` API to use it effectively, which seems off.  I think that the API here should consume and produce expressions.  I wouldn't expect users to know that they have to use the `._name` attribute.  

Given that we can't put exprs in dicts as keys, we need to find some other way to craft the API I think.  

If we expect the common case to be a single 1-1 substitution, then putting the old and new as separate arguments should work

```python
x.subs({y: z}) # before
x.subs(y, z)   # after
```

mrocklin: This seems better than 

```python
x.subs({y._name: z})
```

phofl: Yep makes sense, updated

dask_expr/_expr.py
```diff
@@ -769,32 +775,33 @@ def substitute(self, substitutions: dict) -> Expr:
         >>> (df + 10).substitute({10: 20})
```
mrocklin, line 775:
```suggestion
        >>> (df + 10).substitute(10, 20)
```

mrocklin (approved):

PR #311, opened 2023-09-22
Implement predicate pushdown for set_index

phofl:


dask_expr/_shuffle.py
```diff
@@ -774,6 +774,14 @@ def _simplify_up(self, parent):
                 parent.operand("columns"),
             )
 
+        if isinstance(parent, Filter):
+            p = parent.predicate
+            if any(isinstance(x, Index) for x in p.walk()):
+                # Punt on cases where the new index is part of the filter
+                return
```
mrocklin, line 781: This is fine, but it would probably be more powerful if we were to replace `parent` in `parent.predicate` with some flattened/history-less version of itself.  

For example, if somewhere in `parent`'s history it has a line like the following:

```python
df["z"] = df.x + df.index
```

Then this condition would trigger unnecessarily.  

Probably we want to consider a version of the predicate where parent has no history.

```python
flattened = dx.from_pandas(parent._meta)
predicate = parent.predicate.subs(parent, flattened)

if any(isinstance(x, Index) for x in predicate.walk()):
    return
```

However, this is not necessary.  The condition above is correct, even if sometimes overzealous.  Happy to merge without this change (and it might add some complexity).

mrocklin (approved):

dask_expr/_shuffle.py
phofl: Yeah I agree but his might get complex since predicate can get arbitrarily nested. I want to do this as follow up but that’s not super urgent I think

mrocklin: Agreed on not urgent (merge at will).

PR #314, opened 2023-09-26
[Proposal] Provide limited support for pandas<2.0

rjzamora:
In the past, I've mentioned that cudf had fallen a bit behind in terms of pandas-2.0 support. It is starting to look like cudf will **not** be supporting pandas>1.5.3 until December 2023 at the very earliest.

Since it has become challenging for me to test dask-expr + cudf with pandas-2.0 support, I'd like to propose that dask-expr provide limited support for pandas<2.0 until cudf catches up.  This is **not** a suggestion that we directly prioritize, test or advertise pandas<2.0 support in any way. Rather, I am hoping that we can acknowledge that anyone using a `cudf` backend will also need to have `pandas<=1.5.3` in their environment, and so (for now) we add a few `PANDAS_GE_200` checks to the code base.

Please keep in mind that I really do want to avoid unnecessary compatibility checks in dask-expr.  However, I just don't think that avoiding these checks is worth the current development pain I am facing.

### DIFF
--- a/dask_expr/_reductions.py
+++ b/dask_expr/_reductions.py
@@ -4,6 +4,7 @@ import numpy as np
 import pandas as pd
 import toolz
 from dask.dataframe import hyperloglog, methods
+from dask.dataframe._compat import PANDAS_GE_200
 from dask.dataframe.core import (
     _concat,
     idxmaxmin_agg,
@@ -293,7 +294,9 @@ class DropDuplicates(Unique):
 
     @property
     def chunk_kwargs(self):
-        return {"ignore_index": self.ignore_index, **self._subset_kwargs()}
+        if PANDAS_GE_200:
+            return {"ignore_index": self.ignore_index, **self._subset_kwargs()}
+        return self._subset_kwargs()
 
     def _simplify_up(self, parent):
         if self.subset is not None and isinstance(parent, Projection):
--- a/dask_expr/tests/test_collection.py
+++ b/dask_expr/tests/test_collection.py
@@ -6,7 +6,7 @@ import pickle
 import dask
 import numpy as np
 import pytest
-from dask.dataframe._compat import PANDAS_GE_210
+from dask.dataframe._compat import PANDAS_GE_200, PANDAS_GE_210
 from dask.dataframe.utils import UNKNOWN_CATEGORIES, assert_eq
 from dask.utils import M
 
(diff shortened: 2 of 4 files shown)


mrocklin:
As a heads-up, Patrick is on PTO for the next week or so.

rjzamora:
>As a heads-up, Patrick is on PTO for the next week or so.

Ah right. Thanks, he did tell me about that last week. @mrocklin - Do you have any strong feelings here?

mrocklin:
In general I'd like dask-expr to be future-leaning and not burdened with legacy stuff.  🤔 if it gets you or other NVIDIA folks more active on this project then sure, it's probably worth it (assuming the cost is low, like what's here).  My recommendation would probably be to not include pandas < 2 in CI, and intentionally leave the maintenance burden to you all.  

Probably we'd also have a hard cutoff of EOY, after which point RAPIDS would be very much on its own.

Thoughts?

rjzamora:
>if it gets you or other NVIDIA folks more active on this project then sure, it's probably worth it

Right. I'd really like to use dask-expr + cudf benchmarking to motivate more RAPIDS engagement, but right now it is just not practical to build cudf with pandas-2 support (I've already wasted too much time on this). Therefore, I will need to ignore the pandas>=2.0 requirement one way or another to stay engaged.

>(assuming the cost is low, like what's here).

The amount of code needed to support pandas<2 in the dask-expr API is pretty tiny. As far as I can tell, the only important difference is a new `ignore_index` argument to `drop_duplicates`. We will almost certainly need a lot more code to fully support the differences between pandas and cudf anyway...

>My recommendation would probably be to not include pandas < 2 in CI, and intentionally leave the maintenance burden to you all.

Yeah, sorry for not clarifying that above. We can treat pandas<2 support in the same way we treat cudf support (the maintenance burden is on us).

>Probably we'd also have a hard cutoff of EOY, after which point RAPIDS would be very much on its own.

I'm pretty confident that cudf will prioritize pandas-2 support after the 23.10 release (I know it is important to them), and I can honestly say that I will personally push hard for this. However, RAPIDS/cudf does have other short term priorities (unannounced, but likely to require significant effort from the RAPIDS/Dask team in the near future as well), and so there is always a small chance of slippage that I don't personally have any control over.

phofl (approved):
I am okay with this as well as long as we can keep the changes small.

PR #316, opened 2023-09-27
Revise `Expr.substitute` to distinguish literal-value substitution

rjzamora:
The current implementation of `Expr.substitute` does not work with `cudf`-backed collections, because `cudf` objects are "hashable."  This PR revises the logic a bit to distinguish the logic used for `Expr` and literal terms a bit more clearly.

### DIFF
--- a/dask_expr/_expr.py
+++ b/dask_expr/_expr.py
@@ -827,6 +827,21 @@ class Expr:
                     update = True
                 new_exprs.append(val)
             elif (
+                isinstance(self, Fused)
+                and isinstance(operand, list)
+                and all(isinstance(op, Expr) for op in operand)
+            ):
+                # Special handling for `Fused`.
+                # We make no promise to dive through a
+                # list operand in general, but NEED to
+                # do so for the `Fused.exprs` operand.
+                val = []
+                for op in operand:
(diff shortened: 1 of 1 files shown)


dask_expr/_expr.py
```diff
@@ -789,42 +789,66 @@ def __dask_keys__(self):
...
                 new_exprs.append(val)
+            elif (
+                isinstance(self, Fused)
+                and isinstance(operand, list)
+                and all(isinstance(op, Expr) for op in operand)
+            ):
+                # Special handling for `Fused`.
+                # We make no promise to dive through a
+                # list operand in general, but NEED to
+                # do so for the `Fused.exprs` operand.
```
rjzamora, line 837: I'm having trouble finding a simple reproducer for this. However, this bug does seem to bite us for the specific implementation of q7 used in the pola-rs/tpch repo.

phofl (approved):

dask_expr/_expr.py
phofl: I'll try to take a look as well (I have some practice with coming up with reproducer from these queries.... annoying task), but should be good to merge as is

PR #317, opened 2023-10-02
Use ``_required_attribute`` in ``idxmin/max``

rjzamora:
Since cudf does not support `idxmin/max`, the `_required_attribute` property needs to be set to avoid recursive `getattr` "hang".


rjzamora:
Merging this trivial PR to avoid the need to combine too many unmerged PRs for local cudf testing. Happy to revise this change if needed.

PR #318, opened 2023-10-02
Avoid top-level ``distributed`` import in ``_merge.py``

rjzamora:
We are currently importing `distributed` at the top of `_merge.py`. However, `distributed` is not a required dependency for `dask-expr`. Therefore, this PR moves the offending imports to the `_layer` method that requires it.


phofl (approved):
thx

PR #320, opened 2023-10-03
Add ``_BackendData`` utility class to improve ``FromParquet`` caching

rjzamora:
While exploring the naively-implemented tpch queries in pola-rs/tpch (using scale-factor 1), I noticed that optimization performance is terrible for complicated queries containing `FromPandas` expressions. This is because we end up calling `tokenize` and `_divisions_and_locations` on the backend data repetitively.

This PR adds a `_BackendData` helper class to effectively wrap the backend data so we we can cache the results of  `tokenize` and `_divisions_and_locations` **outside** the `FromPandas` class.


phofl (approved):

phofl:
thx

PR #322, opened 2023-10-09
Fix ``sort_values`` behavior for "tasks"-based shuffle and add missing kwargs

rjzamora:
While exploring ways to make `sort_values` work with a cudf backend, I discovered that `main` will currently return wrong results when the "dataframe.shuffle.method" configuration is set to "tasks". This PR fixes the underlying bug, and adds `shuffle` and `ignore_index` parameters (to `SortValues`) to make this behavior easier to test/control.

### DIFF
--- a/dask_expr/tests/test_shuffle.py
+++ b/dask_expr/tests/test_shuffle.py
@@ -223,20 +223,21 @@ def test_set_index_without_sort(df, pdf):
         df.set_index("y", sorted=True, npartitions=20)
 
 
-def test_sort_values(df, pdf):
-    assert_eq(df.sort_values("x"), pdf.sort_values("x"))
-    assert_eq(df.sort_values("x", npartitions=2), pdf.sort_values("x"))
+@pytest.mark.parametrize("method", [None, "tasks"])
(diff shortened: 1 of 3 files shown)


dask_expr/tests/test_shuffle.py
```diff
@@ -223,20 +223,32 @@ def test_set_index_without_sort(df, pdf):
...
+
+@pytest.mark.parametrize("shuffle", [None, "tasks"])
+def test_sort_values_temporary_column_dropped(shuffle):
+    pdf = lib.DataFrame(
+        {"x": range(10), "y": [1, 2, 3, 4, 5] * 2, "z": ["cat", "dog"] * 5}
+    )
+    df = from_pandas(pdf, npartitions=2)
+    _sorted = df.sort_values(["z"], shuffle=shuffle)
+    result = _sorted.compute()
+    assert "_partitions" not in result.columns
```
rjzamora, line 251: This is the bug I see on `main`

phofl (approved):

phofl:
thx

Issue #323, opened 2023-10-09
`Expr` instances are serialized as part of the graph


fjetter:
Running some profiling on the TCPH benchmarks, I noticed a significant portion of the main thread to be blocked by the calculation of `_meta` during the deserialization of `Expr` objects

![image](https://github.com/dask-contrib/dask-expr/assets/8629629/9d3a3bcc-4478-4f61-8dde-a93a0dfcfa03)

This screenshot indicates that the expr instance that is being (de-)serialized is the `Merge` expression.

Looking through the traces a bit more, I can also find references to a parquet layer which almost looks like it would fetch some data (but I hope this is false). It is definitely performing some pyarrow table conversion.

![image](https://github.com/dask-contrib/dask-expr/assets/8629629/f98b9e2a-c995-4b19-8090-0dc4f012ffd9)

speedscope file (a Worker)
[tls-10_0_32_152-45157.json.zip](https://github.com/dask-contrib/dask-expr/files/12847361/tls-10_0_32_152-45157.json.zip)

Just skimming the code, I saw stuff like this

https://github.com/dask-contrib/dask-expr/blob/52346744cf530ca86108553fb0427f552d91854e/dask_expr/_concat.py#L183

where a method of an expression is used to define a dask task. This inevitably requires us to serialize the instance itself which is something we should avoid. Ideally, the graph only includes global functions such that we can pickle by reference and not by value.

I don't know if there are other places where this pattern comes up.

related https://github.com/dask-contrib/dask-expr/pull/284

rjzamora:
>Just skimming the code, I saw stuff like this ... `[[self._meta, key], 0, self.join, False, True]` ... where a method of an expression is used to define a dask task

In this case, `self.join` is a parameter (not a method). So, we definitely shouldn't need to serialize the expression.

fjetter:
> In this case, self.join is a parameter (not a method). So, we definitely shouldn't need to serialize the expression.

Well, my mistake. Sorry. I haven't really tried to find out where this is happening, yet.

What's interesting is that for whatever objects this path is triggered, these objects have empty caches and according to my tests, `cached_property` is something that is preserved in pickle roundtrips

rjzamora:
>Well, my mistake. Sorry. I haven't really tried to find out where this is happening, yet. I mixed it up with DataFrame.join which is a method

No problem - I just wanted to put your mind at ease about that particular detail. I'm glad to see you investigating serialization behavior in general.

[...]

PR #324, opened 2023-10-09
Avoid ``operation`` methods in ``Blockwise`` subclasses

rjzamora:
Motivated by https://github.com/dask-contrib/dask-expr/issues/323

Modifies all cases where `Blockwise.operation` is currently overridden as a method in a `Blockwise` subclass. Instead, we need to use a property and/or `staticmethod`.

### DIFF
--- a/dask_expr/_reductions.py
+++ b/dask_expr/_reductions.py
@@ -35,8 +35,13 @@ class Chunk(Blockwise):
 
     _parameters = ["frame", "kind", "chunk", "chunk_kwargs"]
 
-    def operation(self, df, *args, **kwargs):
-        return self.chunk(df, *args, **kwargs)
+    @staticmethod
+    def chunk_operation(op, *args, **kwargs):
+        return op(*args, **kwargs)
+
+    @property
+    def operation(self):
+        return functools.partial(self.chunk_operation, self.chunk)
 
     @functools.cached_property
     def _args(self) -> list:
--- a/dask_expr/_accessor.py
+++ b/dask_expr/_accessor.py
@@ -72,7 +72,8 @@ class PropertyMap(Elemwise):
         "attr",
     ]
 
-    def operation(self, obj, accessor, attr):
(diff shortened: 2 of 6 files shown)


phofl (approved):

dask_expr/_reductions.py
```diff
@@ -35,8 +35,13 @@ class Chunk(Blockwise):
...
 
-    def operation(self, df, *args, **kwargs):
-        return self.chunk(df, *args, **kwargs)
+    @staticmethod
+    def chunk_operation(op, *args, **kwargs):
+        return op(*args, **kwargs)
+
+    @property
+    def operation(self):
+        return functools.partial(self.chunk_operation, self.chunk)
```
rjzamora, line 44: Hmm, I suppose we could just return `self.chunk`.

fjetter: I believe this is still problematic.

I suggest to add tests like I described in https://github.com/dask-contrib/dask-expr/issues/323#issuecomment-1753252862 to the project since catching this stuff in reviews alone is difficult

fjetter: The `partial` is still binding to `self.chunk_operation`, i.e. when serializing the `partial` object we have to serialize the entire instance

dask_expr/_reductions.py
```diff
@@ -35,8 +35,13 @@ class Chunk(Blockwise):
 
     _parameters = ["frame", "kind", "chunk", "chunk_kwargs"]
 
-    def operation(self, df, *args, **kwargs):
-        return self.chunk(df, *args, **kwargs)
+    @staticmethod
```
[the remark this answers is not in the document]
fjetter: yes, that's fine but I'm confused why this indirection is necessary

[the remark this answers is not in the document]
rjzamora: Yeah, it isn’t. That was my mistake

PR #326, opened 2023-10-10
Make ``simplify`` and ``lower`` optional within ``Expr.optimize``

rjzamora:
I have found that it can sometimes be useful to avoid lowering within an `Expr.optimize` call, because it becomes a bit easier to inspect the behavior of `simplify` on an expression graph.

### DIFF
--- a/dask_expr/_collection.py
+++ b/dask_expr/_collection.py
@@ -147,12 +147,16 @@ class FrameBase(DaskMethodsMixin):
             return new_collection(self.expr.__getitem__(other.expr))
         return new_collection(self.expr.__getitem__(other))
 
-    def persist(self, fuse=True, combine_similar=True, **kwargs):
-        out = self.optimize(combine_similar=combine_similar, fuse=fuse)
+    def persist(self, simplify=True, fuse=True, combine_similar=True, **kwargs):
+        out = self.optimize(
+            simplify=simplify, combine_similar=combine_similar, fuse=fuse
+        )
         return DaskMethodsMixin.persist(out, **kwargs)
 
-    def compute(self, fuse=True, combine_similar=True, **kwargs):
-        out = self.optimize(combine_similar=combine_similar, fuse=fuse)
+    def compute(self, simplify=True, fuse=True, combine_similar=True, **kwargs):
+        out = self.optimize(
+            simplify=simplify, combine_similar=combine_similar, fuse=fuse
+        )
         return DaskMethodsMixin.compute(out, **kwargs)
 
     def __dask_graph__(self):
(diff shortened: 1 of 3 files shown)


dask_expr/_collection.py
```diff
@@ -147,12 +147,16 @@ def __getitem__(self, other):
...
-        out = self.optimize(combine_similar=combine_similar, fuse=fuse)
+    def persist(self, simplify=True, fuse=True, combine_similar=True, **kwargs):
+        out = self.optimize(
+            lower=True, simplify=simplify, combine_similar=combine_similar, fuse=fuse
+        )
         return DaskMethodsMixin.persist(out, **kwargs)
 
-    def compute(self, fuse=True, combine_similar=True, **kwargs):
-        out = self.optimize(combine_similar=combine_similar, fuse=fuse)
+    def compute(self, simplify=True, fuse=True, combine_similar=True, **kwargs):
```
phofl, line 156: I don’t want to add unnecessary keywords here and this will work differently after #294, so I wouldn’t add control over simplify here

phofl:
Is there an advantage when running computations as well? This makes the logic noticeable more complex for a relatively small benefit.

We don't need everything configurable, I think we agree that there is really no point in not running simplify?

dask_expr/_collection.py
rjzamora: Yeah, the motivation isn't really to control options to `compute`, so I'm happy to roll that back. Thanks for pointing out 294.

rjzamora:
>Is there an advantage when running computations as well? This makes the logic noticeable more complex for a relatively small benefit.

I included that change because I was using it for other experiments and realized we would probably be interested in making it possible to the effects of column-projections etc in a more direct way. However, `compute` is definitely not my primary interest here, so I'll roll that part back.

>We don't need everything configurable, I think we agree that there is really no point in not running simplify?

Well, we **must** run simplify to guarantee  that a low-level graph can be generated. We don't really need to simplify, but it's fine with me if we keep that component simple.

phofl:
I am a bit uncomfortable with making ``def optimise`` more complicated.

Couldn't we use ``.simplify().combine_similar()`` instead of adjusting optimise?

rjzamora:
>I am a bit uncomfortable with making def optimise more complicated.
>
>Couldn't we use .simplify().combine_similar() instead of adjusting optimise?

Sure, that seems fair to me. This PR is a much lower priority than https://github.com/dask-contrib/dask-expr/pull/321, so I have no problem closing it.

I'll admit that I was not being completely transparent about my full motivations in my PR description. I was hoping to slowly nudge the library in the direction of establishing a clearer distinction between "abstract" and "dask-specific" expressions.

I was thinking that this would be simpler if `lower` was optional, but the real blocker is probably the fact that `Expr` is currently required to support dask-specific attributes like `divisions` and `npartitions`. I am starting to feel that this should **not** be the case at the level of the base `Expr` class. However, this is obviously a big change (and is somewhat orthogonal to the original goals of this work), so I'll just plan to revisit these thoughts later :)

PR #329, opened 2023-10-11
Avoid divisions calculation for one partition df in sort_values

phofl:
This came up in tpch, there is no need to trigger a divisions computation if we are reduced to one partition anyway


phofl:
I am merging this to include it in the release

rjzamora (commented):
Seems fine to me - Sorry for missing this. There are probably other cases where we can modify `_lower` for the special case of one partition (e.g. groupby aggregations).

PR #330, opened 2023-10-12
Make repartition accept a callable for npartitions

phofl:
This is something we need after the IO fusion PR. 

Something like ``df.repartition(npartitions=df.npartitions // 2)`` might raise now because we might have squashed the partitions even more in the I/O layers, but you can do ``df.repartition(npartitions=lambda x: x // 2)`` to avoid this problem.

### DIFF
--- a/dask_expr/_collection.py
+++ b/dask_expr/_collection.py
@@ -415,20 +415,28 @@ class FrameBase(DaskMethodsMixin):
     ):
         """Repartition a collection
 
-        Exactly one of `divisions` or `npartitions` should be specified.
-        A ``ValueError`` will be raised when that is not the case.
+        Exactly one of `divisions`, `npartitions` or `partition_size` should be
+        specified. A ``ValueError`` will be raised when that is not the case.
 
         Parameters
         ----------
-        npartitions : int, optional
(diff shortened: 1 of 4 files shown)


rjzamora (commented):
Left a few comments, but I think this solution is reasonable for now.

I think my only reservation is that we still don't want people using `df.npartitions` within the logic of their workflow (like `df.repartition(npartitions=df.npartitions // 2)`), but this is still not clear in the documentation/behavior.

I'm not expecting you to solve this problem here, but I do wonder if you have thoughts. For example, maybe we should state in the doc-string that an `int` should **not** be passed to `npartitions` for the purpose of directly increasing or. decreasing the current partition count by a specific factor, because the input partition count may change during optimization. Instead, users should pass in a `Callable` (e.g. `lambda x: x // 2`) for this purpose (assuming such behavior is necessary).

phofl:
My thoughts are that we should be smart enough to figure most of these things out on the fly so that users won't have to worry about partitions much, but we are still quite a bit away from this. 

In the meantime I am happy with advertising the callable more and explicitly pointing folks away from df.npartitions usage

PR #331, opened 2023-10-12
Add custom serialization check to ``assert_eq``

### LINKED ISSUE #323: `Expr` instances are serialized as part of the graph

rjzamora:
Closes #323

This PR illustrates one possible way to test that we are not including `Expr` object instances in the result of `Expr.dask`. Note that this change helped me find and fix another bug related to #323 (groupby `Var`).

I didn't think very hard about this solution, so suggestions are welcome.

Note that the tests take about 15% longer on my laptop with this change.


mrocklin:
> Note that the tests take about 15% longer on my laptop with this change

Would you want to see this merged in with this cost difference?  This seems significant to me, and this check seems useful, but not super-important.  Do you have thoughts here?

rjzamora:
>Would you want to see this merged in with this cost difference? This seems significant to me, and this check seems useful, but not super-important. Do you have thoughts here?

This pytest runtime regression is the main reason the PR is marked as draft. I would be much happier if we were adding the equivalent of a few tests rather than ~40 new tests. It may make sense to "opt in" to the serialization test in a few places, but I'm still unsure.

mrocklin:
Unfortunately I think that you probably want ~40 new tests because you want this tested for every expression today, and also in the future.  The `assert_eq` style checks have historically been quite good at ensuring this kind of hygeine throughout the project (today if you make a new method in dask.dataframe and have badly named keys for example, you get an informative error, even though you didn't write a new test for your new method on this topic).  Ideally though we find some way to make these sorts of checks very cheap to run (this is sometimes hard).  Maybe it's not possible here though.  I don't know.

phofl:
How long is the test suite running on your machine? It take 12 seconds on mine, so 15% doesn't really matter at this point

rjzamora:
>Unfortunately I think that you probably want ~40 new tests because you want this tested for every expression today, and also in the future.

This is probably  true.

>How long is the test suite running on your machine? It take 12 seconds on mine, so 15% doesn't really matter at this point

It's also <20s on mine, so I agree that a couple seconds is not a big deal. I think the consideration here is more-so: "Are we okay with all new tests taking about 15% longer from here on out?" The answer may just be "yes".

dask_expr/tests/_util.py
```diff
@@ -16,3 +17,22 @@ def xfail_gpu(reason=None):
...
+        # Check that no `Expr` instances are found in
+        # the graph generated by `Expr.dask`
+        try:
+            from distributed.protocol import serialize
+            from distributed.protocol.serialize import ToPickle
+
+            with config.set({"dask-expr-no-serialize": True}):
+                for obj in [a, b]:
+                    if hasattr(obj, "dask"):
+                        serialize(ToPickle(obj.dask))
```
fjetter, line 33: Have you tried plain pickle? While this is closer to what is happening in reality, plain pickle is likely faster for this case and tests what we're looking for

rjzamora: We will currently see errors for local functions (like `lambda`s), and the run time doesn't seem to change much.

rjzamora: Actually, it probably does make sense to try using `pickle.dumps` before falling back to `serialize(ToPickle(obj.dask))`. That does seem to take us below a 10% runtime difference.

fjetter: well, there is a point why this raises with lambdas :) pickle cannot deal with lambda so we'll have to fallback to cloudpickle which is more often than not slower than plain pickle because it again requires pickling by value instead of reference (i.e. it has to serialize the code of the function instead of just a pointer)

I guess this is a different topic but if dask-expr passes a plain pickle test, you'll be faster in the long run.

For instance...

```python
In [1]: import pickle, cloudpickle

In [2]: def f(x):
   ...:     return x
   ...:

In [3]: len(pickle.dumps(f))
Out[3]: 29

In [4]: len(cloudpickle.dumps(lambda x: x))
Out[4]: 455
```

when the thing needs to be pickled by value (cloudpickle) it has to write much more garbage to be able to reconstruct it. This is particularly bad if the lambda references something from the local scope since then the entire local scope is serialized.

For large graphs, this stuff can add up quickly

mrocklin: Definitely good hygiene to avoid things that need cloudpickle in Dask.  This can become costly at large numbers of partitions.

rjzamora: Yeah, I guess there are a few places where we were still using `lambda`s in places that we shouldn't be (I found and fixed some of those cases), but the main reason we need to fall back to `serialize` or `cloudpickle` in `assert_eq` is that we use `lambda`s in the pytests.

rjzamora:
>I'd be ok with merging this

I also agree that the small overhead is worth the mistakes it helps us avoid :)

phofl:
thx @rjzamora

PR #333, opened 2023-10-13
Npartitions are not updated during partitioning for merge

phofl:
We have to keep those 2 steps together, the alternative makes things very complicated and ugly

### DIFF
--- a/dask_expr/_merge.py
+++ b/dask_expr/_merge.py
@@ -157,13 +157,11 @@ class Merge(Expr):
             and get_default_shuffle_method() == "p2p"
         ):
             left = AssignPartitioningIndex(
-                left, shuffle_left_on, _HASH_COLUMN_NAME, self.npartitions
+                left, shuffle_left_on, _HASH_COLUMN_NAME, lambda x: x, right
             )
             right = AssignPartitioningIndex(
-                right, shuffle_right_on, _HASH_COLUMN_NAME, self.npartitions
+                right, shuffle_right_on, _HASH_COLUMN_NAME, lambda x: x, left
             )
-            left = Repartition(left, lambda x: x // 2)
-            right = Repartition(right, lambda x: x // 2)
 
             return HashJoinP2P(
                 left,
--- a/dask_expr/_merge.py
+++ b/dask_expr/_merge.py
@@ -1,7 +1,9 @@
 import functools
 
+import pandas as pd
 from dask.core import flatten
 from dask.dataframe.dispatch import make_meta, meta_nonempty
+from dask.dataframe.shuffle import partitioning_index
 from dask.utils import M, apply, get_default_shuffle_method
 
 from dask_expr._expr import (
@@ -13,7 +15,7 @@ from dask_expr._expr import (
     Projection,
 )
 from dask_expr._repartition import Repartition
-from dask_expr._shuffle import AssignPartitioningIndex, Shuffle, _contains_index_name
+from dask_expr._shuffle import Shuffle, _contains_index_name, _select_columns_or_index
 from dask_expr._util import _convert_to_list
 
 _HASH_COLUMN_NAME = "__hash_partition"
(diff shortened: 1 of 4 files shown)


dask_expr/_merge.py
```diff
@@ -408,6 +402,35 @@ def _simplify_up(self, parent):
         return
 
 
+def create_assign_index_merge_transfer():
```
rjzamora, line 405: I'm not sure I understand the benefit of this function. Is there a reason we cannot just use `assign_index_merge_transfer` directly (with the `pd` and `merge_transfer` imports moved into that function)?

dask_expr/io/tests/test_distributed.py
```diff
@@ -0,0 +1,31 @@
...
+def test_io_fusion_merge(tmpdir):
+    pdf = lib.DataFrame({c: range(100) for c in "abcdefghij"})
+    with LocalCluster(processes=False, n_workers=2) as cluster:
+        with Client(cluster) as client:  # noqa: F841
+            dx.from_pandas(pdf, 10).to_parquet(tmpdir)
+            df = dx.read_parquet(tmpdir).merge(
+                dx.read_parquet(tmpdir).add_suffix("_x"), left_on="a", right_on="a_x"
+            )[["a_x", "b_x", "b"]]
+            out = df.compute()
+    lib.testing.assert_frame_equal(
```
rjzamora, line 26: Can we use `assert_eq`?

rjzamora (commented):
Pretty sure the change makes sense. Gave a quick look and only have some minor questions so far.

dask_expr/_merge.py
phofl: I don't want to do those imports on every function call, pandas isn't cheap for example.

It shouldn't matter, but I'd rather not take that chance. Also, I'd rather fail at graph construction time then during computation time if dependencies are messed up

dask_expr/io/tests/test_distributed.py
phofl: That doesn't play welll here unfortunately.

rjzamora: Can you clarify? I think it would be best to add a comment describing why you are using `assert_frame_equal` here, since not being able to use `assert_eq` is often a signal that something is wrong.

dask_expr/_merge.py
rjzamora: Okay, you only need to import `merge_transfer`, but I understand the interest in raising an error in `_layer`.

Can we just move this function to be a property of `merge_transfer` then?  For some reason, the `func = create_assign_index_merge_transfer()` pattern feels off to me. Sorry for nitpicking.

phofl: This is a pretty common pattern in more functional environments 

I don't understand what you mean with property of merge_transfer

dask_expr/io/tests/test_distributed.py
phofl: This is a pretty common pattern in distributed, so I would rather not add a comment for this.

We can use assert_eq in theory, but we would effectively pass 2 pandas DataFrames, so this seems a bit pointless

rjzamora: Yeah, I can understand that reasoning, but I'm still interested in understanding **why** (or at least **how**) `assert_eq` fails. I certainly don't think it is critical that `assert_eq` is used everywhere, but I'd still like to learn something about this.

phofl: P2P shuffle just flat out fails

It's the same in distributed and when I use Dask DataFrame instead of dask_expr.

I don't intend to look deeper into this.

dask_expr/_merge.py
rjzamora: >This is a pretty common pattern in more functional environments

Yeah I agree, but this just doesn't feel like a common pattern in dask/dask or dask-expr. I don't feel super strong about this, just trying to consider ways to keep style consistent throughout.

>I don't understand what you mean with property of merge_transfer

Oops! That was a typo. I was just saying that `create_assign_index_merge_transfer` could be a property of `HashJoinP2P`, but I don't love that pattern either (it's just a bit closer to what I'd expect).

I'll admit that I don't actually care much about this, so feel free to treat this comment thread as "non-blocking".

dask_expr/io/tests/test_distributed.py
rjzamora: Okay, chatted with Patrick offline, and the issue is that we need to compute before comparing for some reason related to p2p, and so we are relying on the stricter testing in `assert_frame_equal` given that we don't have collections anymore. That works for me.

PR #337, opened 2023-10-16
Fix column projection in merge

phofl:
This wasn't caught before, we weren't pushing them back up correctly if one branch was already correct.


phofl:
This Is the last one that I am aware of

PR #338, opened 2023-10-17
Make Partitions accept a callable for partitions

phofl:
cc @rjzamora thoughts here?

Two reasons for doing this: npartitions might change that's why we have to add the callable support

We have to get partitions in there early if we want to move to the end of our optimisation passes. Do you have a better idea of achieving this?

### DIFF
--- a/dask_expr/_expr.py
+++ b/dask_expr/_expr.py
@@ -6,6 +6,7 @@ import operator
 import os
 from collections import defaultdict
 from collections.abc import Generator, Mapping
+from typing import Callable
 
 import dask
 import numpy as np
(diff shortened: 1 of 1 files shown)


rjzamora (approved):
I'm still thinking about this one. However, I do think it is worth mentioning that things are a bit simpler if we introduce a distinct `LoweredExpr` type for classes that are dask-specific (like `Partitions`).

For example, I seem to have all test passing with changes like this:

```python
class LoweredExpr(Expr):
    pass

...

class Partitions(LoweredExpr):
...

def simplify(self, restrict: None | type = None):
    # Only allow simplify on a node when isinstance(self, restrict).
    # Otherwise, move on to dependencies (with `restrict=restrict`).
...
    
def optimize(expr: Expr, combine_similar: bool = True, fuse: bool = True) -> Expr:

    # "Abstract" phase
    result = expr
    result = result.simplify()
    if combine_similar:
        result = result.combine_similar()

    result = result.lower_completely()

    # "Lowered" phase
    result = result.simplify(restrict=LoweredExpr)
    if combine_similar:
        result = result.combine_similar()

    # Final graph-specific optimizations
    if fuse:
        result = optimize_io_fusion(result)
        result = optimize_blockwise_fusion(result)

    return result
```

phofl:
not needed anymore

PR #340, opened 2023-10-17
Fix combine_similar with repartition in between

phofl:


dask_expr/_shuffle.py
```diff
@@ -619,6 +626,18 @@ def operation(df, index, name: str, npartitions: int):
             index = partitioning_index(index, npartitions)
         return df.assign(**{name: index})
 
+    def _combine_similar(self, root: Expr):
+        return self._combine_similar_branches(root, (Filter, Projection))
+
+    def _remove_operations(self, frame, remove_ops, skip_ops=None):
+        expr, ops = super()._remove_operations(frame, remove_ops, skip_ops)
+        if len(ops) > 0 and isinstance(ops[0], list):
+            if sorted(ops[0]) == sorted(self.frame.columns):
```
phofl, line 635: I don't like that part very much, but will revisit after we re-ordered the optimisation logic, this should make this a lot easier

dask_expr/_expr.py
```diff
@@ -461,23 +461,23 @@ def _combine_similar_branches(self, root, remove_ops, skip_ops=None):
             return None
 
         others = self._find_similar_operations(root, ignore=self._parameters)
-        if isinstance(self.frame, Filter) and all(
-            isinstance(op.frame, Filter) for op in others
-        ):
-            # Avoid pushing filters up if all similar ops
-            # are acting on a Filter-based expression anyway
-            return None
```
phofl, line 469: Doing this early doesn't make too much sense, we should only do this for ops that are potentially compatible with self.

phofl:
merging this for benchmarking purposes, happy to re-iterate as a follow up

PR #341, opened 2023-10-18
Update Fused keynames

mrocklin:
1.  Remove "Fused" from the beginning of the name.  This is mostly an internal detail that we care about, not something that most users care about I think.

2.  Sort expressions by depth, so that the first operation (like readparquet) is first and the last operation (like sum) is last.

    **Warning**, the `_depth` method is currently not very efficient

3.  Keep fused in the name, but now in the middle, where it's a bit less prominent

cc @phofl thoughts?

### DIFF
--- a/dask_expr/_expr.py
+++ b/dask_expr/_expr.py
@@ -2453,11 +2453,12 @@ class Fused(Blockwise):
         return lines
 
     def __str__(self):
-        names = [expr._name.split("-")[0] for expr in self.exprs]
-        if len(names) > 3:
-            names = [names[0], f"{len(names) - 2}", names[-1]]
-        descr = "-".join(names)
-        return f"Fused-{descr}"
+        exprs = sorted(self.exprs, key=M._depth)
+        names = [expr._name.split("-")[0] for expr in exprs]
+        if len(names) > 4:
+            return names[0] + "-fused-" + names[-1]
+        else:
+            return "-".join(names)
 
     @functools.cached_property
     def _name(self):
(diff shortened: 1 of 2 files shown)


dask_expr/_expr.py
```diff
@@ -2453,11 +2453,12 @@ def _tree_repr_lines(self, indent=0, recursive=True):
...
     def __str__(self):
-        names = [expr._name.split("-")[0] for expr in self.exprs]
-        if len(names) > 3:
-            names = [names[0], f"{len(names) - 2}", names[-1]]
-        descr = "-".join(names)
-        return f"Fused-{descr}"
+        exprs = sorted(self.exprs, key=M._depth)
+        names = [expr._name.split("-")[0] for expr in exprs]
+        if len(names) > 4:
+            return names[0] + "-fused-" + names[-1]
```
fjetter, line 2459: Whatever the final policy here is, please ensure that the final key is truly unique since we'll be running again into https://github.com/dask/dask/issues/9888 otherwise (blockwise fusion is reusing keys which the distributed scheduler cannot handle)

This section looks like there is plenty of room for ambiguity but I don't know the internals well enough to judge. Just raising this for context

mrocklin: To be clear, I'm guessing that it's ok for the key prefix to be non-unique, as long as the key itself is unique.  Is that correct?  In other words, we should be able to handle this in the hash, right?

rjzamora: >I'm guessing that it's ok for the key prefix to be non-unique, as long as the key itself is unique

Yes. I certainly hope this is the case (otherwise we are bound to run into problems).

mrocklin:
Merging tomorrow morning my time if no one comments before then.

phofl:
I'll take a look tomorrow, didn't get to it today

rjzamora (approved):
Better than what we were doing before, and LGTM - Thanks @mrocklin

phofl (approved):

phofl:
thx @mrocklin

PR #345, opened 2023-10-19
Get rid of SettingWithCopyWarnings

### LINKED ISSUE #342: Copy on write warning
I get these warnings sometimes:

```
/Users/mrocklin/workspace/dask-expr/dask_expr/_merge.py:516: SettingWithCopyWarning:
A value is trying to be set on a copy of a slice from a DataFrame.
Try using .loc[row_indexer,col_indexer] = value instead

See the caveats in the documentation: https://pandas.pydata.org/pandas-docs/stable/user_guide/indexing.html#returning-a-view-versus-a-copy
  df[name] = index
```

Do we know anyone who knows about pandas' copy-on-write functionality ?

phofl:
cc @mrocklin 
this should get rid of the warnings

closes #342

### DIFF
--- a/dask_expr/_merge.py
+++ b/dask_expr/_merge.py
@@ -513,8 +513,8 @@ def create_assign_index_merge_transfer():
             index = partitioning_index(df[index], npartitions)
         else:
             index = partitioning_index(index, npartitions)
-        df[name] = index
-        meta[name] = 0
+        df = df.assign(**{name: index})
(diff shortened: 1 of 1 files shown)


mrocklin:
And I guess this no longer negatively impacts performance due to copy-on-write?

phofl:
It does, we haven't enabled Copy on write here.

PR #348, opened 2023-10-19
Adapt for dask changes in shuffle

phofl:
CI coverage for the released version is something we should have added already

### DIFF
--- a/ci/environment_released.yml
+++ b/ci/environment_released.yml
@@ -4,7 +4,7 @@ channels:
 dependencies:
   - pytest
   - pytest-cov
-  - dask  # overridden by git tip below
+  - dask
   - pyarrow
   - pandas>=2
   - pre-commit
--- a/ci/environment_released.yml
+++ b/ci/environment_released.yml
@@ -1,0 +1,10 @@
+name: dask-expr
+channels:
+  - conda-forge
+dependencies:
+  - pytest
+  - pytest-cov
+  - dask  # overridden by git tip below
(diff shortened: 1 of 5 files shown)


phofl:
cc @mrocklin for a quick review maybe?

For context: Rick and I agreed that we want to support the latest released version and main for the time being. We will remove this special casing as soon as the next dask release is available and bump the minimum version in the next dask-expr release.

ci/environment_released.yml
```diff
@@ -0,0 +1,10 @@
+name: dask-expr
+channels:
+  - conda-forge
+dependencies:
+  - pytest
+  - pytest-cov
+  - dask  # overridden by git tip below
```
mrocklin, line 7:
```suggestion
  - dask
```

dask_expr/_merge.py
```diff
@@ -515,7 +515,10 @@ def assign_index_merge_transfer(
             index = partitioning_index(index, npartitions)
         df[name] = index
         meta[name] = 0
-        return merge_transfer(df, id, input_partition, npartitions, meta, parts_out)
+        disk = [] if not DASK_GT_20231000 else (True,)
```
mrocklin, line 518:
```suggestion
        disk = () if not DASK_GT_20231000 else (True,)
```

Just me being pedantic.   My apologies.

mrocklin (approved):
Sure.  Seems fine to me.  Two (very trivial) suggestions below.

PR #350, opened 2023-10-19
Adjust key names for FusedIO class

phofl:


dask_expr/io/io.py
```diff
@@ -129,6 +130,16 @@ def _combine_similar(self, root: Expr):
 class FusedIO(BlockwiseIO):
     _parameters = ["expr"]
 
+    @functools.cached_property
+    def _name(self):
+        return (
+            funcname(type(self.operand("expr"))).lower()
+            + "-"
+            + funcname(type(self)).lower()
```
mrocklin, line 138: Maybe just `fused` rather than `fusedio` here?  (unless this is commonly subclassed)

phofl: don't see a need for subclassing at the moment, I adjusted it

PR #248, opened 2023-07-28
[WIP] Support ``split_out>1`` and ``sort=True`` for aggregations

rjzamora:
Introduces a shuffle-based reduction algorithm to support ``split_out>1`` and ``sort=True``. 

(**Note**: Not sure if this is "ready" for eyes, but I wanted to push it somewhere visible as early as possible.)

PR #351, opened 2023-10-19
Support `split_out` for reductions using shuffling

rjzamora:
Introduces a new `ShuffleReduce` expression to support `split_out>1` and `sort=True` within groupby aggregations. This PR also adds support for `split_out > 1` for `drop_duplicates`.

**Notes**:
- This PR introduces the convention that `split_out=True` means that the partition count will be preserved.
- This PR  supersedes #248
- It seems reasonable to use `split_out=True` as the default for `drop_duplicates`. However, the shuffle-algorithm does not preserve the original ordering of unique rows. It is unclear to me if this will be problematic for users?

### DIFF
--- a/dask_expr/_reductions.py
+++ b/dask_expr/_reductions.py
@@ -20,7 +20,15 @@ from dask.dataframe.core import (
 from dask.utils import M, apply, funcname
 
 from dask_expr._concat import Concat
-from dask_expr._expr import Blockwise, Expr, Index, Projection, RenameFrame, ToFrame
+from dask_expr._expr import (
+    Blockwise,
+    Expr,
+    Index,
+    Projection,
+    RenameFrame,
+    ResetIndex,
+    ToFrame,
+)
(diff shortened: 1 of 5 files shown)


dask_expr/_reductions.py
```diff
@@ -240,6 +310,129 @@ def _tree_repr_lines(self, indent=0, recursive=True):
         return lines
 
 
+class ApplyConcatApply(Expr):
```
phofl, line 313: Did anything change here or is it just at a different place?

phofl (commented):
Looks good generally, I'll take a closer look on Monday and will probably merge

dask_expr/_reductions.py
rjzamora: Ah, I move things around, because I "needed" to at one point. Sorry about that. I don't need this re-ordering anymore, so I might try to move it back to reduce the diff.

PR #356, opened 2023-10-23
Allow `pandas-1.5.3` but warn the user on import

rjzamora:
I've been running into some pain while setting up an environment to test/benchmark dask-expr + cudf. Most of this pain is related to the fact that dask-expr is pinned to `pandas>=2`. 

This PR proposes that we adjust the pin to `pandas>=1.5.3`, but warn the user (on import) if the version is `<2.0.0`. E.g.

```
UserWarning: The installed Pandas version is not recommended for `dask_expr`. Please update to `pandas>=2` if possible.
  RAPIDS users can ignore this warning.
```

My hope is that users will only pick up `pandas<2` when they are building an environment with `cudf`, but I like having the warning "just in case".

### DIFF
--- a/dask_expr/__init__.py
+++ b/dask_expr/__init__.py
@@ -2,3 +2,12 @@ from dask_expr import _version, datasets
 from dask_expr._collection import *
 
 __version__ = _version.get_versions()["version"]
+
+from dask.dataframe._compat import PANDAS_GE_200
+
+if not PANDAS_GE_200:
+    warnings.warn(
+        "The installed Pandas version is not recommended for "
+        "`dask_expr`. Please update to `pandas>=2` if possible."
+        "\nRAPIDS users can ignore this warning."
+    )
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -27,7 +27,7 @@ requires-python = ">=3.9"
 dependencies = [
     "dask >= 2023.09.2",
     "pyarrow",
-    "pandas >= 2",
+    "pandas >= 1.5.3",
 ]
 
 dynamic = ["version"]


mrocklin:
This seems suboptimal to me.  Thoughts on making a `dask-expr[rapids]` target instead?

rjzamora:
Sure - That works for me 👍

rjzamora:
>Thoughts on making a dask-expr[rapids] target instead?

Is it possible to define optional dependencies that conflict with the *base* dependencies? I'm realizing that I don't actually know how to properly define a target like this, so suggestions/references are welcome :)

mrocklin:
I don't know.  I rarely handle targets like this today. My knowledge here
is dated.  In this situation I might ask on Stack Overflow?

phofl:
Same here. @graingert might know

Doing a separate Conda package would work

rjzamora:
I've been discussing this with @charlesbluca a bit. So far, our understanding is that a base dependency (e.g. `pandas>=2`) cannot conflict with an optional dependency (e.g. `pandas<=1.5.3`). (we are still looking into this, but I have a feeling that there isn't much we can do about that).

So far, I'm most confident in Charles` suggestion that we:

- loosen the base dependency to `pandas<=1.5.3` (since we do know this "works", and is required for cudf)
- Add a `dask-expr[strict]` target with a `pandas>=2` dependency. This would be used for testing, packaging, and would be used in the installation instructions.
- Add a `dask-expr[rapids]` target with a `pandas==1.5.3` dependency. This would only be used by RAPIDS people wanting to use dask-expr with cudf.

mrocklin:
I'm a strong -1 on requiring pandas users to type in something more complex.  My experience is that most users don't read instructions well enough to do that well.

rjzamora:
>I'm a strong -1 on requiring pandas users to type in something more complex.

Fair enough. The user currently needs to do this to install `dask.dataframe` today, so I was hoping something like that could help me here.

I suppose it's probably "okay" to tell RAPIDS users to pip to install `dask_expr` with a `--no-deps` flag for now.

Somewhat related: There is a good chance that Pandas>=2 support will not land in cudf until February or April. We have some tricky short-term blockers/priorities, and most of our users are still on python 3.8 and/or pandas<2 for now.  I don't want any non-RAPIDS users installing dask-expr with pandas-1 between now and then, but I will certainly need to continue doing this to use dask-expr + cudf.

phofl:
Just a FYI: pandas 3.0 is expected in April next year

mrocklin:
Maybe the increased performance of dask-cudf will help push people towards more modern versions?

rjzamora:
>Just a FYI: pandas 3.0 is expected in April next year

Yeah, I bring this up often :)
My understanding is that addressing pandas-2 changes in cudf is significantly more involved than addressing the the move to pandas-3.

Unfortunately, the pandas-2 question is not really something RAPIDS engineers have any say over at this point. If dask or dask-expr decides to completely drop pandas<2 support in the first half of the coming year, people like me will just need to waste a lot of time figuring out ugly workarounds.

rjzamora:
>Maybe the increased performance of dask-cudf will help push people towards more modern versions?

So far the numbers look promising - Even without overriding any groupby aggregations for better GPU utilization :)
There is a lot of interest in dask-cudf + cudf, but we don't have the engineering cycles to accelerate pandas-2 compatibility.

mrocklin:
> If dask or dask-expr decides to completely drop pandas<2 support in the first half of the coming year, people like me will just need to waste a lot of time figuring out ugly workarounds

My condolences.  However, it sounds like this is something that you all should probably start preparing for soon.

rjzamora:
>My condolences. However, it sounds like this is something that you all should probably start preparing for soon.

(By first half of the coming year, I really mean before we can pin for the April release). I do think it is reasonable for dask to phase out python-3.8 support beyond April 2024, at which point it could also make sense to phase out pandas<2.

@jrbourbeau @fjetter @phofl - I am definitely asking (begging?) that dask/dask **not** drop pandas<2 support before April. Please let me know if there is anything I can do to take the pandas<2 maintenance burden upon myself (I can raise an issue in dask/dask if you think that makes sense). I do recognize that pandas>=2 has wonderful performance advantages over pandas<2, and that you don't want to worry about introducing performance regressions for older versions - this is not lost on me at all.

PR #358, opened 2023-10-25
Reorder optimization logic simplify-lower-cull

phofl:
sorry @rjzamora, I have trouble pushing to other forks and this needed a small fix


dask_expr/_expr.py
```diff
@@ -256,41 +257,44 @@ def simplify(self):
...
-                return out
-            if out._name != expr._name:
-                expr = out
-                continue
-
-            # Allow children to simplify their parents
-            for child in expr.dependencies():
-                out = child._simplify_up(expr)
+            # Rewrite this node
+            if down_name in expr.__dir__():
```
rjzamora, line 266: This was something I wasn't sure about. This approach takes extensibility to a bit of an extreme, because we do not require a new "kind" of rewrite to be registered with the base `Expr` class in any way. Instead, a user can just define methods like `_foo_down` and/or `_foo_up`, and `Expr.rewrite("foo")` will run without complaining.

This approach makes things easy at this point. However, we may want to make it more difficult to silently ignore bugs (e.g. `Expr.rewrite("fool")  # This will do nothing, and won't complain`. With that said, I also get the feeling that the "loose" design used here won't be difficult to revise when we are ready for something more formal.

phofl: > With that said, I also get the feeling that the "loose" design used here won't be difficult to revise when we are ready for something more formal.

Yeah that was what pushed me to keep this as well, this is easy to enforce as long as we don't have third party expressions

rjzamora: >this is easy to enforce as long as we don't have third party expressions

Yeah exactly. Should we use `Expr._rewrite` instead of `Expr.rewrite` to emphasize that this is an internal API that is likely to change?

rjzamora (approved):
My review is obviously biased, but do I think the approach makes sense for now :)

dask_expr/_expr.py
phofl: I would keep it public for now, feels a bit better to me

phofl:
Let's get this in

PR #359, opened 2023-10-25
Implement io fusion as a rewrite layer

phofl:
sits on top of #358


dask_expr/_expr.py
```diff
@@ -2190,18 +2215,24 @@ def optimize(expr: Expr, combine_similar: bool = True, fuse: bool = True) -> Exp
...
-        result = out
+    # Simplify
+    result = expr.simplify()
 
+    # Combine similar
     if combine_similar:
         result = result.combine_similar()
 
+    # Manipulate Expression to make it more efficient
+    result = result.rewrite(kind="tune")
```
rjzamora, line 2226: Nice.

I find it interesting that the "tune" step comes before `lower_completely`, given that this kind of optimization is related to the partition count. My understanding is that we **need** to do this after `combine_similar`. Do we also **need** to do this before `lower_completely`? Or, is the idea that you might as well aggregate partitions before the expression graph gets larger?

Note that I don't mind the ordering you have here, but I'm slightly worried if it **needs** to be that way for some reason.

phofl: It definitely needs to be in that order. Tune (open for other names btw) changes the number of partitions while lower locks the number of partitions in. We run into all kinds of trouble if we lock them in and then try to change them afterwards

rjzamora: Ah okay, that makes sense! Sorry for the confusion.

Personal Notes: This optimization lives in that "fuzzy zone" where you don't really care if you have "lowered" expressions, but you cannot have "frozen" partitioning anywhere.  Right now, a "lowered" expression is just one that can directly generate a task graph, which is slightly different from meaning that it has frozen partitions. However, we have certainly avoided frozen partitioning outside of lowered expressions so far (so this ordering makes sense).

phofl:
I agree with the sentiment, Ideally I want to work towards freezing
divisions even less, but there is another reason why we "need" (not
strictly) this order

We want to reorder joins at some point and this is significantly easier if
a join is one Merge Expression instead of a AssignPartioningIndex, Shuffle
and Blockwise Merge. Doing these things with abstract expressions should be
easier in the long run

rjzamora:
>I agree with the sentiment, Ideally I want to work towards freezing
divisions even less, but there is another reason why we "need" (not
strictly) this order...

Absolutely agree that we want to do join reordering before lowering. I think we both had that in mind as a motivation for something like https://github.com/dask-contrib/dask-expr/pull/358.

My earlier confusion about the ordering of "tune" was just a mistake on my part, and mostly unrelated - Sorry about that.

phofl:
Just a general note: I think this won't fit into rewrite, this is more of a combine_similar type of thing where you need a global view of the expression, but I would like to start with this before we make mit more complex.

rjzamora:
>I think this won't fit into rewrite

By "this", do you mean join re-ordering (in which case I probably agree), or do you mean IO aggregation?

phofl:
Both things, the IO fusion is pretty dumb at the moment.

Let's assume the following case:

You only read 50% of your columns, then there are a couple of blockwise ops that drop another 50% of the remaining columns. Blockwise ops won't create memory spikes, so I/O fusion could operate with a factor of 4 instead of 2, but we would have to look at the whole graph.

I just want to get this to the place where it needs to be (mostly for testing purposes) and then iterate when we want to make more complicated decisions

rjzamora:
**Disclaimer**: None of my questions above are blockers for this PR (which LGTM).  I'm just interested in your thoughts :)

PR #360, opened 2023-10-25
Ensure unique shuffle IDs in hash join

hendrikmakait:
P2P shuffling requires shuffle IDs to be unique. This PR makes sure that they are unique even if dataframes get reused in the same or several subsequent shuffles with different parameters.

Fixes dask/distributed#8301

### DIFF
--- a/dask_expr/_merge.py
+++ b/dask_expr/_merge.py
@@ -428,52 +428,56 @@ class HashJoinP2P(Merge, PartitionsFiltered):
         from distributed.shuffle._shuffle import shuffle_barrier
 
         dsk = {}
-        name_left = "hash-join-transfer-" + self.left._name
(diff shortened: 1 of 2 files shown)


rjzamora (commented):
Change looks good to me. Any ideas for a simple test? Perhaps you can just generate the graph twice (e.g. `df1.merge(df2, ...).dask`) and compare keys?

phofl (approved):

phofl:
Some context: 

@hendrikmakait and I looked into this. The fix is valid and something I missed when initially implementing the HashJoin layer, but it should not have been necessary where this started failing (query 7). The optimization messes something up when we have more than one row group per file. I will look into this.

This will also get a follow up that is more precise about the left and right keys (similar to what is currently in distributed)

phofl:
thx @hendrikmakait

PR #361, opened 2023-10-26
Reuse identical shuffles in P2P hash join

hendrikmakait:
IFF a shuffle has the exact same input and parameters, P2P hash join should reuse it.

Follow-up to #360

### DIFF
--- a/dask_expr/_merge.py
+++ b/dask_expr/_merge.py
@@ -429,10 +429,18 @@ class HashJoinP2P(Merge, PartitionsFiltered):
 
         dsk = {}
         token_left = _tokenize_deterministic(
-            self.left._name, self.shuffle_left_on, self.npartitions, self._partitions
+            "hash-join",
+            self.left._name,
+            self.shuffle_left_on,
+            self.npartitions,
+            self._partitions,
         )
         token_right = _tokenize_deterministic(
-            self.right._name, self.shuffle_right_on, self.npartitions, self._partitions
+            "hash-join",
+            self.right._name,
+            self.shuffle_right_on,
+            self.npartitions,
+            self._partitions,
         )
         _barrier_key_left = barrier_key(ShuffleId(token_left))
         _barrier_key_right = barrier_key(ShuffleId(token_right))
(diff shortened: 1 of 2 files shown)


dask_expr/_merge.py
```diff
@@ -428,9 +428,12 @@ def _layer(self) -> dict:
         from distributed.shuffle._shuffle import shuffle_barrier
 
         dsk = {}
-        token = self._name.split("-")[-1]
-        token_left = token + "-left"
-        token_right = token + "-right"
+        token_left = _tokenize_deterministic(
+            self.left._name, self.shuffle_left_on, self.npartitions, self._partitions
```
phofl, line 432: I assume how doesn't matter in the shuffle algorithm?

hendrikmakait: Yes, that will only become relevant in the actual merge, which happens _after_ we shuffled all the data.

dask_expr/_merge.py
```diff
@@ -428,9 +428,20 @@ def _layer(self) -> dict:
         from distributed.shuffle._shuffle import shuffle_barrier
 
         dsk = {}
-        token = self._name.split("-")[-1]
-        token_left = token + "-left"
-        token_right = token + "-right"
+        token_left = _tokenize_deterministic(
+            "hash-join",
```
hendrikmakait, line 432: Added this to ensure that we avoid any unwanted sharing between merges and shuffles. I don't have a good-enough overview right now to guarantee that sharing between them works.

fjetter: If we can rely on `name_input_{left/right}` to truly be unique and identify distinct input dataframes, I believe we can even share between merges and ordinary shuffles. I wonder where this use case would come up, though

hendrikmakait: My main concern here is that `shuffle_transfer` and `merge_transfer` do different things. We'd have to refactor this so that they are in fact interchangeable and shareable. I guess there are only few cases where this would come in handy though. Maybe something like joining a dataframe on `x` and aggregating that same dataframe on `x` within the same graph?

phofl:
thx @hendrikmakait

PR #364, opened 2023-10-26
Implement multi-columns assign

phofl:


dask_expr/tests/test_collection.py
```diff
@@ -1261,3 +1261,48 @@ def test_replace_filtered_combine_similar():
...
+    #
+    # pdf = pdf[["x", "y", "z", "d"]]
+    # assert_eq(pdf, df)
+
+    pdf = lib.DataFrame({"a": [1, 2, 3, 4], "d": 4})
+    df = from_pandas(pdf, npartitions=2)
+
+    # This does not introduce an intermediate projection
+    df["x"] = df.a
+    df["y"] = df.a
```
mrocklin, line 1298: Would this also work if this was `df.a + 1`?

phofl: Yeah that would work, but I don't know what would happen with df["y"] = df["x"], I'll do more tests later

phofl:
please don't merge yet

phofl:
Yikes this is a PITA.

I think I am warming up to "simplify needs a global view of the expression to avoid eager pushdowns"

The problem here is as follows: ``_simplify_up`` might remove whole assign statements (which is good!). It becomes less good if we have 2 different branches with the same assignment statement and it's only removed in one of them. combine_similar has to inject this assign statement again so that we can squash branches together.

mrocklin:
Let's not worry about this for a while then

phofl:
Hm no this is actually not directly related to the multi column assign, we get the same behaviour bug without the multi column assign change. We just don't have a test on main that triggers it.

phofl:
Took care of this in another pr

PR #366, opened 2023-10-26
Make are_co_algined less reliant on divisions

phofl:
cc @rjzamora 

This *should* fix the problem with statistics collection

I think that we actually don't run into the broadcast able case very often because the more likely path is that we have the same root

### DIFF
--- a/dask_expr/_expr.py
+++ b/dask_expr/_expr.py
@@ -2257,14 +2257,18 @@ def non_blockwise_ancestors(expr):
 
 def are_co_aligned(*exprs):
     """Do inputs come from different parents, modulo blockwise?"""
-    exprs = [expr for expr in exprs if not is_broadcastable(expr)]
     ancestors = [set(non_blockwise_ancestors(e)) for e in exprs]
     unique_ancestors = {
         # Account for column projection within IO expressions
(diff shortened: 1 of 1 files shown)


dask_expr/_expr.py
```diff
@@ -2257,14 +2257,18 @@ def non_blockwise_ancestors(expr):
...
     ancestors = [set(non_blockwise_ancestors(e)) for e in exprs]
     unique_ancestors = {
         # Account for column projection within IO expressions
         _tokenize_partial(item, ["columns", "_series"])
         for item in flatten(ancestors, container=set)
     }
-    return len(unique_ancestors) <= 1
+    if len(unique_ancestors) <= 1:
+        return True
+    exprs_except_broadcast = [expr for expr in exprs if not is_broadcastable(expr)]
```
rjzamora, line 2268:
```suggestion
    # We tried avoiding an `npartitions` check above, but
    # now we need to consider "broadcastable" expressions.
    exprs_except_broadcast = [expr for expr in exprs if not is_broadcastable(expr)]
```

rjzamora (approved):
Thanks @phofl ! Just one (optional) comment suggestion.

Issue #365, opened 2023-10-26
Parquet statistics are collected twice


fjetter:
Currently, the parquet reader is using `split_row_groups='infer'` to infer whether it should split a file into multiple dask tasks or not which is done by collecting parquet statistics from every file.

Running the code below triggers this statistics collection twice. Once during the definition of the computation / while I'm mutating the dataframe and once as soon as I call compute.

```python
import dask_expr as dd

VAR1 = datetime(1998, 9, 2)

lineitem_ds = dd.read_parquet("s3://coiled-runtime-ci/tpc-h/scale-1000/lineitem")

lineitem_filtered = lineitem_ds[lineitem_ds.l_shipdate <= VAR1]
lineitem_filtered["sum_qty"] = lineitem_filtered.l_quantity
lineitem_filtered["sum_base_price"] = lineitem_filtered.l_extendedprice
lineitem_filtered["avg_qty"] = lineitem_filtered.l_quantity
lineitem_filtered["avg_price"] = lineitem_filtered.l_extendedprice

# This line now triggers a statistics collection iff `split_row_groups` is not False
lineitem_filtered["sum_disc_price"] = lineitem_filtered.l_extendedprice * (
    1 - lineitem_filtered.l_discount
)


lineitem_filtered["sum_charge"] = (
    lineitem_filtered.l_extendedprice
    * (1 - lineitem_filtered.l_discount)
    * (1 + lineitem_filtered.l_tax)
)

lineitem_filtered["avg_disc"] = lineitem_filtered.l_discount
lineitem_filtered["count_order"] = lineitem_filtered.l_discount

gb = lineitem_filtered.groupby(["l_returnflag", "l_linestatus"])

total = gb.agg(
    {
        "sum_qty": "sum",
        "sum_base_price": "sum",
        "sum_disc_price": "sum",
        "sum_charge": "sum",
        "avg_qty": "mean",
        "avg_price": "mean",
        "avg_disc": "mean",
        "count_order": "count",
    }
)

# Once I compute, another stats collection is triggered
total.compute()
```

![image](https://github.com/dask-contrib/dask-expr/assets/8629629/ae45d60e-5a32-4bbf-8bfc-fe29c9c0c71e)

related to https://github.com/dask-contrib/dask-expr/issues/363

fjetter:
My naive expectation would be that the second collection is not necessary since we're doing some caching. Is this a false cache miss or do we actually have to collect this twice for some reason?

[...]

PR #367, opened 2023-10-26
Separate `io_func` from `ReadParquet._plan` for better caching

rjzamora:
Right now, column projection will invalidate our `ReadParquet._plan` cache. There is no reason for this to happen, because the column projection does not actually affect the partitioning plan (it only affects the arguments to `ParquetFunctionWrapper`).

This PR separates the `ParquetFunctionWrapper` (used as the `io_func`) from `ReadParquet._plan`.

This change is partially related to https://github.com/dask-contrib/dask-expr/issues/365

### DIFF
--- a/dask_expr/io/parquet.py
+++ b/dask_expr/io/parquet.py
@@ -542,24 +542,43 @@ class ReadParquet(PartitionsFiltered, BlockwiseIO):
         meta = self.engine._create_dd_meta(dataset_info)
         index = dataset_info["index"]
         index = [index] if isinstance(index, str) else index
-        meta, index, columns = set_index_columns(
-            meta, index, self.operand("columns"), auto_index_allowed
+        meta, index, all_columns = set_index_columns(
+            meta, index, None, auto_index_allowed
         )
         if meta.index.name == NONE_LABEL:
             meta.index.name = None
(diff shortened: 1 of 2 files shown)


dask_expr/io/parquet.py
```diff
@@ -542,25 +542,44 @@ def _dataset_info(self):
...
 
     @property
     def _meta(self):
-        meta = self._dataset_info["meta"]
+        meta = self._dataset_info["base_meta"]
+        columns = _convert_to_list(self.operand("columns"))
         if self._series:
-            column = _convert_to_list(self.operand("columns"))[0]
-            return meta[column]
+            assert columns
```
phofl, line 561: I don't like this pattern very much. could we do ``assert len(columns) > 0``?

phofl (commented):
one small comment, otherwise lgtm

dask_expr/io/parquet.py
rjzamora, line 561:
```suggestion
            assert len(columns) > 0
```

rjzamora: Yeah, sounds good.

phofl (approved):

phofl:
Thx

PR #369, opened 2023-10-26
Implement multi assign

phofl:
cc @mrocklin 

We can merge this version, the bug I mentioned in the other PR is already there and we are not making it any worse with this change, this just adds a test that we didn't have before


phofl:
Took care of this in another pr

PR #377, opened 2023-10-30
Fix split_out=True in groupby reductions

phofl:
had some time before boarding

cc @mrocklin

### DIFF
--- a/dask_expr/_reductions.py
+++ b/dask_expr/_reductions.py
@@ -412,7 +412,7 @@ class ApplyConcatApply(Expr):
         chunked = self._chunk_cls(
             self.frame, type(self), chunk, chunk_kwargs, *self._chunk_cls_args
         )
-        if self.split_out == 1 or sort:
+        if not isinstance(self.split_out, bool) and self.split_out == 1 or sort:
             # Lower into TreeReduce(Chunk)
             return TreeReduce(
                 chunked,
--- a/dask_expr/tests/test_groupby.py
+++ b/dask_expr/tests/test_groupby.py
@@ -185,3 +185,10 @@ def test_groupby_agg_split_out(pdf, df, spec, sort, split_out):
 
     expect = pdf.groupby("x", sort=sort).agg(spec)
     assert_eq(agg, expect, sort_results=not sort)
+
+
+def test_groupby_reduction_shuffle(df, pdf):
+    q = df.groupby("x").sum(split_out=True)
+    assert q.optimize().npartitions == df.npartitions
+    expected = pdf.groupby("x").sum()
+    assert_eq(q, expected)


phofl:
There is an actual bug in the implementation as well for Series-like objects

mrocklin:
I was *just* writing up that as a comment 🙂 

```python
import dask_expr
df = dask_expr.datasets.timeseries()
df.groupby("name").x.sum(split_out=True).optimize(fuse=False).pprint()
```

mrocklin:
cc also @rjzamora

phofl:
drop_duplicates tests cover this case

mrocklin:
> drop_duplicates tests cover this case

For TPC-H at larger scales (1000+) this matters for more than just drop_duplicates.  There are some groupby-aggregations that are quite large.

phofl:
Yeah I agree, what I wanted to say is that the Series-bug that is fixed now has test coverage thanks to drop duplicates. 

Do you think it makes sense to choose the number of partitions automatically for these cases? Like previous number of partitions divided by something that depends on the length of the group keys or just go with split_out=True?

mrocklin:
> Do you think it makes sense to choose the number of partitions automatically for these cases? Like previous number of partitions divided by something that depends on the length of the group keys or just go with split_out=True?

I'll say some obvious thoughts to make sure we're all on the same track:

- we're forced to choose a number of partitions
- the question here is should that be 1, npartitions, or something in between
- it should probably always be something in between
- this depends on the values of the data, which we don't know ahead of time
- today our policy is "always choose 1"
- any policy we choose is likely to be often wrong

This seems like a case where we should probably build up a benchmark suite and choose something that is reliably less bad than the others.  There are lots of things that I can suggest, but I don't really know.  I'm open to whatever.  It sounds like maybe you have some ideas?

phofl:
Yeah I agree with all of your thoughts. There is only one semi-reliable way of guessing the number of output partitions that I am aware of which is the number of grouping columns. The assumption is the larger the number of group keys the larger the size of the result.

Obviously this isn't a measure that's good in general cases, but it might be wrong less often than the others. Figuring out the factor is another thing that will be hard. I don't have a very good idea either, just that npartitions shouldn't be 1 when we have more than one group key in general cases.

rjzamora:
@phofl is correct that the cardinality often grows proportionally to the number of grouping columns. Unfortunately, we do see pretty dramatic variation between datasets in single-column cardinality. 

My sense is that `split_out=1` is a reasonable default in many cases.  However, when we are grouping on one or more columns that have originated from a parquet dataset, we may be able to use the size of the dictionary encoding as a proxy for the cardinality.

rjzamora:
Thanks for looking into `split_out` fixes @phofl - #351 was definitely merged a bit pre-maturely (my fault for getting distracted by other things and not labeling it as draft).

I think it makes sense to get these PRs in. I may still need to take a follow-up pass over the code to "fix" a few other possible issues.

PR #378, opened 2023-10-30
Clean up backwards compat for Dask

phofl:


mrocklin:
Nice.  I'm glad that we're doing a good job of not keeping this stuff around.

PR #379, opened 2023-10-30
Fix bug in split_out and column projection in groupby

phofl:
found this in query 3


mrocklin:
@phofl I'd love to try this out, presumably with https://github.com/dask-contrib/dask-expr/pull/377 .  Unfortunately it looks like they have merge conflicts.  I'm not bothering right now to understand what's going on internally, but if you're confident and can merge that would be welcome.

PR #381, opened 2023-10-30
Automatically set split_out if it's not given

phofl:
This would handle the adjustment in the tune step

### DIFF
--- a/dask_expr/_groupby.py
+++ b/dask_expr/_groupby.py
@@ -31,6 +31,10 @@ def _as_dict(key, value):
     return {} if value is None else {key: value}
 
 
+def _adjust_split_out_for_group_keys(npartitions):
+    return 1 if npartitions < 100 else True
+
+
 ###
 ### Groupby-aggregation expressions
 ###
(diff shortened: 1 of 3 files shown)


dask_expr/_groupby.py
```diff
@@ -43,6 +43,10 @@ def _as_dict(key, value):
     return {} if value is None else {key: value}
 
 
+def _adjust_split_out_for_group_keys(npartitions):
+    return 1 if npartitions < 100 else True
```
rjzamora, line 47: This seems like a pretty dramatic change at a somewhat arbitrary partition count. We are suddenly switching from a  single-partition result to a 100-partition result.

Should `split_out` be proportional to the partition count instead? For example:

```python
def _adjust_split_out_for_group_keys(npartitions: int):
    return math.ceil(npartitions / 100)
```

phofl: Yeah that seems fine too

phofl: I've included the grouping columns as well as a parameter to reduce less if we have more group columns

PR #382, opened 2023-10-31
Inject repartition for compute

phofl:
cc @mrocklin 

minimal solution for what we discussed earlier. This also highlighted some bugs if you introduce the repartition manually.

### DIFF
--- a/dask_expr/_collection.py
+++ b/dask_expr/_collection.py
@@ -151,8 +151,14 @@ class FrameBase(DaskMethodsMixin):
         out = self.optimize(combine_similar=combine_similar, fuse=fuse)
         return DaskMethodsMixin.persist(out, **kwargs)
 
-    def compute(self, fuse=True, combine_similar=True, **kwargs):
-        out = self.optimize(combine_similar=combine_similar, fuse=fuse)
+    def compute(
+        self, fuse=True, combine_similar=True, inject_repartition=True, **kwargs
+    ):
+        if inject_repartition:
+            out = self.repartition(npartitions=1)
+        else:
+            out = self
+        out = out.optimize(combine_similar=combine_similar, fuse=fuse)
         return DaskMethodsMixin.compute(out, **kwargs)
 
     def __dask_graph__(self):
@@ -229,13 +235,13 @@ class FrameBase(DaskMethodsMixin):
     def head(self, n=5, compute=True):
         out = new_collection(expr.Head(self.expr, n=n))
         if compute:
-            out = out.compute()
+            out = out.compute(inject_repartition=False)
         return out
 
     def tail(self, n=5, compute=True):
         out = new_collection(expr.Tail(self.expr, n=n))
         if compute:
-            out = out.compute()
+            out = out.compute(inject_repartition=False)
         return out
 
     def copy(self):
--- a/dask_expr/_repartition.py
+++ b/dask_expr/_repartition.py
@@ -40,6 +40,14 @@ class Repartition(Expr):
     def _meta(self):
         return self.frame._meta
 
+    @property
+    def npartitions(self):
+        if "new_partitions" in self._parameters:
+            npartitions = self.operand("new_partitions")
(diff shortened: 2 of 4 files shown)


dask_expr/_collection.py
```diff
@@ -151,8 +151,14 @@ def persist(self, fuse=True, combine_similar=True, **kwargs):
         out = self.optimize(combine_similar=combine_similar, fuse=fuse)
         return DaskMethodsMixin.persist(out, **kwargs)
 
-    def compute(self, fuse=True, combine_similar=True, **kwargs):
-        out = self.optimize(combine_similar=combine_similar, fuse=fuse)
+    def compute(
+        self, fuse=True, combine_similar=True, inject_repartition=True, **kwargs
```
mrocklin, line 155: Hrm, this feels unclean to me.  I would expect to see this in the `optimize` function maybe, or somewhere where it will be called regardless (like `dask.compute(...)`).  I'm also nervous about the special-case. It looks like it was for head/tail?  I'm curious, what was the problem there?

phofl: It's just unnecessary for head and tail, but you are correct there is no reason not to add it. it will be a single partition anyway.

Yeah I tend to forget about dask.compute, sorry for this. We can't do this in optimise, we definitely don't want to repartition for persist.

mrocklin: I'll be happier if we don't have to add a keyword argument, even if we have an unnecessary call.  Maybe `Repartition(Head(...), 1)` can simplify down to just `Head(...)`?


I'll be mildly sad if we're inconsistent between `.compute()` and `dask.compute(...)` but that at least seems innocuous.

phofl: We can remove the argument.

We can special case in optimise if we want to add support for dask.compute

dask_expr/_repartition.py
```diff
@@ -131,6 +144,13 @@ class RepartitionToFewer(Repartition):
     def _divisions(self):
         return tuple(self.frame.divisions[i] for i in self._partitions_boundaries)
 
+    def _lower(self):
+        # Multiple lowering steps might have changed the npartitions
+        npartitions = self.new_partitions
+        npartitions_input = self.frame.npartitions
+        if npartitions_input <= npartitions:
+            return self.frame
```
mrocklin, line 152: This seems more like a simplify call to me?  Why here?  Why aren't simplify calls being used at this stage?

Also, slight style nit, but I would probably avoid the temporary variables here and just write this as

```python
if self.new_partitions <= self.frame.npartitions:
    return self.frame
```

Also, I'm surprised that this isn't `==`.  Does this mean that we're not supposed to reduce the number of partitions?

phofl: Yeah I don't like this pattern more than you do.

The check is the other way round,

```
if self.frame.npartitions <= self.new_partitions:
    return self.frame
```

Lower might reduce the number of partitions of self.frame, which might make something that has been a repartition to fewer call earlier into a no-op. We are lowering from the root, which makes us vulnerable to things like this (sort_values is an example that will reduce the number of partitions)

mrocklin: Sorry, I don't understand yet.  The situation I think you're describing is when we're calling `Repartition(X, 10)` where X has, say, 20 partitions. But then suddenly X has 5 partitions and so we're saying "well we no longer need to repartition X to 10 partitions, it's already at the nice low value of 5".  Is that right?

If so, I can see how this would make sense if our goal is to reduce partition counts, but this doesn't make sense to me in general.  What if someone wants precisely 10 partitions because they have ten machines or whatever and they want to have one partition per machine, no matter what.

phofl:
Kind of yeah. Assuming we want to keep the repartition call: Our lower
recursion logic is the problem then, We are recursing from the root, which
means we lock the repartitioning algorithm in before the operands are
lowered. This means we are screwed if the operands change their number of
partitions during lowering. Changing the order here would solve that problem

PR #384, opened 2023-11-02
Optimize repartition away after head

phofl:
pre-cursor for #382

### DIFF
--- a/dask_expr/tests/test_collection.py
+++ b/dask_expr/tests/test_collection.py
@@ -512,6 +512,16 @@ def test_head_head(df):
     assert a.optimize()._name == b.optimize()._name
 
 
+def test_head_tai_repartition(df):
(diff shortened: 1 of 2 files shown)


dask_expr/tests/test_collection.py
```diff
@@ -512,6 +512,16 @@ def test_head_head(df):
     assert a.optimize()._name == b.optimize()._name
 
 
+def test_head_tai_repartition(df):
```
rjzamora, line 515:
```suggestion
def test_head_tail_repartition(df):
```
Spelling nit.

rjzamora (approved):
LGTM besides the minor spelling suggestion.
I'd be interested to learn where you ran into this, but the change seems fine either way.

phofl:
#382  is where this comes up. Compute should inject a repartition layer that reduces to 1 so that we can optimise our way through other ops (sort values for example)

PR #385, opened 2023-11-06
Switch split_out in drop_duplicates to True

phofl:
True makes more sense for drop duplicates if we want to scale well


rjzamora (approved):
I agree that this is the way to go.

I'm mildly concerned about the different ordering you will get compared to pandas. However, this minor divergence seem much better than the risk of OOM errors we currently have with `split_out=1`.

phofl:
Yep agreed

PR #387, opened 2023-11-10
Add ``from_map`` API for custom collection creation

rjzamora:
The `dask.dataframe.from_map` API[ is a useful way to create a DataFrame collection](https://blog.dask.org/2023/04/12/from-map) using specialized logic or from unsupported file formats. It is also a reasonable way for users to implement a custom `read_parquet` API.

Key differences between the `dask.dataframe` and  `dask_expr` implementations:

- The `token` argument is not supported, because we need to have complete control over tokenization in dask-expr
- The `DataFrameIOFunction` protocol is not supported, because dask-expr makes it **much** easier to support column projection by more-convenient means. In the dask-expr implementation, we assume that the user wants column projection if `inspect.signature` reveals that `func` supports a `"columns"` argument. If the user has such an argument, but **doesn't** want column projection, they can just specify `allow_projection=False`.

### DIFF
--- a/dask_expr/_collection.py
+++ b/dask_expr/_collection.py
@@ -1265,7 +1265,7 @@ def from_map(
     divisions=None,
     label=None,
     token=None,
-    enforce_metadata=False,
+    enforce_metadata=True,
     allow_projection=True,
     **kwargs,
 ):
@@ -1273,8 +1273,7 @@ def from_map(
     from dask_expr.io import FromMap, FromMapProjectable
 
     if token is not None:
-        raise NotImplementedError()
-    if enforce_metadata:
+        # This option doens't really make sense in dask-expr
         raise NotImplementedError()
 
     if allow_projection:
@@ -1304,6 +1303,7 @@ def from_map(
                 args,
                 kwargs,
                 meta,
+                enforce_metadata,
                 divisions,
                 label,
             )
@@ -1316,6 +1316,7 @@ def from_map(
                 args,
                 kwargs,
                 meta,
+                enforce_metadata,
                 divisions,
                 label,
             )
--- a/dask_expr/_collection.py
+++ b/dask_expr/_collection.py
@@ -1254,3 +1254,36 @@ def concat(
             *[df.expr for df in dfs],
         )
     )
+
+
+def from_map(
+    func,
+    *iterables,
+    args=None,
+    meta=no_default,
+    divisions=None,
+    label=None,
+    token=None,
+    enforce_metadata=False,
+    **kwargs,
+):
+    from dask_expr.io import FromMap
+
+    if token is not None:
+        raise NotImplementedError()
+    if enforce_metadata:
+        raise NotImplementedError()
+
+    args = [] if args is None else args
+    kwargs = {} if kwargs is None else kwargs
+    return new_collection(
+        FromMap(
+            func,
+            iterables,
+            args,
+            kwargs,
+            meta,
+            divisions,
+            label,
+        )
+    )
--- a/dask_expr/_collection.py
+++ b/dask_expr/_collection.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import functools
+import inspect
 import warnings
 from collections.abc import Callable, Hashable, Mapping
 from numbers import Number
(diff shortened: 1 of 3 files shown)


dask_expr/_collection.py
```diff
@@ -1254,3 +1255,68 @@ def concat(
...
+    **kwargs,
+):
+    """Create a dask-expr collection from a custom function map"""
+    from dask_expr.io import FromMap, FromMapProjectable
+
+    if "token" in kwargs:
+        # This option doesn't really make sense in dask-expr
+        raise NotImplementedError("dask_expr does not support a token argument.")
+
+    if allow_projection:
```
phofl, line 1278: I might be missing something, but is it possible that we set allow_projection silently to False when the user sets True explicitly? That wouldn't be ideal

dask_expr/io/io.py
```diff
@@ -181,6 +183,168 @@ def _tune_up(self, parent):
...
+        "enforce_metadata": False,
+        "user_divisions": None,
+        "label": None,
+        "_partitions": None,
+    }
+    _absorb_projections = False
+
+    @functools.cached_property
+    def _name(self):
+        if self.label:
```
phofl, line 209: Can we check for None explicitly?

dask_expr/io/io.py
```diff
@@ -181,6 +183,168 @@ def _tune_up(self, parent):
...
+            )
+        return kwargs
+
+    @functools.cached_property
+    def frame_meta(self):
+        # This is our `_meta` result before possibly
+        # converting to a Series
+        meta = super()._meta
+        if meta.ndim > 1 and self.columns_operand is not None:
+            return meta[self.columns_operand]
```
phofl, line 328: This is confusing, shouldn't we use ``self.operand("columns")``? ``columns_operand`` is always a list, so we never coerce to a Series

phofl: Ah sorry, forget it, missed meta below

dask_expr/_collection.py
rjzamora: Yeah, good question. There is no documentation for this yet, so silently ignoring the `allow_projection=True` default certainly seems strange. The idea is that we **cannot** support column projection unless the `func` argument includes a `"columns"` argument. The purpose of `allow_projection` is to provide the user with an "opt out" mechanism for the case that their `func` argument does include a `"columns"` argument, but they do not want dask-expr to project column selections into their `FromMap` layer.

Maybe there is a better API for this? I get the sense that we want users to get column projection as automatically as possible. I suppose we can remote `allow_projection` as a `from_map` argument if the documentation is clear that `simplify` will always **use**  the `"columns"` argument if `func` has one in its signature?

phofl: >  I suppose we can remote allow_projection as a from_map argument if the documentation is clear that simplify will always use the "columns" argument if func has one in its signature?

I think I like starting with this and moving to more arguments later on if it becomes necessary

rjzamora: Yeah, I agree with this.

dask_expr/io/io.py
```diff
@@ -181,6 +183,168 @@ def _tune_up(self, parent):
...
+    def _series(self):
+        # Only need to convert to _series if func
+        # doesn't produce a Series already
+        return self.operand("_series") and self.frame_meta.ndim > 1
+
+    @functools.cached_property
+    def kwargs(self):
+        options = self.operand("kwargs")
+        if self.columns_operand:
+            options["columns"] = self.columns_operand
```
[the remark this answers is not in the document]
rjzamora: Ah, good catch

dask_expr/io/io.py
```diff
@@ -181,6 +183,168 @@ def _tune_up(self, parent):
...
+    def _series(self):
+        # Only need to convert to _series if func
+        # doesn't produce a Series already
+        return self.operand("_series") and self.frame_meta.ndim > 1
+
+    @functools.cached_property
+    def kwargs(self):
+        options = self.operand("kwargs")
+        if self.columns_operand:
+            options["columns"] = self.columns_operand.copy()
```
phofl, line 306: Sorry I meant both here, ``options`` is also an input as far as I understand? If you insert columns then this will propagate to the user

phofl: Similar to what you did in ``def apply_kwargs(self):`` above

dask_expr/io/tests/test_io.py
```diff
@@ -365,3 +373,70 @@ def test_combine_similar_no_projection_on_one_branch(tmpdir):
 
     pdf["xx"] = pdf.x != 0
     assert_eq(df, pdf)
+
+
+# @pytest.mark.parametrize("enforce_metadata", [True, False])
```
phofl, line 378: This is for later?

phofl (approved):

phofl:
thx @rjzamora

PR #388, opened 2023-11-13
Fuse sequential filters

mrocklin:
I just noticed this in some user code and wanted to play around with dask-expr.  This isn't important.

This may also be wrong?  I haven't thought deeply yet about combining predicates from different dataframes and how that might interact.  I *think* everything here should be ok, but I haven't thought through it deeply.

### DIFF
--- a/dask_expr/_expr.py
+++ b/dask_expr/_expr.py
@@ -1689,6 +1689,11 @@ class Filter(Blockwise):
         if isinstance(parent, Index):
             return self.frame.index[self.predicate]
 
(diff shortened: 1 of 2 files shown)


phofl (commented):
There are cases where this makes performance worse, e.g. when the first filter restricts the number of rows heavily for example.

It will make cases faster where the first filter doesn't remove many rows

phofl:
This works now (I chose a path that's a little different because there were nasty corner cases)

PR #389, opened 2023-11-14
Implement groupby transform

phofl:
There are some differences to the dask DataFrame API, but the current implementation there makes me sad, so this is a good start


phofl:
Merging this, want to build shift on top of it

PR #395, opened 2023-11-16
Rewrite optimisations to avoid over-eager pushdown

phofl:
First step towards a new optimization organisation


mrocklin:
This may be the right approach.  I'll raise the concern we mentioned when we spoke last, just to be explicit.

```python
df = dd.read_parquet(...)
train = df[["x", "y", "z"]]

df.y.sum().compute()
```

When we're computing `df.y.sum()` we may ask `df` what columns it dependents need and it'll say "all of x, y, and z" but for this computation we really only need `y`.   It could be that dependent tracking is really only helpful within the context of a single computation.

phofl:
Yeah you are correct, this is also causing troubles in other places too. I removed that part of the logic and replaced it with an explicit trigger that collects dependents.

This is just a POC. We are doing one simplify step at a time before we have to refresh our dependents, and then repeat. This is an exemplary implementation for Blockwise ops (I am pretty sure that we can make this more efficient than it currently is, but it's a good start)


```
df = pd.DataFrame({"a": [1, 2, 3, 1, 2, 2], "b": [1, 2, 3, 4, 5, 6], "c": 1, "d": 1})

df = from_pandas(df, npartitions=2)

df = df.replace(1, 5)
df["x"] = df[["b", "c"]].b * 2
df["y"] = df.replace(2, 6).fillna(10).c * 3
df = df[["a", "x", "y"]].simplify()
```

This results in:

```
Assign: key='y'
  Assign: key='x'
    Projection: columns=['a']
      Replace: to_replace=1 value=5
        FromPandas: frame='<pandas>' npartitions=2 columns=['a', 'b', 'c']
    Mul: right=2
      Projection: columns='b'
        Replace: to_replace=1 value=5
          FromPandas: frame='<pandas>' npartitions=2 columns=['a', 'b', 'c']
  Mul: right=3
    Projection: columns='c'
      Fillna: value=10
        Replace: to_replace=2 value=6
          Projection: columns=['c']
            Replace: to_replace=1 value=5
              FromPandas: frame='<pandas>' npartitions=2 columns=['a', 'b', 'c']
```

So we can avoid over-eager pushdown of projections.

dask_expr/_expr.py
```diff
@@ -309,20 +339,84 @@ def rewrite(self, kind: str):
...
             output expression
         changed:
             whether or not any change occured
         """
-        return self.rewrite(kind="simplify")
+        expr = self
+
+        while True:
+            if expr._name in cache:
+                return cache[expr._name]
```
mrocklin, line 365: This could be done in `__new__` if we want to do it globally.

dask_expr/_expr.py
```diff
@@ -60,6 +61,35 @@ def __init__(self, *args, **kwargs):
...
+        seen = set()
+        while stack:
+            node = stack.pop()
+            if node._name in seen:
+                continue
+            seen.add(node._name)
+
+            for dep in node.dependencies():
+                stack.append(dep)
+                dep._add_dependents(node)
```
mrocklin, line 83: Maybe a dumb idea, but we could also put this in `__init__` but only use it if we're in an optimization phase or something.  Maybe we turn it on when we enter optimize, and then recreate the full expression tree as a first stage.  Then everything will be kept up to date automatically during optimization and there's no need to explicitly refresh it.  

But again, maybe a dumb idea.

phofl: Yeah I thought about this too for a little bit, I am not sure if we can ensure that a new node always gets all dependents of the old node right away without moving the logic into simplify (which is something I am trYing to avoid). I was running into too over-eager pushdowns when I tried this

dask_expr/_expr.py
```diff
@@ -311,25 +317,79 @@ def rewrite(self, kind: str):
 
         return expr
 
-    def simplify(self):
+    def simplify_once(self, dependents: defaultdict):
```
fjetter, line 320: I'm a little worried about runtime complexity here. I know that expr graphs are supposed to be small but what I see is definitely a non-linear runtime in the number of edges `E` (at the least)

Generally, this function is called more than once per node. In fact, due to line 365, this is called `E` times. 
`_simplify_up` in L350 then accepts the dependents and may call things like `determine_column_projection` which iterates over the dependents which effectively brings us to a quadratic runtime in `E`

fjetter: This is not necessarily a showstopper but it is concerning, paritcularly if `simplify_up` became more complex down the road.

fjetter: `substitute_parameters` which is also called occasionally in simplify_up is also iterating over the edges so that's a second source besides `determine_column_projection` for this runtime complexity.

fjetter (commented):
Re: runtime complexity. I haven't reviewed the old code so I don't know if this was a regression or not. I assume that most expression graphs have very few edges (and are small-ish overall) so I guess this is not a huge deal but it is a little worrisome since whatever is being written in `simplify_up` has to be fast. Passing a collection of the size of the entire graph to that function feels... tempting to do the wrong thing.
The largest HLG graphs that were reported on the dask/dask tracker were a couple of (hundred) thousand layers (which I believe was a user error and in the array landscape). Still, I could see this grow for complex queries.

Other than this, I like the direction considering that the code complexity seems to go down a lot

dask_expr/_expr.py
```diff
@@ -2923,6 +2838,62 @@ def _execute_task(graph, name, *deps):
         return dask.core.get(graph, name)
 
 
+def collect_depdendents(expr) -> defaultdict:
```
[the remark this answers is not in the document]
phofl: This is currently not working as it should if you do dask.compute(a,b,c), but neither does the status quo.

The tuple expression is what will address this problem

> This will only return the correct result if expr is a leaf, doesn't it?

depends on what you are looking for, it would be correct if you are only interested in a subtree, otherwise not

dask_expr/_expr.py
```diff
@@ -2923,6 +2838,62 @@ def _execute_task(graph, name, *deps):
...
+    seen = set()
+    while stack:
+        node = stack.pop()
+        if node._name in seen:
+            continue
+        seen.add(node._name)
+
+        for dep in node.dependencies():
+            stack.append(dep)
+            dependents[dep._name].append(weakref.ref(node))
```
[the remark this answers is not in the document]
phofl: This is a little bit of a shortcut, in case we can prune whole branches so that we don't have to consider them when doing optimisations

You could also interpret this as premature optimization, but I was a bit worried that I would keep things alive that will screw me over if I have hard refs

dask_expr/_expr.py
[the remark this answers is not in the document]
fjetter: > depends on what you are looking for, it would be correct if you are only interested in a subtree, otherwise not

Well, we're not always dealing with trees, are we? Consider

```mermaid
graph BT;
    A1-->B1;
    A2-->B1;
    A3-->B2;
    A4-->B2;
    B1-->C1;
    B2-->C1;
    B2-->C2;
    C1-->D;
    C2 -->D

```

Unless you start with `D` you will not get the dependents of `B2` right since you're not walking the entire graph.

I _assume_ this is fine since we are always starting at a terminal node?

dask_expr/_expr.py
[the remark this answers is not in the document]
fjetter: I'm not blocking on this but it feels a little brittle to rely on weakref semantics to get this right. I'd be worried if this logic would not work out without weakrefs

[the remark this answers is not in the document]
phofl: No the logic works without, it would just need more simplify steps to get to the final result

dask_expr/_expr.py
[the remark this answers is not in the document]
phofl: Sorry I wasn't clear enough.

We are always starting with the node that is the leaf node for the thing that we want to compute, so yes

phofl:
Rebased and I intend to merge later today

dask_expr/_expr.py
phofl: @fjetter and I chatted offline a little about this

It's a potential issue for us that I intend to investigate in January. We can avoid some of this by caching the projected columns on every simplify step, which would hopefully solve the runtime complexity issue. Current main exhibits the same pattern though, so this does not block this pr

rjzamora: @fjetter - Do you think this complexity may have contributed to https://github.com/dask-contrib/dask-expr/issues/796? Sorry - I never got a chance to look through these changes, so I'm not really sure how the new system works yet.

fjetter: > Do you think this complexity may have contributed to

Possibly. I'll try to have a look at this shortly

PR #399, opened 2023-11-20
Add remaining _resample.Resampler API

milesgranger:


milesgranger:
Was a bit confused with the `test_resample_agg` in https://github.com/dask-contrib/dask-expr/pull/399/commits/db591f7f658e7b65e9f366c02f5c0f08ae9c456a. At top level, they are exactly equal but in `assert_eq` seems the indexes are different. However found calling compute then comparing made it happy. Something is off there though.

milesgranger:
Also, all the other methods didn't support arguments in dask, so just ignored them. For example `quantile`:

```
Parameters
----------
q : float or array-like, default 0.5 (50% quantile)  (Not supported in Dask)
```
so likewise, I didn't add any parameter to that method.. which is probably wrong(?)

dask_expr/tests/test_resample.py
```diff
@@ -20,15 +20,56 @@ def df(pdf):
...
+        q = result.simplify()
+        eq = getattr(df["foo"].resample("2T"), api)().simplify()
+        assert q._name == eq._name
+
+
+def test_resample_agg(df, pdf):
+    def my_sum(vals, foo=None, *, bar=None):
+        return vals.sum()
+
+    # TODO: w/o .compute() assert_eq fails w/ dtype assertion error
```
phofl, line 64: Can you post a traceback?

milesgranger:
```
ddf = <dask_expr.expr.Index: expr=Index(frame=ResampleAgg(frame=df, rule=2T, kwargs={}, fill_value=nan, how_args=(<function test_resample_agg.<locals>.my_sum at 0x7f11fbdf4f70>, 'foo'), how_kwargs={'bar': 'bar'}))>
res = DatetimeIndex(['2000-01-01 00:00:00', '2000-01-01 00:02:00',
               '2000-01-01 00:04:00', '2000-01-01 00:06:00',
               '2000-01-01 00:08:00', '2000-01-01 00:10:00'],
              dtype='datetime64[ns]', freq='2T')
numeric_equal = True

    def assert_dask_dtypes(ddf, res, numeric_equal=True):
        """Check that the dask metadata matches the result.

        If `numeric_equal`, integer and floating dtypes compare equal. This is
        useful due to the implicit conversion of integer to floating upon
        encountering missingness, which is hard to infer statically."""

        eq_type_sets = [{"O", "S", "U", "a"}]  # treat object and strings alike
        if numeric_equal:
            eq_type_sets.append({"i", "f", "u"})

        def eq_dtypes(a, b):
            return any(
                a.kind in eq_types and b.kind in eq_types for eq_types in eq_type_sets
            ) or (a == b)

        if not is_dask_collection(res) and is_dataframe_like(res):
            for a, b in pd.concat([ddf._meta.dtypes, res.dtypes], axis=1).itertuples(
                index=False
            ):
                assert eq_dtypes(a, b)
        elif not is_dask_collection(res) and (is_index_like(res) or is_series_like(res)):
            a = ddf._meta.dtype
            b = res.dtype
>           assert eq_dtypes(a, b)
E           AssertionError

../../mambaforge/envs/dask-expr/lib/python3.10/site-packages/dask/dataframe/utils.py:699: AssertionError
>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>> entering PDB >>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>

>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>> PDB post_mortem >>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>
> /home/milesg/mambaforge/envs/dask-expr/lib/python3.10/site-packages/dask/dataframe/utils.py(699)assert_dask_dtypes()
-> assert eq_dtypes(a, b)
(Pdb) p a
dtype('O')
(Pdb) p b
dtype('<M8[ns]')
```

dask_expr/tests/test_resample.py
```diff
@@ -20,15 +20,56 @@ def df(pdf):
...
+    expected = pdf.resample("2T").agg(my_sum, "foo", bar="bar")
+    assert_eq(result, expected)
+
+    result = df.resample("2T").agg(my_sum)["foo"].compute()
+    expected = pdf.resample("2T").agg(my_sum)["foo"]
     assert_eq(result, expected)
 
-    q = result.simplify()
-    eq = df["foo"].resample("2T").count().simplify()
+    q = df.resample("2T").agg(my_sum)["foo"].simplify()
```
phofl, line 73: Can you block optimizations for agg in the beginning? I am not sure how agg works exactly, it could be that this is not allowed since the callable in agg can access bar and this optimising would raise

dask_expr/_resample.py
```diff
@@ -125,14 +193,66 @@ def __init__(self, obj, rule, **kwargs):
     def _single_agg(
         self,
         expr_cls,
+        how_args=ResampleReduction._defaults["how_args"],
```
phofl, line 196: I'd rather not access the _defaults variable here, you can either be explicit or do a **kwargs (I'd prefer being explicit)

dask_expr/tests/test_resample.py
phofl: Could you try:

```
getattr(meta_nonempty(self.frame._meta).resample(self.rule, **self.kwargs), self.how)()
```

as _meta for both classes? Maybe that keeps the dtype properly

milesgranger: That's the ticket! Thanks! https://github.com/dask-contrib/dask-expr/pull/399/commits/42d4355dbc5f609f4cb73da93d90c2fc2610785e

dask_expr/tests/test_resample.py
milesgranger: Something like https://github.com/dask-contrib/dask-expr/pull/399/commits/403447aaecf1de033bf52c9612f19bc8f1f5f1f8?

dask_expr/_resample.py
milesgranger: Let it be explicit then. :+1:  https://github.com/dask-contrib/dask-expr/pull/399/commits/0f6fba3be04d459088e55c837dc927a3215a8cdc

dask_expr/tests/test_resample.py
phofl: I'd prefer overriding ``_simplify_up`` in the Agg class

dask_expr/_resample.py
```diff
@@ -91,7 +95,9 @@ class ResampleAggregation(Blockwise):
 
     @functools.cached_property
     def _meta(self):
-        return getattr(self.frame._meta.resample(self.rule, **self.kwargs), self.how)()
+        return getattr(
```
phofl, line 98: Can you wrap both statements into ``make_meta``? We don't want a meta object with values in it by default

milesgranger: https://github.com/dask-contrib/dask-expr/pull/399/commits/29ddb3a44ff14376566920e20d5d2d37b9d1bdb6

dask_expr/tests/test_resample.py
milesgranger: Ya that makes sense. https://github.com/dask-contrib/dask-expr/pull/399/commits/63b327301047df3e6ee2583a892ee9b424ee6000

phofl (approved):

phofl:
thx!

PR #400, opened 2023-11-20
Fix from_pandas for sort

phofl:
This is a little bit embarrassing...


rjzamora (approved):
Oof. Thanks for the fix!

PR #403, opened 2023-11-20
Fix bug when isin is the filter predicate

phofl:


dask_expr/_expr.py
```diff
@@ -1440,10 +1440,22 @@ class Replace(Elemwise):
 
 
 class Isin(Elemwise):
+    _filter_passthrough = False
     _projection_passthrough = True
     _parameters = ["frame", "values"]
     operation = M.isin
 
+    def _simplify_up(self, parent):
+        if self._filter_passthrough and isinstance(parent, Filter):
```
mrocklin, line 1449: Isn't this always `False`?

PR #405, opened 2023-11-21
Initial implementation of _rolling.Rolling

milesgranger:
Only valid for single partition DataFrames on this iteration

### DIFF
--- a/dask_expr/_rolling.py
+++ b/dask_expr/_rolling.py
@@ -86,7 +86,7 @@ class RollingAggregation(Blockwise):
 
     @functools.cached_property
     def _meta(self):
-        return self.frame.meta
+        return self.frame._meta
 
 
 class RollingCount(RollingReduction):
--- a/dask_expr/_rolling.py
+++ b/dask_expr/_rolling.py
@@ -53,6 +53,8 @@ class RollingReduction(Expr):
         )
 
     def _lower(self):
+        if not self._has_single_partition:
+            raise NotImplementedError("Not implemented for more than 1 partitions")
         return RollingAggregation(
             self.frame,
             self._has_single_partition,
@@ -79,11 +81,8 @@ class RollingAggregation(Blockwise):
     def operation(
         frame, _has_single_partition, window, kwargs, how, how_args, how_kwargs
     ):
-        if _has_single_partition:
-            rolling = frame.rolling(window, **kwargs)
-            return getattr(rolling, how)(*how_args, **(how_kwargs or {}))
-        else:
-            raise NotImplementedError("Not implemented for more than 1 partitions")
+        rolling = frame.rolling(window, **kwargs)
+        return getattr(rolling, how)(*how_args, **(how_kwargs or {}))
 
     @functools.cached_property
     def _meta(self):
--- a/dask_expr/_rolling.py
+++ b/dask_expr/_rolling.py
@@ -86,9 +86,7 @@ class RollingAggregation(Blockwise):
 
     @functools.cached_property
     def _meta(self):
-        rolling = self.frame._meta.rolling(self.window, **self.kwargs)
-        meta = getattr(rolling, self.how)(*self.how_args, **self.how_kwargs or {})
-        return make_meta(meta)
+        return self.frame.meta
 
 
 class RollingCount(RollingReduction):
--- a/dask_expr/_rolling.py
+++ b/dask_expr/_rolling.py
@@ -93,11 +93,6 @@ class RollingAggregation(Blockwise):
         meta = getattr(rolling, self.how)(*self.how_args, **self.how_kwargs or {})
         return make_meta(meta)
 
-    def _blockwise_arg(self, arg, i):
-        if isinstance(arg, BlockwiseDep):
-            return arg.iterable[i]
-        return super()._blockwise_arg(arg, i)
-
 
 class RollingCount(RollingReduction):
     how = "count"
--- a/dask_expr/_rolling.py
+++ b/dask_expr/_rolling.py
@@ -1,0 +1,218 @@
+import functools
+from collections import namedtuple
+from numbers import Integral
+
+from dask.dataframe.dispatch import meta_nonempty
+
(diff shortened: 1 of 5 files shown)


dask_expr/_rolling.py
```diff
@@ -0,0 +1,218 @@
...
+    @functools.cached_property
+    def kwargs(self):
+        return {} if self.operand("kwargs") is None else self.operand("kwargs")
+
+    def _simplify_up(self, parent):
+        if isinstance(parent, Projection):
+            return type(self)(self.frame[parent.operand("columns")], *self.operands[1:])
+
+    @property
+    def _has_single_partition(self):
```
phofl, line 50: I don't like this name (not a blocker, can do in a follow up). It will let multi-partition frames through when window=1 or we roll over the columns

dask_expr/_rolling.py
```diff
@@ -0,0 +1,218 @@
...
+    ):
+        if _has_single_partition:
+            rolling = frame.rolling(window, **kwargs)
+            return getattr(rolling, how)(*how_args, **(how_kwargs or {}))
+        else:
+            raise NotImplementedError("Not implemented for more than 1 partitions")
+
+    @functools.cached_property
+    def _meta(self):
+        rolling = meta_nonempty(self.frame._meta).rolling(self.window, **self.kwargs)
```
phofl, line 92: I think we can do self.frame._meta here (also not a blocker)

dask_expr/_rolling.py
```diff
@@ -0,0 +1,218 @@
...
+        meta = getattr(rolling, self.how)(*self.how_args, **self.how_kwargs or {})
+        return make_meta(meta)
+
+    def _blockwise_arg(self, arg, i):
```
phofl, line 96: We don't need this method here

dask_expr/_rolling.py
```diff
@@ -0,0 +1,218 @@
...
+
+    @staticmethod
+    def operation(
+        frame, _has_single_partition, window, kwargs, how, how_args, how_kwargs
+    ):
+        if _has_single_partition:
+            rolling = frame.rolling(window, **kwargs)
+            return getattr(rolling, how)(*how_args, **(how_kwargs or {}))
+        else:
+            raise NotImplementedError("Not implemented for more than 1 partitions")
```
phofl, line 88: We should put the npartitions> 1 case into a different class, I'd rather not mix those up, the check needs to go into the lower method above

dask_expr/tests/test_rolling.py
```diff
@@ -0,0 +1,83 @@
...
+        ("max", ()),
+        ("var", ()),
+        ("std", ()),
+        ("median", ()),
+        ("skew", ()),
+        ("quantile", (0.5,)),
+        ("kurt", ()),
+    ],
+)
+@pytest.mark.parametrize("window", (2,))
```
phofl, line 42: can you add 1?

dask_expr/tests/test_rolling.py
```diff
@@ -0,0 +1,83 @@
...
+    result = getattr(df.rolling(*args, **kwargs), api)(*how_args)["foo"]
+    expected = getattr(pdf.rolling(*args, **kwargs), api)(*how_args)["foo"]
+    assert_eq(result, expected)
+
+    q = result.simplify()
+    eq = getattr(df["foo"].rolling(*args, **kwargs), api)(*how_args).simplify()
+    assert q._name == eq._name
+
+
+@pytest.mark.parametrize("window", (2,))
```
phofl, line 63: Same here

dask_expr/_rolling.py
milesgranger: Stole this from Dask, wasn't sure how far to deviate from logic/naming. Do you have a preference on the preferred name?

dask_expr/_rolling.py
milesgranger: https://github.com/dask-contrib/dask-expr/pull/405/commits/5814462ed075c03554fd7792e433861394b4ce76

dask_expr/_rolling.py
milesgranger: Gone! https://github.com/dask-contrib/dask-expr/pull/405/commits/a557e674456c2848022fd1acee483f3b0eb77e55

dask_expr/_rolling.py
milesgranger: I moved it into RollingReduction._lower as suggested. But wasn't sure about the moving into different class. Just inherit from RollingReduction and call it RollingReductionMultiPartition(?). 
https://github.com/dask-contrib/dask-expr/pull/405/commits/5b68fbbbb4c95e4da68fd0f93a8b77aef8c8525e

milesgranger: Or sorry, confused, meant dispatch to `RollingAggregationMultiPartition`?

dask_expr/tests/test_rolling.py
milesgranger: Done https://github.com/dask-contrib/dask-expr/pull/405/commits/608f7e8549d07e3933d81bcb2939ace69733495f

dask_expr/tests/test_rolling.py
milesgranger: Done https://github.com/dask-contrib/dask-expr/pull/405/commits/608f7e8549d07e3933d81bcb2939ace69733495f

dask_expr/_rolling.py
phofl: ``_is_blockwise_op`` is better suited

dask_expr/_rolling.py
phofl: Yeah kind of, but we will most likely only dispatch to MapOverlap instead (see my other pr).

Anyway, this decision should be made in lower generally speaking

dask_expr/_rolling.py
phofl: Fine to rename in a follow up

phofl:
Thx!

PR #406, opened 2023-11-21
Implement map_overlap

phofl:
Planning on merging this soonish to unblock rolling work, but feedback welcome

### DIFF
--- a/dask_expr/_expr.py
+++ b/dask_expr/_expr.py
@@ -1332,7 +1332,7 @@ class MapOverlap(MapPartitions):
         return after
 
     def _lower(self):
-        overlapped = MapOverlapInterleavePartitions(self.frame, self.before, self.after)
+        overlapped = CreateOverlappingPartitions(self.frame, self.before, self.after)
 
         return MapPartitions(
             overlapped,
@@ -1345,7 +1345,7 @@ class MapOverlap(MapPartitions):
         )
 
 
-class MapOverlapInterleavePartitions(Expr):
+class CreateOverlappingPartitions(Expr):
     _parameters = ["frame", "before", "after"]
 
     @functools.cached_property
--- a/dask_expr/_expr.py
+++ b/dask_expr/_expr.py
@@ -24,6 +24,7 @@ from dask.dataframe.core import (
     make_meta,
 )
 from dask.dataframe.dispatch import meta_nonempty
+from dask.dataframe.rolling import CombinedOutput, _head_timedelta, overlap_chunk
 from dask.dataframe.utils import clear_known_categories, drop_by_shallow_copy
 from dask.typing import no_default
 from dask.utils import M, apply, funcname, import_required, is_arraylike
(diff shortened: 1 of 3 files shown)


dask_expr/_expr.py
```diff
@@ -1275,6 +1276,196 @@ def _task(self, index: int):
             )
 
 
+class MapOverlap(MapPartitions):
```
rjzamora, line 1279: Should we keep this in a new `dask_expr/_rolling.py`?

dask_expr/tests/test_collection.py
```diff
@@ -1291,3 +1291,59 @@ def test_replace_filtered_combine_similar():
...
+    result = df.map_overlap(func, before=0, after="2D")
+    expected = lib.DataFrame([5, 5, 5, 3, 3], index=idx, columns=["a"])
+    assert_eq(result, expected)
+    result = df.map_overlap(func, before=0, after=1)
+    assert_eq(result, expected)
+
+    # Bug in dask/dask
+    # result = df.map_overlap(func, before=0, after="1D")
+    # expected = lib.DataFrame([4, 4, 4, 3, 3], index=idx, columns=["a"])
+    # assert_eq(result, expected)
```
rjzamora, line 1314: Is there an issue open?

rjzamora (commented):
Thanks for putting this together.  This PR is relatively large, and you are signaling that it is also time sensitive. Are there specific areas/concerns that you are unsure of or are interested in feedback on?

dask_expr/tests/test_collection.py
phofl: https://github.com/dask/dask/pull/10639

phofl:
The complexity is in class MapOverlapInterleavePartitions(Expr):, everything else is fairly straightforward, so a set of eyeballs would be appreciated there

dask_expr/_expr.py
phofl: Hm I'd rather not, this is not rolling specific (fillna uses this as well for example)

rjzamora: Okay cool. I noticed that the `dask.dataframe` version was in a `rolling` module, so I just wanted to check.

phofl: Yeah that part confused me :)

rjzamora, line 1279: Mostly a "note to self": I'm wondering if there is any "danger" in treating the pre-lowered version of `MapOverlap` as a `Blockwise` expression. It is very important that we don't do any partition-related optimizations (e.g. culling) until after this expression is lowered. It seems like we are in the clear for the current `optimize` implementation, but it isn't clear to me that we are gaining much by inheriting from `MapPartitions`  to begin with?

dask_expr/_expr.py
```diff
@@ -1275,6 +1276,196 @@ def _task(self, index: int):
...
+            _overlap_chunk,
+            self._meta,
+            self.enforce_metadata,
+            self.transform_divisions,
+            self.clear_divisions,
+            self._kwargs,
+        )
+
+
+class MapOverlapInterleavePartitions(Expr):
```
rjzamora, line 1348: Another note to self: The `Interleave` term comes from the fact that we are extracting before/after rows for each partition and effectively "interleaving" the duplicated data between the "real" partitions. I don't love this name, but I also don't really have a better suggestion (maybe `CreateOverlappingPartitions`?)

rjzamora (commented):
Still working through this whenever I have free moments. Not seeing any serious problems yet.

dask_expr/_expr.py
phofl: Culling in general has to happen after lower, but I agree that this isn't ideal.

Maybe we have to create a new class that sits somewhere in between Expr and Blockwise, but removing this information removes a lot of information that is helpful

dask_expr/_expr.py
phofl: No strong opinion either way

phofl:
I am merging this to unblock rolling work, but happy to iterate further

Issue #402, opened 2023-11-20
Running into RecursionError: maximum recursion depth exceeded 


aimran-adroll:
Running a fairly simple workload to test out the expr. But quickly ran into a RecursionError issue

I can not share data but the code is following. I have ran it on coiled as well as on my laptop

```python
pdf_raw = dd.read_parquet(
    files,  # list of 100 parquet files on s3
    columns=["cid, "pid", "pg", "pa"],
)   


pdf_raw = pdf_raw.rename(columns={"pid": "product_id"})
pdf_raw = pdf_raw.fillna(value={"pg":"", "pa":""})
pdf_raw = pdf_raw[pdf_raw.pa.isin(['','YWRkdG9jYXJ0','cHVyY2hhc2Vk'])]

pdf_raw = pdf_raw.set_index("cid")

pdf_raw.head()
```

It begins by taking forever to "think". At least with regular dask, you can see task stream marching with each partitions to know its loading data etc.

In any case, almost within 5 seconds, it fails here

```
File ~/Code/python/revealwise/.venv/lib/python3.10/site-packages/pandas/core/dtypes/generic.py:38, in create_pandas_abc_type.<locals>._check(inst)
     37 def _check(inst) -> bool:
---> 38     return getattr(inst, attr, "_typ") in comp

RecursionError: maximum recursion depth exceeded while calling a Python object
```

My System details:  I have ran it on coiled as well as on my laptop

```
Python version       : 3.10.13
IPython version      : 8.17.2

Compiler    : Clang 16.0.3 
OS          : Darwin
Release     : 22.6.0
Machine     : arm64
Processor   : arm
CPU cores   : 8
Architecture: 64bit


s3fs     : 2023.10.0
pandas   : 2.1.3
numpy    : 1.26.2
coiled   : 1.1.14
dask     : 2023.11.0
dask_expr: 0.2.0
```

phofl:
Thx for the report. We are a little susceptible to recursions still because a few things aren't supported yet, I will take a look

phofl:
I was able to fix the issue (is already on main)

I pushed out a release to pypi, conda-forge will follow after their bots pick up the release 

A note regarding it takes a little to load: We do a bunch of optimisations before we start running actual code, your reproducer failed during the optimization step, that's why you can't see anything on the dask dashboard

aimran-adroll:
Thanks Patrick for the ⚡ response!

Confirming that RecursionError is gone. 

[...]

PR #407, opened 2023-11-21
Add upper bound for compression factor in IO fusion

phofl:
addresses https://github.com/dask-contrib/dask-expr/issues/402

### DIFF
--- a/dask_expr/io/io.py
+++ b/dask_expr/io/io.py
@@ -173,9 +173,12 @@ class FusedIO(BlockwiseIO):
 
     @functools.cached_property
     def _fusion_buckets(self):
-        step = math.ceil(1 / self.operand("expr")._fusion_compression_factor)
         partitions = self.operand("expr")._partitions
         npartitions = len(partitions)
+
+        step = math.ceil(1 / self.operand("expr")._fusion_compression_factor)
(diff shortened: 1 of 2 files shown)


dask_expr/io/io.py
```diff
@@ -173,9 +173,12 @@ def _task(self, index: int):
 
     @functools.cached_property
     def _fusion_buckets(self):
-        step = math.ceil(1 / self.operand("expr")._fusion_compression_factor)
         partitions = self.operand("expr")._partitions
         npartitions = len(partitions)
+
+        step = math.ceil(1 / self.operand("expr")._fusion_compression_factor)
+        step = min(step, math.ceil(math.sqrt(npartitions)), 100)
```
mrocklin, line 180: My guess is that 100 is too high here, and that if we are genuinely collecting from 100 different files in one task that we'll want more parallelism.  I would probably aim to a maximum task-reduction factor of 10 or 20 or so.  

I'm happy to defer this discussion to later though with some benchmarking if it comes up.

mrocklin (approved):

dask_expr/io/io.py
phofl: Yeah I'd rather do some benchmarking before lowering this too drastically, 100 also means that we have at least 10k files, which is probably a large dataset or many many small files

PR #411, opened 2023-11-21
Raise for unsupported engine and pop from kwargs

phofl:
keeping the engine in the kwargs passes it along to arrow which raises

### DIFF
--- a/dask_expr/io/parquet.py
+++ b/dask_expr/io/parquet.py
@@ -187,14 +187,11 @@ def to_parquet(
     schema="infer",
     name_function=None,
     filesystem=None,
+    engine=None,
     **kwargs,
 ):
     from dask_expr._collection import new_collection
     from dask_expr.io.parquet import NONE_LABEL, ToParquet
-
-    engine = kwargs.pop("engine", None)
-    if engine == "fastparquet":
-        raise NotImplementedError("Fastparquet engine is not supported")
 
     engine = _set_parquet_engine(engine=engine, meta=df._meta)
     compute_kwargs = compute_kwargs or {}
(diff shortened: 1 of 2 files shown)


dask_expr/io/tests/test_io.py
```diff
@@ -328,6 +328,13 @@ def test_to_parquet(tmpdir, write_metadata_file):
         df2.to_parquet(tmpdir, overwrite=True)
 
 
+def test_to_parquet_engne(tmpdir):
```
rjzamora, line 331:
```suggestion
def test_to_parquet_engine(tmpdir):
```

dask_expr/io/parquet.py
```diff
@@ -192,7 +192,11 @@ def to_parquet(
     from dask_expr._collection import new_collection
     from dask_expr.io.parquet import NONE_LABEL, ToParquet
 
-    engine = _set_parquet_engine(meta=df._meta)
+    engine = kwargs.pop("engine")
+    if engine == "fastparquet":
+        raise NotImplementedError("Fastparquet engine is not supported")
+
+    engine = _set_parquet_engine(engine=engine, meta=df._meta)
```
rjzamora, line 199: This seems fine. We could also include `engine=None` in the explicit argument list above and/or to move the "fastparquet" check into `_set_parquet_engine`.

rjzamora (approved):

dask_expr/io/parquet.py
```diff
@@ -192,7 +192,11 @@ def to_parquet(
     from dask_expr._collection import new_collection
     from dask_expr.io.parquet import NONE_LABEL, ToParquet
 
-    engine = _set_parquet_engine(meta=df._meta)
+    engine = kwargs.pop("engine")
```
rjzamora, line 195:
```suggestion
    engine = kwargs.pop("engine", None)
```

dask_expr/io/parquet.py
phofl: I like this

PR #419, opened 2023-11-22
Implement npartitions for merge

phofl:
sits on #416 and #418

### DIFF
--- a/dask_expr/_shuffle.py
+++ b/dask_expr/_shuffle.py
@@ -73,6 +73,8 @@ class Shuffle(Expr):
         to its necessary components.
     options: dict
         Algorithm-specific options.
+    index_shuffle : bool
+        Whether to perform the shuffle on the index.
     """
 
     _parameters = [
@@ -617,6 +619,8 @@ class AssignPartitioningIndex(Blockwise):
         New column name to assign.
     npartitions_out: int
         Number of partitions after repartitioning is finished.
+    assign_index : bool
+        Whether to use the index as partitioning column.
     """
 
     _parameters = [
@@ -631,6 +635,8 @@ class AssignPartitioningIndex(Blockwise):
     def operation(df, index, name: str, npartitions: int, assign_index):
         """Construct a hash-based partitioning index"""
         if assign_index:
+            # columns take precedence over index in _select_columns_or_index, so
+            # circumvent that, to_frame doesn't work because it loses the index
             index = df[[]]
             index["_index"] = df.index
         else:
--- a/dask_expr/_shuffle.py
+++ b/dask_expr/_shuffle.py
@@ -82,11 +82,13 @@ class Shuffle(Expr):
         "ignore_index",
         "backend",
         "options",
+        "index_shuffle",
     ]
     _defaults = {
         "ignore_index": False,
         "backend": None,
         "options": None,
+        "index_shuffle": None,
     }
     _is_length_preserving = True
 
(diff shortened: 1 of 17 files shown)


dask_expr/_shuffle.py
```diff
@@ -615,12 +619,23 @@ class AssignPartitioningIndex(Blockwise):
...
+    ]
 
     @staticmethod
-    def operation(df, index, name: str, npartitions: int):
+    def operation(df, index, name: str, npartitions: int, assign_index):
         """Construct a hash-based partitioning index"""
-        index = _select_columns_or_index(df, index)
+        if assign_index:
+            index = df[[]]
+            index["_index"] = df.index
```
rjzamora, line 635: Might be useful to drop a comment here to clarify that we need to convert the index to a frame because `partitioning_index` will ignore the index? (Only saying this because I was initially confused by this code block.

Note that we could probably do something like `index = df.index.to_frame()`, but I don't think it really matters.

dask_expr/_shuffle.py
```diff
@@ -615,12 +619,23 @@ class AssignPartitioningIndex(Blockwise):
         Number of partitions after repartitioning is finished.
     """
 
-    _parameters = ["frame", "partitioning_index", "index_name", "npartitions_out"]
+    _parameters = [
+        "frame",
+        "partitioning_index",
+        "index_name",
+        "npartitions_out",
+        "assign_index",
```
rjzamora, line 627: Can we add this to the doc-string (I think the name is fine, but is still a bit vague)?

dask_expr/_shuffle.py
```diff
@@ -82,11 +82,13 @@ class Shuffle(Expr):
         "ignore_index",
         "backend",
         "options",
+        "index_shuffle",
```
rjzamora, line 85: It would be helpful to document this operand as well.

dask_expr/tests/test_distributed.py
```diff
@@ -72,6 +72,24 @@ async def test_self_merge_p2p_shuffle(c, s, a, b):
...
+    )
+    pdf2 = lib.DataFrame(
+        {"b": [1, 2, 3, 4, 5, 6]}, index=lib.Index([1, 2, 7, 4, 5, 6], name="a")
+    )
+    df = from_pandas(pdf, npartitions=2, sort=False)
+    df2 = from_pandas(pdf2, npartitions=3, sort=False)
+
+    result = df.join(df2, shuffle_backend=shuffle)
+    x = await c.compute(result)
+    # assert result.npartitions == 6
```
rjzamora, line 89: Remove comment?

rjzamora (commented):
This seems fine so far - Only a few minor comments.

dask_expr/_shuffle.py
phofl: Yeah comment is a good idea


```
index = df.index.to_frame()
```

That doesn't work (was my first idea as well), but to_frame loses the index which will cause alignment issues later on...

rjzamora: Okay cool, thanks for checking

dask_expr/_shuffle.py
phofl: yep

dask_expr/_shuffle.py
phofl: yep

dask_expr/tests/test_distributed.py
phofl: Oh yeah that needs an update

PR #420, opened 2023-11-22
Add compatibility for keywords

phofl:
I don't have a good idea what we could do here. Renaming keywords makes testing with the dask test suite a PITA, but I also don't want to introduce too much debt.

We might want to consider a deprecation period in dask/dask instead so that we can switch over fresh to dask-expr when we enable dask-expr by default?

### DIFF
--- a/dask_expr/_collection.py
+++ b/dask_expr/_collection.py
@@ -681,6 +681,7 @@ class DataFrame(FrameBase):
         suffixes=("_x", "_y"),
         indicator=False,
         shuffle_backend=None,
+        **kwargs,
     ):
         """Merge the DataFrame with another DataFrame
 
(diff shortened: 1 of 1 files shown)


mrocklin:
Are there keywords aside from `shuffle=` that this protects against?  Why is there a mismatch between dask/dask and dask-expr in this case?

phofl:
Only shuffle for now, 

Some keywords were renamed when re-implementing stuff in dask-expr, most of them happened before my involvement, so not sure about the reasoning (that said I generally in favour of cleaning up names), cc @rjzamora

mrocklin:
My recommendation would be to just align dask-expr to dask/dask for these cases for now and use the names that we don't like, making a note of what we want to change.  Then when we flip the switch we make a bunch of changes back to the names that we like.  (better to do one big change than two medium changes)

PR #424, opened 2023-11-24
Change default parquet compression format from Snappy to LZ4

mrocklin:
Snappy's status as default is maybe just due to history.  Snappy had better Java support and LZ4 wasn't always available in systems like Spark.  Today Spark and other systems support LZ4 as well, and LZ4 generally performs a bit better, especially on decompression.

This is a significant change, but I think the only reason not to do it is historical, which I think maybe isn't a good enough reason these days.

### DIFF
--- a/dask_expr/io/parquet.py
+++ b/dask_expr/io/parquet.py
@@ -173,7 +173,7 @@ class ToParquetBarrier(Expr):
 def to_parquet(
     df,
     path,
-    compression="snappy",
+    compression="lz4",
     write_index=True,
     append=False,
     overwrite=False,


mrocklin:
@rjzamora any concerns here?  (no urgency during the holiday)

rjzamora:
Just a side note that `LZ4` is technically "deprecated" in the Parquet spec, so we should also make sure that pyarrow is actually using the interoperable `LZ4_RAW` codec under the hood when the user specifies `compression="lz4"`.

mrocklin:
> It seems like there has been some commotion around the LZ4 Parquet codec recently

This came out of this issue: https://github.com/apache/arrow/issues/38389#issuecomment-1824684681

To give more context, most Parquet deserialization cost seems to be SNAPPY decompression.  The use of SNAPPY seems to be a significant performance bottleneck.  You might find that issue of interest.

> we should probably double check with the pyarrow folks that using compression="lz4" is not about to become a problem for some reason

To be clear, the fact that DuckDB doesn't support LZ4 already I think makes this a problem.

PR #428, opened 2023-11-27
Support pandas objects in merge

phofl:


dask_expr/_collection.py
```diff
@@ -1316,17 +1283,56 @@ def merge(
...
+
+    if not is_dask_collection(right):
+        if left_index and right_on:  # change to join on index
+            right = right.set_index(right[right_on])
+            right_on = None
+            right_index = True
+        right = from_pandas(right, npartitions=1)
+
+    left = left.expr
+    right = right.expr if isinstance(right, FrameBase) else from_pandas(right, 1).expr
```
rjzamora, line 1320:
```suggestion
    right = right.expr
```
Isn't it now impossible for `isinstance(right, FrameBase)` to return `True` now that you have added the `is_dask_collection` check above?

rjzamora (commented):
Where should the doc-string live in cases like this? Do we move it to the stand-alone `merge` function, or leave it in the `merge` method? Either way, we probably want to clarify that `left`/`right` may be a `pd.DataFrame` object.

dask_expr/_collection.py
phofl: Good point, thx

phofl:
Hm that's a good question, it would put it onto the collection, but no strong opionion

pandas uses something like shared docstrings for cases like this, but I don't like them at all

PR #429, opened 2023-11-27
Adjust class hierarchy for overlapping stuff

phofl:
Not sure what to think of this (looking into cumulative stuff, so this came up again)

MapOverlap behaves like Blockwise stuff if we look at the expression when trying to make decisions about alignment and things like this but Blockwise obviously isn't quite correct, since we need neighbouring partitions

Injecting another layer solves this kind of, but I don't want to end up in inheritance hell...

### DIFF
--- a/dask_expr/_expr.py
+++ b/dask_expr/_expr.py
@@ -1093,7 +1093,19 @@ class Literal(Expr):
         return self.value
 
 
-class Blockwise(Expr):
+class BlockwiseOverlapping(Expr):
+    """Super class that operates like a blockwise op
+
+    We require information from neighboring partitions, so we can't prune partitions
+    before lowering, but the spec is the same as for Blockwise ops, we don't reoder
+    things, alignment stays consistent, ...
+
+    """
+
+    pass
+
+
+class Blockwise(BlockwiseOverlapping):
     """Super-class for block-wise operations
 
     This is fairly generic, and includes definitions for `_meta`, `divisions`,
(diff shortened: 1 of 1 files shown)


dask_expr/_expr.py
```diff
@@ -1093,7 +1093,19 @@ def _task(self, index: int):
...
+
+    We require information from neighboring partitions, so we can't prune partitions
+    before lowering, but the spec is the same as for Blockwise ops, we don't reoder
+    things, alignment stays consistent, ...
+
+    """
+
+    pass
+
+
```
mrocklin, line 1107: Why does this exist?  Is there some shared structure between these two that we need to handle?

My intuition here is just to have 

```python
class Blockwise(Expr):
    ...

class MapOverlap(Expr):
    ...
```

At least from this PR it doesn't seem like we're using the intermediate classes.

mrocklin: My recommendation would be to wait until we have a demonstrable need, and even in that case to consider using `(Blockwise, MapOverlap)` instead if possible.

dask_expr/_expr.py
```diff
@@ -2454,7 +2489,7 @@ def non_blockwise_ancestors(expr):
         e = stack.pop()
         if isinstance(e, IO):
             yield e
-        elif isinstance(e, Blockwise):
+        elif isinstance(e, BlockwiseOverlapping):
```
mrocklin, line 2492: If this is the only application, then this seems lighter weight to me

```suggestion
        elif isinstance(e, (Blockwise, MapOverlap)):
```

PR #431, opened 2023-11-28
Add split_out support to groupby nunique

phofl:
split_out=True is probably a more sensible default for nunqiue

### DIFF
--- a/dask_expr/tests/test_groupby.py
+++ b/dask_expr/tests/test_groupby.py
@@ -85,6 +85,7 @@ def test_groupby_nunique(df, pdf):
         df.groupby("x").nunique()
 
     assert_eq(df.groupby("x").y.nunique(), pdf.groupby("x").y.nunique())
+    assert_eq(df.groupby("x").y.nunique(split_out=True), pdf.groupby("x").y.nunique())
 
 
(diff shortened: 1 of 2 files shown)


dask_expr/tests/test_groupby.py
```diff
@@ -84,7 +84,8 @@ def test_groupby_nunique(df, pdf):
     with pytest.raises(AssertionError):
         df.groupby("x").nunique()
 
-    assert_eq(df.groupby("x").y.nunique(), pdf.groupby("x").y.nunique())
+    assert_eq(df.groupby("x").y.nunique(split_out=1), pdf.groupby("x").y.nunique())
+    assert_eq(df.groupby("x").y.nunique(split_out=True), pdf.groupby("x").y.nunique())
```
rjzamora, line 88: Not a blocker, but I think it might be nice to expand these checks slightly. For example, we may want to validate the actual partition count. E.g.

```python
assert df.groupby("x").y.nunique().npartitions == df.npartitions
```

Also, can we check that `split_out: Callable` is working?

rjzamora (approved):
Suggested a bit more test coverage, but LGTM

PR #433, opened 2023-11-28
Implement cumulative aggregation

phofl:


dask_expr/_cumulative.py
```diff
@@ -0,0 +1,117 @@
...
+def _take_last(a, skipna=True):
+    if skipna:
+        a = a.bfill()
+    return a.tail(n=1).squeeze()
+
+
+class TakeLast(Blockwise):
+    _parameters = ["frame", "skipna"]
+    operation = staticmethod(_take_last)
+    _projection_passthrough = True
```
crusaderky, line 62:
```suggestion
class TakeLast(Blockwise):
    _parameters = ["frame", "skipna"]
    _projection_passthrough = True
    
    @staticmethod
    def operation(a, skipna=True):
        if skipna:
            a = a.bfill()
        return a.tail(n=1).squeeze()
```

dask_expr/tests/test_collection.py
```diff
@@ -242,6 +242,15 @@ def test_std_kwargs(axis, skipna, ddof):
...
 
+@pytest.mark.parametrize("func", ["cumsum", "cumprod", "cummin", "cummax"])
+def test_cumulative_methods(df, pdf, func):
+    assert_eq(getattr(df, func)(), getattr(pdf, func)(), check_dtype=False)
+    assert_eq(getattr(df.x, func)(), getattr(pdf.x, func)())
+
+    q = getattr(df, func)()["x"]
+    assert q.simplify()._name == getattr(df.x, func)()
+
+
```
crusaderky, line 253: Could you add a test for skipna=False (running on data that does include NA's)?

crusaderky (changes requested):

dask_expr/tests/test_collection.py
phofl: added

phofl:
dask broke the ci

PR #443, opened 2023-11-30
Make groupby over Index work

phofl:


dask_expr/_groupby.py
```diff
@@ -785,9 +785,12 @@ def __init__(
...
             self.by = [by] if np.isscalar(by) else list(by)
             for key in self.by:
-                if not (np.isscalar(key) and key in self.obj.columns):
+                if (
+                    not (np.isscalar(key) and key in self.obj.columns)
+                    and not key == self.obj._meta.index.name
+                ):
                     raise NotImplementedError(
-                        "Can only group on column names (for now)."
+                        f"Key={key} neither in columns nor in the index."
```
milesgranger, line 793: Seems like a KeyError, now. Not the current NotImplementedError?

phofl: Good catch, thx. 

I just removed the whole check, this surfaces the pandas error which is what we would want anyway

PR #444, opened 2023-11-30
Implement broadcast merge

phofl:


dask_expr/tests/test_distributed.py
```diff
@@ -90,6 +91,28 @@ async def test_merge_index_precedence(c, s, a, b, shuffle):
     lib.testing.assert_frame_equal(x.sort_index(ascending=False), pdf.join(pdf2))
 
 
+@gen_cluster(client=True)
+@pytest.mark.parametrize("shuffle", ["tasks", "p2p"])
+@pytest.mark.parametrize("broadcast", [True, 0.6])
```
rjzamora, line 96: We will probably want to do a parameter sweep over `how` as well (since the algorithm is sometimes dependent on this).

phofl: good point, thx

dask_expr/_collection.py
```diff
@@ -726,6 +727,9 @@ def merge(
             Shuffle backend to use if shuffling is necessary.
         npartitions : int, optional
             The number of output partitions
+        broadcast : float, bool, optional
+            Whether to perform a broadcast merge if bool, or the broadcast bias if
```
hendrikmakait, line 731: What's the broadcast bias?

dask_expr/_merge.py
```diff
@@ -103,13 +108,44 @@ def _divisions(self):
...
+        broadcast_bias, broadcast = 0.5, None
+        broadcast_side = self.broadcast_side
+        if isinstance(self.broadcast, float):
+            broadcast_bias = self.broadcast
+        elif isinstance(self.broadcast, bool):
+            broadcast = self.broadcast
+
+        s_backend = self.shuffle_backend or get_default_shuffle_method()
+        if (
+            s_backend in ("tasks", "p2p")
```
fjetter, line 138: I assume you are just copying dask/dask logic but in the case where the shuffle backend is not explicitly set and all other conditions evaluate to true, I would expect it to be beneficial to switch from disk to broadcast

fjetter (approved):

dask_expr/_merge.py
```diff
@@ -194,6 +228,41 @@ def _lower(self):
...
+                        left = Repartition(left, new_partitions=self._npartitions)
+                    else:
+                        right = Repartition(right, new_partitions=self._npartitions)
+
+                if self.how != "inner":
+                    if self.broadcast_side == "left":
+                        left = Shuffle(
+                            left,
+                            shuffle_left_on,
+                            backend="tasks",
```
fjetter, line 243: please don't

hendrikmakait, line 243:
```suggestion
```

dask_expr/_merge.py
```diff
@@ -194,6 +228,41 @@ def _lower(self):
...
+                            npartitions_out=right.npartitions,
+                        )
+                    else:
+                        right = Shuffle(
+                            right,
+                            shuffle_right_on,
+                            backend="tasks",
```
hendrikmakait, line 250:
```suggestion
```

dask_expr/_merge.py
phofl: we can't end up with disk from the ``get_default_shuffle_method``, so we will only end up with disk if it's explicitly given

dask_expr/_merge.py
phofl: updated those

dask_expr/_collection.py
phofl: copied the docstring over from dask/dask now

PR #452, opened 2023-12-05
Implement `.mask`

hendrikmakait:


dask_expr/_collection.py
```diff
@@ -591,6 +591,11 @@ def isna(self):
     def isnull(self):
         return new_collection(self.expr.isnull())
 
+    def mask(self, cond, other=np.nan):
```
hendrikmakait, line 594: Note: IIUC, extension types would default to `pd.NA` so using `np.nan` isn't perfect.

phofl (approved):

PR #455, opened 2023-12-05
Add not_null and isnull

fjetter:
Anything I can test for the optimization flags? (mostly for training purposes, not necessarily for correctness)


phofl:
What we generally do is:

```
q = df.notnull()["a"]
expected = df["a"].notnull()

assert q.optimize()._name == expected.optimize()._name
```

PR #458, opened 2023-12-05
Index.drop_duplicates()

crusaderky:


dask_expr/_shuffle.py
```diff
@@ -1034,6 +1034,10 @@ def _lower(self):
         )
         return SortIndexBlockwise(index_set)
 
+    @property
+    def npartitions(self):
```
phofl, line 1038: You don't need this, the parent ``npartitions`` implementation does the same thing

phofl (approved):

PR #460, opened 2023-12-05
Implement rolling groupby

fjetter:
So, I'm just expanding the existing `Rolling` because the groupby rolling is actually doing the same thing as the ordinary rolling just with a local groupby. The implementation feels very unclean. I'll review the class hierarchy again if there is a more elegant way to modify the logic here. Any pointers appreciated


dask_expr/_rolling.py
```diff
@@ -49,7 +74,19 @@ def kwargs(self):
...
+            by = self.groupby_kwargs.get("by", []) if self.groupby_kwargs else []
+            by_columns = by if not isinstance(by, Expr) else []
+            columns = sorted(set(parent.columns + by_columns))
+            if columns == self.frame.columns:
+                return
+            if self.groupby_kwargs is not None:
+                return type(parent)(
+                    type(self)(self.frame[columns], *self.operands[1:]),
+                    *parent.operands[1:],
+                )
```
fjetter, line 86: I'm honestly not even sure what this is doing and it does feel like we're duplicating work but this is what I found for the groupby implementations and this seems to work.

phofl: This drops all columns except the parent columns and the by columns from the df before the groupby and then drops the by column after the groupby if not in parent.columns

Yes this is needed for groupby to avoid dropping too many columns

dask_expr/tests/test_rolling.py
```diff
@@ -27,16 +27,16 @@ def df(pdf, request):
...
-        ("sum", ()),
-        ("min", ()),
-        ("max", ()),
-        ("var", ()),
-        ("std", ()),
-        ("median", ()),
-        ("skew", ()),
-        ("quantile", (0.5,)),
-        ("kurt", ()),
+        # ("mean", ()),
```
phofl, line 30: Are those failing?

dask_expr/tests/test_groupby.py
```diff
@@ -336,3 +336,43 @@ def test_groupby_var_dropna_observed(dropna, observed, func):
...
+            "group1": 5 * ["g" + str(i) for i in range(120)],
+        },
+        index=lib.date_range("20190101", periods=60).repeat(10),
+    )
+
+    ddf = from_pandas(df, npartitions=8)
+
+    actual = ddf.groupby("group1").rolling("1D").sum()["column1"]
+    expected = df.groupby("group1").rolling("1D").sum()["column1"]
+    actual.optimize()
```
phofl, line 377: Ideally, we would test this against 

```
df[["column1", "group1"]].groupby("group1").rolling("1D").sum()["column1"].optimize()._name
```

fjetter: done

phofl (approved):

phofl:
thx

PR #469, opened 2023-12-05
Make slicing a series raise a not implemented error

### LINKED ISSUE #463: `Series.__getitem__` always returns Scalar
`Series[...]` should return either a Scalar or a Series depending on the argument.
Instead, it always returns a Scalar:

```python
>>> import dask_expr as dx
>>> import pandas
>>> s = dx.from_pandas(pandas.Series(range(10)))
>>> s
<dask_expr.expr.Series: expr=df>
>>> s2 = s[:4]
>>> s2
<dask_expr.expr.Scalar: expr=df[slice(None, 4, None)]>
```
`compute()` correctly returns a pd.Series:
```python
>>> s2.compute()
0    0
1    1
2    2
3    3
dtype: int64
```

phofl:
closes #463

cc @crusaderky

### DIFF
--- a/dask_expr/_collection.py
+++ b/dask_expr/_collection.py
@@ -154,6 +154,8 @@ class FrameBase(DaskMethodsMixin):
     def __getitem__(self, other):
         if isinstance(other, FrameBase):
             return new_collection(self.expr.__getitem__(other.expr))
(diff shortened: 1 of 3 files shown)


rjzamora (approved):
Thanks!

PR #470, opened 2023-12-05
Split out graph Expr code from Dataframe Expr code

### LINKED ISSUE #142: Remove DataFrame assumptions from Expr

mrocklin:
Closes #142
Supercedes #158

This is the smallest step I could think of to start this change.  I may want to play around a little with arrays this week in anticipation of hanging out with @dcherian next week.  I figured it would be good to split this off now if possible.

(sorry for the poor timing, I know it's an active week for this repository)

cc @phofl

### DIFF
--- a/dask_expr/_core.py
+++ b/dask_expr/_core.py
@@ -1,0 +1,642 @@
+from __future__ import annotations
+
+import functools
+import os
+from collections.abc import Generator
+
+import dask
+import pandas as pd
+import toolz
+from dask.dataframe.core import is_dataframe_like, is_index_like, is_series_like
+from dask.utils import funcname, import_required, is_arraylike
+
+from dask_expr._util import _BackendData, _tokenize_deterministic, _tokenize_partial
+
+
+class Expr:
+    _parameters = []
+    _defaults = {}
+
+    def __init__(self, *args, **kwargs):
+        operands = list(args)
+        for parameter in type(self)._parameters[len(operands) :]:
+            try:
+                operands.append(kwargs.pop(parameter))
+            except KeyError:
+                operands.append(type(self)._defaults[parameter])
+        assert not kwargs, kwargs
+        self.operands = operands
+        if self._required_attribute:
+            dep = next(iter(self.dependencies()))._meta
+            if not hasattr(dep, self._required_attribute):
+                # Raise a ValueError instead of AttributeError to
+                # avoid infinite recursion
+                raise ValueError(f"{dep} has no attribute {self._required_attribute}")
+
+    @property
+    def _required_attribute(self) -> str:
+        # Specify if the first `dependency` must support
+        # a specific attribute for valid behavior.
+        return None
+
+    def __str__(self):
+        s = ", ".join(
+            str(param) + "=" + str(operand)
+            for param, operand in zip(self._parameters, self.operands)
+            if isinstance(operand, Expr) or operand != self._defaults.get(param)
+        )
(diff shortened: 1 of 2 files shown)


mrocklin:
```python
from dask_expr import _core as core, _expr as expr
everything = set(dir(object))
c = set(dir(core.Expr)) - everything
d = set(dir(expr.Expr)) - everything - c
```

### What's in core.Expr

```
{'__dask_graph__',
 '__dask_keys__',
 '__dict__',
 '__module__',
 '__weakref__',
 '_defaults',
 '_depth',
 '_find_similar_operations',
 '_layer',
 '_lower',
 '_meta',
 '_name',
 '_node_label_args',
 '_parameters',
 '_remove_operations',
 '_required_attribute',
 '_simplify_down',
 '_simplify_up',
 '_task',
 '_to_graphviz',
 '_tree_repr_lines',
 'dependencies',
 'find_operations',
 'lower_completely',
 'lower_once',
 'operand',
 'pprint',
 'rewrite',
 'simplify',
 'substitute',
 'substitute_parameters',
 'tree_repr',
 'visualize',
 'walk'}
```

### What's in expr.Expr

```python
{'__add__',
 '__and__',
 '__bool__',
 '__getattr__',
 '__getitem__',
 '__invert__',
 '__mul__',
 '__neg__',
 '__or__',
 '__pos__',
 '__radd__',
 '__rand__',
 '__rge__',
 '__rgt__',
 '__rle__',
 '__rlt__',
 '__rmul__',
 '__ror__',
 '__rsub__',
 '__rtruediv__',
 '__rxor__',
 '__sub__',
 '__truediv__',
 '__xor__',
 '_combine_similar',
 '_combine_similar_branches',
 '_divisions',
 '_is_length_preserving',
 'abs',
 'align',
 'all',
 'any',
 'apply',
 'astype',
 'clip',
 'columns',
 'combine_first',
 'combine_similar',
 'count',
 'divisions',
 'dtypes',
 'fillna',
 'idxmax',
 'idxmin',
 'index',
 'isna',
 'isnull',
 'known_divisions',
 'mask',
 'max',
 'mean',
 'memory_usage_per_partition',
 'min',
 'mode',
 'nbytes',
 'ndim',
 'npartitions',
 'nunique_approx',
 'optimize',
 'prod',
 'rename_axis',
 'replace',
 'round',
 'size',
 'std',
 'sum',
 'to_timestamp',
 'var',
 'where'}
```

mrocklin:
There are other things we could think about doing, like renaming `_expr.Expr` to `Frame` or something, or moving stuff within a `dataframe` directory, but I think that all of that could also come later and doesn't necessarily need to block this first step.  My hope is that this first step is pretty innocuous.

phofl:
Do you have an idea how we can solve the issue that defining things like addition and multiplication on the expression gets more complex if we split this up? This is why we postponed the initial pr IIRC

phofl:
Beforehand meaning before calling this on the expression level or on the expression level? I think this is better suited on the collection level generally speaking.

phofl (approved):

PR #473, opened 2023-12-06
Support `freq` and `axis=1` parameters in `shift` method

milesgranger:


dask_expr/_collection.py
```diff
@@ -644,10 +644,7 @@ def shift(self, periods=1, freq=None, axis=0):
             raise TypeError("periods must be an integer")
 
         axis = _validate_axis(axis)
-        if axis == 1:
-            raise NotImplementedError("shift on axis 1 not supported yet")
-
-        return new_collection(Shift(self.expr, periods, freq))
+        return new_collection(Shift(self.expr, periods, freq, axis))
```
phofl, line 647: Please create a different expression class that handles axis=1

dask_expr/_util.py
```diff
@@ -19,6 +20,41 @@
...
+        freq = pd.tseries.frequencies.to_offset(freq)
+
+    is_offset = isinstance(freq, pd.DateOffset)
+    if is_offset:
+        if freq.is_anchored() or not hasattr(freq, "delta"):
+            # Can't infer divisions on relative or anchored offsets, as
+            # divisions may now split identical index value.
+            # (e.g. index_partitions = [[1, 2, 3], [3, 4, 5]])
+            # return df.clear_divisions()
+            return None
```
phofl, line 48: Should this result In all None divisions?

milesgranger: Depends on the caller, for _divisions it'll return from here or use the existing .frame.divisions. If it's dispatching to MapPartitions then it'll set `clear_divisions` if this returns None.

phofl: Yikes ok

milesgranger: Updated the `_divisions` method when this returns none.
https://github.com/dask-contrib/dask-expr/pull/473/commits/676330ef2cb6d3313873f797625f4bc5d40eae48

dask_expr/_collection.py
milesgranger: done, maybe you might like the `ShiftAxis0/1` naming though? https://github.com/dask-contrib/dask-expr/pull/473/commits/677594793981756e669f049253b74b29f74d4efb

dask_expr/_expr.py
```diff
@@ -2787,7 +2796,20 @@ def _lower(self):
         return None
 
     def _simplify_down(self):
-        if self.freq is None:
+        if self.freq:
```
phofl, line 2799: How is this MapPartitions and why don't we need data from previous partitions?

dask_expr/_expr.py
```diff
@@ -2800,8 +2822,21 @@ def _simplify_down(self):
                 enforce_metadata=self.enforce_metadata,
                 kwargs=self.kwargs,
             )
-        else:
-            raise NotImplementedError()
+
+
+class ShiftAxis1(ShiftAxis0):
```
phofl, line 2827: Can you call map_partitions on the collection level?

We might not need a new class for now

dask_expr/_expr.py
milesgranger: Just following the implementation here.. 
https://github.com/dask/dask/blob/0253207f8b3bc2a48e059a856783a4fb7c4eb850/dask/dataframe/core.py#L2197, do you suspect that is wrong?

dask_expr/_expr.py
milesgranger: something like this? https://github.com/dask-contrib/dask-expr/pull/473/commits/27e37157e714d2e8af2e74e82aeb017398802e76

dask_expr/_collection.py
```diff
@@ -662,10 +662,7 @@ def shift(self, periods=1, freq=None, axis=0):
...
```
phofl, line 665: No, please call map_partitions on this level, don't even go into the Shift expression for axis=1

milesgranger: Ah, I see now. https://github.com/dask-contrib/dask-expr/pull/473/commits/5015f8f23d1698b03ebb40c3c3056ec533e627b3

dask_expr/_expr.py
```diff
@@ -2767,17 +2772,21 @@ class Shift(MapOverlap):
     enforce_metadata = True
     before = 0
     after = 0
+    axis = 0
```
phofl, line 2775: This shouldn't be needed

dask_expr/_collection.py
```diff
@@ -662,10 +668,19 @@ def shift(self, periods=1, freq=None, axis=0):
...
 
         axis = _validate_axis(axis)
-        if axis == 1:
-            raise NotImplementedError("shift on axis 1 not supported yet")
-
-        return new_collection(Shift(self.expr, periods, freq))
+        if axis == 0:
+            return new_collection(Shift(self.expr, periods, freq))
+
+        clear_divisions = _calc_maybe_new_divisions(self.expr, periods, freq) is None
```
phofl, line 674: We generally want to avoid accessing the divisions as long as possible. Can we just set those to None for now and add a todo?

phofl (approved):

PR #477, opened 2023-12-06
Fix meta for reduction scalars

fjetter:
So, it turns out that https://github.com/dask/dask/pull/10671 is not necessarily a problem. The problem why the reduction test originally failed is because the `Scalar` object had a python bool as it's meta instead of a numpy bool. Dask's assert logic has a special path for numpy dtypes that implements a more forgiving equality check I ran into here.

I suspect this is something we'll run into many times when comparing with the dask/dask test suite. This feels like something `make_meta` should deal with already properly. I'll poke around dask/dask if we can fix it at the source.


fjetter:
hmm this fix is not working well it seems. I think https://github.com/dask/dask/pull/10677 will be better

PR #481, opened 2023-12-06
Add `.get_partition(n)` for parity with `dask/dask`

hendrikmakait:


dask_expr/tests/test_collection.py
```diff
@@ -798,6 +798,13 @@ def test_partitions(pdf, df):
     assert_eq(out, pdf.iloc[10:20])
 
 
+def test_get_partition(pdf, df):
+    assert_eq(df.get_partition(0), pdf.iloc[:10])
+    assert_eq(df.get_partition(1), pdf.iloc[10:20])
+    assert_eq(df.get_partition(-1), pdf.iloc[90:])
```
hendrikmakait, line 804: `dask/dask` fails on negative indices, but I see no reason for doing so here.

phofl (approved):

phofl:
thx

PR #487, opened 2023-12-07
Add GroupBy.ffill and bfill

phofl:


dask_expr/tests/test_groupby.py
```diff
@@ -381,6 +381,19 @@ def test_groupby_median(df, pdf):
     assert_eq(df.groupby("x").median()["y"], pdf.groupby("x").median()["y"])
 
 
+def test_groupby_ffill_bfill(pdf):
+    pdf["y"] = pdf["y"].astype("float64")
+    pdf.loc[pdf.index % 2 == 0, "y"] = np.nan
+    df = from_pandas(pdf, npartitions=10)
```
crusaderky, line 387: y is either always NaN or always non-NaN within each group, which makes ffill/bfill a no-op.
This passes:
```python
    assert_eq(pdf.groupby("x").ffill().sort_index(), pdf[["y", "z"]])
    assert_eq(pdf.groupby("x").bfill().sort_index(), pdf[["y", "z"]])
```

I think it's also important that the test ensures that we're testing the use case of cross-partition ffill and bfill.

We should also make sure that a change in the pdf fixture in the future won't accidentally neuter the unit test by making these two use cases disappear.

dask_expr/_groupby.py
```diff
@@ -735,6 +735,30 @@ def grp_func(self):
...
+        if isinstance(parent, Projection):
+            by_columns = self.by if not isinstance(self.by, Expr) else []
+            columns = sorted(set(parent.columns + by_columns))
+            if columns == self.frame.columns:
+                return
+            columns = [col for col in self.frame.columns if col in columns]
+            return type(parent)(
+                type(self)(self.frame[columns], *self.operands[1:]),
+                *parent.operands[1:],
+            )
```
crusaderky, line 755: There is no test coverage for this method.

crusaderky (changes requested):

dask_expr/_groupby.py
phofl: thx, addressed now

dask_expr/_groupby.py
```diff
@@ -753,6 +753,30 @@ def grp_func(self):
...
 
+def _fillna(group, *, what, **kwargs):
+    return getattr(group, what)(**kwargs)
+
+
+class GroupByBFill(GroupByTransform):
+    func = staticmethod(functools.partial(_fillna, what="bfill"))
+
+    def _simplify_up(self, parent):
+        if isinstance(parent, Projection):
```
crusaderky, line 764: if I replace this line with `if True:` all tests remain green

dask_expr/tests/test_groupby.py
```diff
@@ -383,6 +383,25 @@ def test_groupby_median(df, pdf):
...
+
+    actual = df.groupby("x")["y"].ffill()
+    expect = df[["x", "y"]].groupby("x")["y"].ffill()
+    assert actual.optimize()._name == expect.optimize()._name
+    assert_eq(actual, pdf.groupby("x")["y"].ffill())
+
+    actual = df.groupby("x").ffill()["y"]
+    expect = df[["x", "y"]].groupby("x").ffill()["y"]
+    assert actual.optimize()._name == expect.optimize()._name
+    assert_eq(actual, pdf.groupby("x")["y"].ffill())
```
crusaderky, line 402: I tried testing that `.optimize()._name` is the same across all of these:

- `df.groupby("x")["y"].ffill()`
- `df.groupby("x").ffill()["y"]`
- `df[["x", "y"]].groupby("x")["y"].ffill()`
- `df[["x", "y"]].groupby("x").ffill()["y"]`


but that's not the case. Is it expected?

dask_expr/_groupby.py
phofl: I am not too worried about this, we can add a general test for all groupby methods in a follow up though

dask_expr/tests/test_groupby.py
phofl: Yes this behaves a little different. The first caste reduces the groupby op to a series while the second does the groupby on a DataFrame with one column and reduces to a series in the end. 

Making this consistent is not worth the effort for now, since it is the same behavior and performance wise

phofl: The slice is not pushed into groupby, which is the difference that you are seeing here

crusaderky (approved):

PR #522, opened 2023-12-12
Add method operators to DataFrame and Series

### LINKED ISSUE #500: Add add, mul methods
Dask supports ``__add__`` and ``DataFrame.add``, we don't have the latter

milesgranger:
Will close #500 

Adds the following method operators to `Series` and `DataFrame`

  - add
  - div
  - divide
  - floordiv
  - mod
  - mul
  - pow
  - radd
  - rdiv
  - rfloordiv
  - rmod
  - rmul
  - rpow
  - rsub
  - rtruediv
  - sub
  - truediv


dask_expr/_collection.py
```diff
@@ -103,6 +103,62 @@ def _wrap_expr_op(self, other, op=None):
...
+        def method(self, other, axis="columns", level=None, fill_value=None):
+            if level is not None:
+                raise NotImplementedError("level must be None")
+            if not isinstance(other, DataFrame):
+                raise NotImplementedError("only other=DataFrame implemented")
+
+            axis = _validate_axis(axis)
+
+            if axis in (1, "columns"):
+                if isinstance(other, Series):
```
phofl, line 122: How can we hit this if we are a DataFrame?

dask_expr/_collection.py
```diff
@@ -103,6 +103,62 @@ def _wrap_expr_op(self, other, op=None):
...
+    """
+    Add method operators to Series or DataFrame like DataFrame.add.
+    _wrap_expr_method_operator("add", DataFrame)
+    """
+    if class_ == DataFrame:
+
+        def method(self, other, axis="columns", level=None, fill_value=None):
+            if level is not None:
+                raise NotImplementedError("level must be None")
+            if not isinstance(other, DataFrame):
```
phofl, line 116: Why is this not implemented?

dask_expr/_collection.py
```diff
@@ -103,6 +103,62 @@ def _wrap_expr_op(self, other, op=None):
...
+                    right=other.expr,
+                    axis=axis,
+                    level=level,
+                    fill_value=fill_value,
+                )
+            )
+
+    elif class_ == Series:
+
+        def method(self, other, level=None, fill_value=None, axis=0):
```
phofl, line 138: I think this should raise for axis=columns instead of the thing above?

dask_expr/tests/test_collection.py
```diff
@@ -341,6 +341,58 @@ def test_boolean_operators(func):
...
+        "radd",
+        "rsub",
+        "rmul",
+        "rdiv",
+        "rtruediv",
+        "rfloordiv",
+        "rmod",
+        "rpow",
+    ),
+    ids=lambda op: f"{op=}",
```
phofl, line 372: why do you need ids here?

dask_expr/_collection.py
milesgranger: Maybe.. although I'm basing my opinion [from the current reference impl](https://github.com/dask/dask/blob/191d39177009d2cce25b818878118e35329b6db3/dask/dataframe/core.py#L4431), but that a series (self) can be used with both columns and index axis as it works with both so long as they're the same size. Does that seem correct?

dask_expr/_collection.py
milesgranger: The `other` might not be though?

dask_expr/tests/test_collection.py
milesgranger: Helped for debugging.. I can remove if they're undesired.

dask_expr/_collection.py
phofl: Oh no that's fine, can you add a todo though?

dask_expr/_collection.py
phofl: ok that's interesting, happy to leave as is then

dask_expr/tests/test_collection.py
milesgranger: Gone https://github.com/dask-contrib/dask-expr/pull/522/commits/82cffb13c6b4873635f5caa647c80f581b38e669

dask_expr/_collection.py
milesgranger: https://github.com/dask-contrib/dask-expr/pull/522/commits/01b1517405883124dac7848cb495beb986d42eb9 done, will follow up tomorrow.

phofl:
thx

Issue #525, opened 2023-12-12
Working towards getting the dask-ci green with dask-expr


phofl:
I started going through the ci and I am either filing tickets or fixing the things I find immediately. Will Link to the PRs here.

PR #526, opened 2023-12-12
Add keep to drop duplicates

phofl:
ref #525

### DIFF
--- a/dask_expr/_collection.py
+++ b/dask_expr/_collection.py
@@ -1640,7 +1640,7 @@ class Series(FrameBase):
             return uniqs.size
 
     def drop_duplicates(
-        self, ignore_index=False, split_every=None, split_out=True, keep="first"
+        self, ignore_index=False, split_every=None, split_out=True, keep=None
     ):
         if keep is False:
             raise NotImplementedError("drop_duplicates with keep=False")
--- a/dask_expr/_collection.py
+++ b/dask_expr/_collection.py
@@ -27,6 +27,7 @@ from dask.dataframe.dispatch import meta_nonempty
 from dask.dataframe.utils import has_known_categories, index_summary
 from dask.utils import (
     IndexCallable,
+    get_default_shuffle_method,
     memory_repr,
     put_lines,
     random_state_data,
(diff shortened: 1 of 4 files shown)


rjzamora (commented):
Just a quick note that I'm skeptical that `keep="first"` and `keep="last"` are actually "working".  When we perform the shuffle, there is no explicit guarantee that the original ordering will be preserved. I have a suspicion that we are just getting lucky.

phofl:
We generally preserve row order with task based or p2p shuffle, but wouldn’t guarantee this

rjzamora:
>We generally preserve row order with task based or p2p shuffle, but wouldn’t guarantee this

Right exactly. I do think "tasks" will always preserve order, but it seems like "p2p" may depend on the order that messages are received/buffered by a worker (maybe/hopefully I'm wrong about this?).

rjzamora:
**Update**: Ran some small tests, and it does look like "p2p" will produce wrong results here:

```python
import dask
from dask.distributed import LocalCluster, Client
import dask.dataframe as dd
from dask.datasets import timeseries

client = Client(LocalCluster())
shuffle = "p2p"

df = timeseries()
result_pd = df.compute().drop_duplicates(subset=["name"], keep="first")
result_dd = df.drop_duplicates(subset=["name"], keep="first", split_out=df.npartitions, shuffle=shuffle).compute()

# Fails for shuffle = "p2p"
# Passes for shuffle = "tasks"
dd.assert_eq(result_pd, result_dd)
```

I realize this is more of a `dask.dataframe` issue than a `dask-expr` issue. I'll raise an issue over there to discuss how this behavior should be handled and/or communicated.

hendrikmakait:
Just to assert what you already figured out empirically: P2P shuffling has no ordering guarantees.

rjzamora:
>I am not too concerned about this, users can still request split_out=1

Yeah, I'm assuming we have similar feelings, but just to be explicit. The two things we definitely want to avoid:

1. We silently return incorrect results
2. The code crashes

My loosely-held opinion on this: In the short term, we should adopt `how=None` and warn the user if they use `how="first"/"last"` when "p2p" is active. In the long term, figure out a simple/reliable way to **optionally** enforce stable ordering in "p2p" for cases like this.

dask_expr/_collection.py
```diff
@@ -1622,13 +1639,26 @@ def nunique(self, dropna=True):
         else:
             return uniqs.size
 
-    def drop_duplicates(self, ignore_index=False, split_every=None, split_out=True):
+    def drop_duplicates(
+        self, ignore_index=False, split_every=None, split_out=True, keep="first"
```
milesgranger, line 1643: Curious why keep=None in DataFrame and keep="first" in Series?

milesgranger (approved):
One question, but otherwise looks goodie!

dask_expr/_collection.py
phofl: that was a mistake, thanks for catching

milesgranger:
Thanks @phofl!

PR #541, opened 2023-12-13
Add test that repartition with equal divisions is a noop

phofl:
ref #525

### DIFF
--- a/dask_expr/tests/test_repartition.py
+++ b/dask_expr/tests/test_repartition.py
@@ -1,7 +1,7 @@
 import numpy as np
 import pytest
 
-from dask_expr import from_pandas
+from dask_expr import Repartition, from_pandas, repartition
 from dask_expr.tests._util import _backend_library, assert_eq
 
 lib = _backend_library()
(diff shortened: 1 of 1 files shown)


dask_expr/tests/test_repartition.py
```diff
@@ -34,6 +34,65 @@ def test_repartition_combine_similar(kwargs):
...
+    assert (
+        len(
+            [
+                x
+                for x in ds3.expr.optimize(fuse=False).walk()
+                if isinstance(x, Repartition)
+            ]
+        )
+        == 0
+    )
```
milesgranger, line 93:
```suggestion
    def assert_not_repartitions(expr, fuse=False):
        repartitions = [
            x for x in expr.optimize(fuse=fuse).walk() if isinstance(x, Repartition)
        ]
        assert len(repartitions) == 0

    # DataFrame method
    df2 = df.repartition(divisions=type_ctor(df.divisions))
    assert_not_repartitions(df2.expr)

    # Top-level dask.dataframe method
    df3 = repartition(df, divisions=type_ctor(df.divisions))
    assert_not_repartitions(df3.expr)

    # Series method
    ds2 = ds.repartition(divisions=type_ctor(ds.divisions))
    assert_not_repartitions(ds2.expr)

    # Top-level dask.dataframe method applied to a Series
    ds3 = repartition(ds, divisions=type_ctor(ds.divisions))
    assert_not_repartitions(ds3.expr)
```

Maybe slightly easier on the eyes?

milesgranger (approved):
Only a nit. :+1:

dask_expr/tests/test_repartition.py
phofl: oh yes, nice catch!

PR #542, opened 2023-12-13
Avoid string keys in shuffle

phofl:
ref #525


phofl:
There are tests that sort the keys, which fails when you mix tuples and strings, but that’s not really important

PR #548, opened 2023-12-14
Support Index.shift

phofl:
ref #525

### DIFF
--- a/dask_expr/tests/test_collection.py
+++ b/dask_expr/tests/test_collection.py
@@ -1872,6 +1872,52 @@ def test_keys(df, pdf):
     assert_eq(df.x.keys(), pdf.x.keys())  # Alias for Series.index
 
 
+@pytest.mark.parametrize("data_freq, divs1", [("B", False), ("D", True), ("h", True)])
+def test_shift_with_freq_datetime(pdf, data_freq, divs1):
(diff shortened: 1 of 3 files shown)


dask_expr/tests/test_collection.py
```diff
@@ -1872,6 +1872,52 @@ def test_keys(df, pdf):
...
+    df = from_pandas(pdf, npartitions=4)
+    for freq in ["s", lib.Timedelta(10, unit="h")]:
+        for d, p in [(df, pdf), (df.x, pdf.x), (df.index, pdf.index)]:
+            res = d.shift(2, freq=freq)
+            assert_eq(res, p.shift(2, freq=freq))
+            assert res.known_divisions
+    # Index shifts also work with freq=None
+    res = df.index.shift(2)
+    assert_eq(res, df.index.shift(2))
+    assert res.known_divisions
```
milesgranger, line 1918: Maybe a matter of preference, but looping in a test is less fun debugging than a full parameterized test. Then I know exactly what parameters caused a failure.

milesgranger (approved):

dask_expr/tests/test_collection.py
phofl: Yeah 100% agree, but didn't want to put too much work into it since this is a copy from dask/dask

PR #549, opened 2023-12-14
Add split_every and split out support

phofl:
ref #525


dask_expr/_reductions.py
```diff
@@ -967,7 +1000,7 @@ def aggregate(cls, results: list[pd.Series], dropna=None):
         [df] = results
         max = df.max(skipna=dropna)
         out = df[df == max].index.to_series().sort_values().reset_index(drop=True)
-        out.name = results[0].name
+        # out.name = results[0].name
```
rjzamora, line 1003: Are we removing this line?

phofl: Oh yes, frorgot to delete. I don't know why tests were passing before...

milesgranger (approved):

PR #554, opened 2023-12-14
Ensure that assign works with cumulative method

phofl:
ref #525

### DIFF
--- a/dask_expr/_expr.py
+++ b/dask_expr/_expr.py
@@ -2041,12 +2041,15 @@ def is_broadcastable(s):
 
 def non_blockwise_ancestors(expr):
     """Traverse through tree to find ancestors that are not blockwise or are IO"""
+    from dask_expr._cumulative import CumulativeAggregations
+
     stack = [expr]
     while stack:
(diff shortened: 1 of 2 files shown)


dask_expr/_expr.py
```diff
@@ -2041,12 +2041,15 @@ def is_broadcastable(s):
...
+    from dask_expr._cumulative import CumulativeAggregations
+
     stack = [expr]
     while stack:
         e = stack.pop()
         if isinstance(e, IO):
             yield e
-        elif isinstance(e, Blockwise):
+        elif isinstance(e, (Blockwise, CumulativeAggregations)):
+            # TODO: Capture this in inheritance logic
```
milesgranger, line 2052: If this is pertinent to dask CI, do we want to create and issue for it?

milesgranger (approved):
Just small comment, looks great otherwise.

dask_expr/_expr.py
phofl: I have a PR already but It needs a bit more thought, not super urgent for now

PR #560, opened 2023-12-15
Rewrite GroupBy.median

phofl:
ref #525

having median in a different structure made everything else difficult, this is slightly hacky, but reduces the mental load significantly


milesgranger (approved):
A little confused with `split_out` defaulting to True but treated as an int otherwise in median, but I think it's okay overall.

phofl:
median only reduces after the shuffle, so true is a better default than anything else. It's similar to what drop_duplicates does

PR #576, opened 2023-12-15
Refactor structure in GroupBy to dispatch to SeriesGroupy earlier

phofl:
ref #525

the initial method on the DataFrame groupby goes back to missing series groupby support, this cleans that up


phofl:
merging, others are building on top of this one

Issue #584, opened 2023-12-18
GroupBy.idxmin and idxmax are very broken


phofl:
The implementation is broken as soon as the max or min value is not in the first partition

This was broken for years, so only adding for API compatibility now but we should fix this

```
def test_df_groupby_idxmax():
    pdf = pd.DataFrame(
        {"idx": list(range(4)), "group": [1, 1, 2, 2, 1], "value": [10, 20, 20, 10, 40]}
    ).set_index("idx")

    ddf = dd.from_pandas(pdf, npartitions=3)

    expected = pd.DataFrame({"group": [1, 2], "value": [4, 2]}).set_index("group")

    result_pd = pdf.groupby("group").idxmax()
    result_dd = ddf.groupby("group").idxmax()

    assert_eq(result_pd, result_dd)
    assert_eq(expected, result_dd)
```

This is a very simple reproducer for the underlying issue

PR #585, opened 2023-12-18
Implement GroupBy.idxmin and GroupBy.idxmax

phofl:
ref #525

Adding for api compatibility, although the implementation is broken

see #584

### DIFF
--- a/dask_expr/tests/test_groupby.py
+++ b/dask_expr/tests/test_groupby.py
@@ -34,7 +34,8 @@ def test_groupby_unsupported_by(pdf, df):
 
 @pytest.mark.parametrize("split_every", [None, 5])
 @pytest.mark.parametrize(
-    "api", ["sum", "mean", "min", "max", "prod", "first", "last", "var", "std"]
+    "api",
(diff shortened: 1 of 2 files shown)


dask_expr/tests/test_groupby.py
```diff
@@ -34,7 +34,8 @@ def test_groupby_unsupported_by(pdf, df):
 
 @pytest.mark.parametrize("split_every", [None, 5])
 @pytest.mark.parametrize(
-    "api", ["sum", "mean", "min", "max", "prod", "first", "last", "var", "std"]
+    "api",
+    ["sum", "mean", "min", "max", "prod", "first", "last", "var", "std", "idxmin"],
```
milesgranger, line 38:
```suggestion
    ["sum", "mean", "min", "max", "prod", "first", "last", "var", "std", "idxmin", "idxmax"],
```
?

phofl: no that raises, #584

milesgranger: ah, right. :+1:

PR #588, opened 2023-12-18
GroupBy.rolling - Integer window type

### LINKED ISSUE #493: GroupBy rolling should raise with integer window
dask/dask raises here, it's not clear to me why. We should check this and potentially make it more flexible

milesgranger:
Will close #493 

Two approaches here 

First (904646bd77b035bb37d30349005bc950268edca8) alters `GroupBy.rolling` to simply check window is an integer and raise the exact error dask/dask does currently.

Second (bdb61d9db6e0ddb1eeaa15c3e06a9597b346203a) allows it and only adds tests for the expected failures. Fine to use integer window if only numeric data in the dataframe, otherwise it raises the expected "Cannot aggregate non-numeric type".

If we want the second, I suppose we could follow up with a `numeric_only` as a convenience?

### DIFF
--- a/dask_expr/_groupby.py
+++ b/dask_expr/_groupby.py
@@ -1,6 +1,7 @@
 import functools
 import math
 import warnings
+from numbers import Integral
 
 import numpy as np
 from dask import is_dask_collection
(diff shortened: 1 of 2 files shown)


phofl:
I think we should be able to allow an integer window even if we have a date time index?

milesgranger:
> I think we should be able to allow an integer window even if we have a date time index?

Yes, that's right.. when I said non-numeric in dataframe, I should have clarified it's okay to have datetime in the index. The first test case checks this in the second commit https://github.com/dask-contrib/dask-expr/pull/588/commits/bdb61d9db6e0ddb1eeaa15c3e06a9597b346203a.

phofl (approved):

phofl:
thx

PR #594, opened 2023-12-18
Ensure that accessors are not available on invalid dtypes

phofl:
ref #525

### DIFF
--- a/dask_expr/tests/test_string_accessor.py
+++ b/dask_expr/tests/test_string_accessor.py
@@ -98,3 +98,13 @@ def test_string_accessor(ser, dser, func, kwargs):
         )
     else:
         assert_eq(getattr(dser.index.str, func)(**kwargs), pdf_result)
+
+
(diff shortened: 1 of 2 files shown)


dask_expr/tests/test_string_accessor.py
```diff
@@ -98,3 +98,13 @@ def test_string_accessor(ser, dser, func, kwargs):
...
+
+
+def test_str_accessor_not_available():
+    pdf = lib.DataFrame({"a": [1, 2, 3]})
+    df = from_pandas(pdf, npartitions=2)
+    # Not available on invalid dtypes
+    with pytest.raises(AttributeError, match=".str accessor"):
+        df.a.str
+
+    assert "str" not in dir(df.a)
```
milesgranger, line 110: Should we also have one for `cat`?

phofl: We have them in dask/dask, so didn't bother here

PR #598, opened 2023-12-18
Fixup some groupby warning related stuff

phofl:
ref #525

### DIFF
--- a/dask_expr/_groupby.py
+++ b/dask_expr/_groupby.py
@@ -1421,12 +1421,13 @@ class SeriesGroupBy(GroupBy):
     ):
         # Raise pandas errors if applicable
         if isinstance(obj, Series):
-            if (
-                isinstance(by, FrameBase)
-                or isinstance(by, (list, tuple))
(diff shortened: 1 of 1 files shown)


milesgranger (approved):

dask_expr/_groupby.py
```diff
@@ -1034,7 +1034,7 @@ def __init__(
 
         self.obj = obj[projection] if projection is not None else obj
         self.sort = sort
-        self.observed = observed
+        self.observed = observed if observed is not None else False
```
rjzamora, line 1037: Ah - Looks like cudf uses an `observed=True` default, and doesn't support `observed=False`, so this breaks things a bit with cudf. Do you remember the dask/dask failure this was meant to fix? (I don't think I see failures in dask-expr when I change this back)

phofl: You'll only see this with a pandas version that has the observed deprecation generally speaking and I indeed think that there is no test in dask-expr, only over in dask/dask.

IIRC we were showing the deprecation warning multiple times without this change

PR #603, opened 2023-12-19
Fix merge dtype warning and divisons after set_index

phofl:
we need the divisions after the shuffle but we don't wan to lock them in to early, so this is a middle ground solution

ref #525


milesgranger (approved):
LGTM, thanks!

PR #608, opened 2023-12-19
Unpack collections in expression constructor

phofl:
cc @mrocklin thoughts about this? That's what we discussed today

Side note: We have to figure out imports at some point, runtime imports might become problematic otherwise

ref #525

### DIFF
--- a/dask_expr/_accessor.py
+++ b/dask_expr/_accessor.py
@@ -57,19 +57,17 @@ class Accessor:
 
         if isinstance(self._series, Index):
             return new_collection(
-                FunctionMapIndex(
-                    self._series.expr, self._accessor_name, attr, args, kwargs
-                )
+                FunctionMapIndex(self._series, self._accessor_name, attr, args, kwargs)
             )
 
         return new_collection(
-            FunctionMap(self._series.expr, self._accessor_name, attr, args, kwargs)
+            FunctionMap(self._series, self._accessor_name, attr, args, kwargs)
         )
 
     def _property_map(self, attr, *args, **kwargs):
         from dask_expr._collection import new_collection
 
-        return new_collection(PropertyMap(self._series.expr, self._accessor_name, attr))
+        return new_collection(PropertyMap(self._series, self._accessor_name, attr))
 
 
 class PropertyMap(Elemwise):
--- a/dask_expr/_core.py
+++ b/dask_expr/_core.py
@@ -30,8 +30,6 @@ class Expr:
     _defaults = {}
 
     def __init__(self, *args, **kwargs):
-        from dask_expr._collection import FrameBase, Scalar
-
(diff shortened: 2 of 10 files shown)


dask_expr/_core.py
```diff
@@ -20,13 +20,18 @@ class Expr:
...
         operands = list(args)
         for parameter in type(self)._parameters[len(operands) :]:
             try:
                 operands.append(kwargs.pop(parameter))
             except KeyError:
                 operands.append(type(self)._defaults[parameter])
         assert not kwargs, kwargs
+        operands = [
+            o.expr if isinstance(o, (FrameBase, Scalar)) else o for o in operands
+        ]
```
mrocklin, line 34: Maybe instead we check for some `__dask_expr__` attribute?  This seems like a case where a protocol might be better than type checking.

dask_expr/_accessor.py
```diff
@@ -57,19 +57,17 @@ def _function_map(self, attr, *args, **kwargs):
 
         if isinstance(self._series, Index):
             return new_collection(
-                FunctionMapIndex(
-                    self._series.expr, self._accessor_name, attr, args, kwargs
-                )
```
mrocklin, line 62: It's nice seeing all of this `.expr` complexity going away

dask_expr/_core.py
phofl: I am Not a big fan of these type of protocols, that said the scheduler integration will make this specific issue go away so that’s only a temporary solution anyway

dask_expr/_accessor.py
phofl: Yep 100% agree

dask_expr/_core.py
mrocklin: I think that protocols are more aligned with what you actually want here.  You want to exprify anything that knows how to turn itself into an expr.  You don't actually care about the FrameBase or Scalar classes.  

If you're against protocols in principle in some way then we should probably talk about that.  A lot of Dask depends on protocols.  Going against the grain there is likely to cause friction down the road.

phofl: I am not against protocols in principle but I am cautious about them. They have caused me pain in the past through thinks like what we have with __dask_graph__ where a too expensive computation might get triggered if you are not careful or through catching things that you don’t actually want to catch.

Checking for an expr attribute here should be fine though

mrocklin: > They have caused me pain in the past through thinks like what we have with dask_graph where a too expensive computation might get triggered if you are not careful or through catching things that you don’t actually want to catch

Yeah, in that case I think it has a lot more to do with how we implemented things than with the protocol itself.  Certainly I'm not in favor of doing things badly :)

PR #611, opened 2023-12-20
Add index setter

phofl:
ref #525

cc @hendrikmakait

### DIFF
--- a/dask_expr/_collection.py
+++ b/dask_expr/_collection.py
@@ -328,6 +328,15 @@ class FrameBase(DaskMethodsMixin):
     def index(self):
         return new_collection(self.expr.index)
 
+    @index.setter
+    def index(self, value):
+        assert isinstance(value, Index), type(value)
+        assert expr.are_co_aligned(
+            self.expr, value.expr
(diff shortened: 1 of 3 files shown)


dask_expr/_collection.py
```diff
@@ -328,6 +328,14 @@ def visualize(self, tasks: bool = False, **kwargs):
     def index(self):
         return new_collection(self.expr.index)
 
+    @index.setter
+    def index(self, value):
+        assert expr.are_co_aligned(
```
hendrikmakait, line 333: Out of scope: Is the docstring of `are_co_aligned` inverted? From my superficial understanding, it should rather be `Do inputs come from the same parents, ignoring blockwise?` to match the `True`/`False` results.

hendrikmakait (approved):
Thanks, @phofl! One out-of-scope comment for my understanding.

dask_expr/_collection.py
phofl: Yep :) Tells you how much I read my own docstrings (at least the function name is helpful :)). I'll adjust in my next pr

PR #612, opened 2023-12-20
Keep npartitions=1 for merge if both sides only have 1

phofl:
ref #525

splitting the partitions seems odd here


hendrikmakait (approved):
Thanks, @phofl!

PR #614, opened 2023-12-20
Implement from_dask_array

phofl:
This is a cheap implementation, but it should do for now

ref #525

### DIFF
--- a/dask_expr/io/tests/test_io.py
+++ b/dask_expr/io/tests/test_io.py
@@ -503,5 +503,5 @@ def test_from_dask_array():
 
     arr = da.ones((20, 4), chunks=(2, 2))
     df = from_dask_array(arr, columns=["a", "b", "c", "d"])
-    pdf = lib.DataFrame(arr.compute(), columns=list("abcd"))
+    pdf = lib.DataFrame(arr.compute(), columns=["a", "b", "c", "d"])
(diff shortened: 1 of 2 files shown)


dask_expr/io/tests/test_io.py
```diff
@@ -495,3 +496,12 @@ def test_from_pandas_divisions_duplicated():
...
     assert_eq(df.partitions[1], pdf.loc[5:6])
     assert_eq(df.partitions[2], pdf.loc[8:])
+
+
+def test_from_dask_array():
+    import dask.array as da
+
+    arr = da.ones((20, 4), chunks=(2, 2))
+    df = from_dask_array(arr, columns=["a", "b", "c", "d"])
+    pdf = lib.DataFrame(arr.compute(), columns=list("abcd"))
```
hendrikmakait, line 506: nit: Let's be as consistent as possible
```suggestion
    pdf = lib.DataFrame(arr.compute(), columns=["a", "b", "c", "d"])
```

hendrikmakait (approved):
Thanks, @phofl!

dask_expr/io/tests/test_io.py
phofl: got tired of typing :)

PR #616, opened 2023-12-21
Implement from_dict

phofl:
ref #525

### DIFF
--- a/dask_expr/_collection.py
+++ b/dask_expr/_collection.py
@@ -1810,6 +1810,58 @@ def from_graph(*args, **kwargs):
     return new_collection(FromGraph(*args, **kwargs))
 
 
+def from_dict(
+    data,
+    npartitions,
+    orient="columns",
+    dtype=None,
+    columns=None,
+    constructor=pd.DataFrame,
+):
+    """
+    Construct a Dask DataFrame from a Python Dictionary
+
+    Parameters
+    ----------
+    data : dict
+        Of the form {field : array-like} or {field : dict}.
+    npartitions : int
+        The number of partitions of the index to create. Note that depending on
+        the size and index of the dataframe, the output may have fewer
+        partitions than requested.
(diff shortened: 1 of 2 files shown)


dask_expr/_collection.py
```diff
@@ -1810,6 +1816,58 @@ def from_graph(*args, **kwargs):
...
+    constructor: class, default pd.DataFrame
+        Class with which ``from_dict`` should be called with.
+
+    Examples
+    --------
+    >>> import dask.dataframe as dd
+    >>> ddf = dd.from_dict({"num1": [1, 2, 3, 4], "num2": [7, 8, 9, 10]}, npartitions=2)
+    """
+
+    collection_types = {type(v) for v in data.values() if is_dask_collection(v)}
```
hendrikmakait, line 1858: nit: The error message might be more helpful if we stored these as a dict to highlight where the collections were provided.

dask_expr/io/tests/test_io.py
```diff
@@ -505,3 +507,11 @@ def test_from_dask_array():
     df = from_dask_array(arr, columns=["a", "b", "c", "d"])
     pdf = lib.DataFrame(arr.compute(), columns=["a", "b", "c", "d"])
     assert_eq(df, pdf)
+
+
+def test_from_dict():
```
hendrikmakait, line 512: Should we test the error case here?

hendrikmakait (approved):
The changes look good to me. Two non-blocking comments around error handling for dask collections.

dask_expr/_collection.py
phofl: I agree generally speaking, but this can produce very long error messages that screw your terminal (we had cases like this in pandas)

dask_expr/io/tests/test_io.py
phofl: They are covered in dask/dask, so I wasn't too careful here.

PR #617, opened 2023-12-21
Add from_delayed

phofl:
The expression representation is a bit weird for now, but I'd like to have from_delayed for testing purposes before we implement it properly

ref #525

### DIFF
--- a/dask_expr/io/_delayed.py
+++ b/dask_expr/io/_delayed.py
@@ -2,7 +2,7 @@ from __future__ import annotations
 
 import functools
 from collections.abc import Iterable
-from typing import TYPE_CHECKING, Literal
+from typing import TYPE_CHECKING
 
 from dask.dataframe.dispatch import make_meta
 from dask.dataframe.utils import check_meta
@@ -80,7 +80,7 @@ def identity(x):
 def from_delayed(
     dfs: Delayed | distributed.Future | Iterable[Delayed | distributed.Future],
     meta=None,
-    divisions: tuple | Literal["sorted"] | None = None,
+    divisions: tuple | None = None,
     verify_meta: bool = True,
 ):
     """Create Dask DataFrame from many Dask Delayed objects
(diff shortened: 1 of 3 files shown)


dask_expr/io/_delayed.py
```diff
@@ -0,0 +1,125 @@
...
+        if self.verify_meta:
+            return (
+                functools.partial(check_meta, meta=self._meta, funcname="from_delayed"),
+                key,
+            )
+        else:
+            return identifiy, key
+
+
+def identifiy(x):
```
hendrikmakait, line 76: Did you mean
```suggestion
def identity(x):
```

dask_expr/io/_delayed.py
```diff
@@ -0,0 +1,125 @@
...
+        If True check that the partitions have consistent metadata, defaults to True.
+    """
+    if isinstance(dfs, Delayed) or hasattr(dfs, "key"):
+        dfs = [dfs]
+
+    if len(dfs) == 0:
+        raise TypeError("Must supply at least one delayed object")
+
+    dfs = [
+        delayed(df) if not isinstance(df, Delayed) and hasattr(df, "key") else df
```
hendrikmakait, line 115: I'm not entirely sure what you want to test here, but it feels like you may want to check whether this is an actual or duck-typed delayed which would mirror the check above.
```suggestion
        delayed(df) if not (isinstance(df, Delayed) or hasattr(df, "key")) else df
```

dask_expr/io/_delayed.py
phofl: Oh yes, thx

dask_expr/io/_delayed.py
```diff
@@ -0,0 +1,125 @@
...
+
+    def _filtered_task(self, index: int):
+        key = self.dfs[index]._name
+        if self.verify_meta:
+            return (
+                functools.partial(check_meta, meta=self._meta, funcname="from_delayed"),
+                key,
+            )
+        else:
+            return identifiy, key
```
hendrikmakait, line 73: Sorry, you gotta change this as well.

```suggestion
            return identity, key
```

hendrikmakait (approved):
Thanks, @phofl!

PR #624, opened 2023-12-22
Don't raise in set_index for length one list

phofl:
ref #525

### DIFF
--- a/dask_expr/_collection.py
+++ b/dask_expr/_collection.py
@@ -1228,6 +1228,8 @@ class DataFrame(FrameBase):
         upsample: float = 1.0,
         partition_size: float = 128e6,
     ):
+        if isinstance(other, list) and len(other) == 1:
+            other = other[0]
         if isinstance(other, list):
             if any([isinstance(c, FrameBase) for c in other]):
                 raise TypeError("List[FrameBase] not supported by set_index")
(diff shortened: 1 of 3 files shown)


hendrikmakait (approved):
Thanks, @phofl!

phofl:
Yeah my fault, we are now supporting something that didn't work before :)

phofl:
I pushed another fix here because I forgot to create a new branch and both touch the same place anyway

dask_expr/_collection.py
```diff
@@ -1228,6 +1228,8 @@ def set_index(
         upsample: float = 1.0,
         partition_size: float = 128e6,
     ):
+        if isinstance(other, list) and len(other) == 1:
+            other = other[0]
```
milesgranger, line 1232: This will miss the `FrameBase` check below, is that okay? See dask/dask checks for Sequence, so guessing its some similar error checking missing?
https://github.com/dask/dask/blob/f2d448e7a8561c2488806cb398470459984f11b0/dask/dataframe/core.py#L5406

milesgranger (approved):
A clarifying question/comment. If it's not applicable, feel free to merge. :100:

dask_expr/_collection.py
phofl: Yep that's fine, length one lists can be supported

Issue #625, opened 2023-12-22
Map partitions on a series that yields a scalar per partition cannot be computed


fjetter:
```python
from dask_expr.datasets import timeseries
from dask.utils import M

df = timeseries()
df.x.map_partitions(M.min).compute()
```

```python-traceback
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
Cell In[4], line 1
----> 1 df.x.map_partitions(M.min).compute()

File ~/workspace/dask-expr/dask_expr/_collection.py:265, in FrameBase.compute(self, fuse, **kwargs)
    263     out = out.repartition(npartitions=1)
    264 out = out.optimize(fuse=fuse)
--> 265 return DaskMethodsMixin.compute(out, **kwargs)

File ~/workspace/dask/dask/base.py:342, in DaskMethodsMixin.compute(self, **kwargs)
    318 def compute(self, **kwargs):
    319     """Compute this dask collection
    320
    321     This turns a lazy Dask collection into its in-memory equivalent.
   (...)
    340     dask.compute
    341     """
--> 342     (result,) = compute(self, traverse=False, **kwargs)
    343     return result

File ~/workspace/dask/dask/base.py:628, in compute(traverse, optimize_graph, scheduler, get, *args, **kwargs)
    625     postcomputes.append(x.__dask_postcompute__())
    627 with shorten_traceback():
--> 628     results = schedule(dsk, keys, **kwargs)
    630 return repack([f(r, *a) for r, (f, a) in zip(results, postcomputes)])

File ~/workspace/dask/dask/dataframe/dispatch.py:67, in concat(dfs, axis, join, uniform, filter_warning, ignore_index, **kwargs)
     65     return dfs[0]
     66 else:
---> 67     func = concat_dispatch.dispatch(type(dfs[0]))
     68     return func(
     69         dfs,
     70         axis=axis,
   (...)
     75         **kwargs,
     76     )

File ~/workspace/dask/dask/utils.py:635, in Dispatch.dispatch(self, cls)
    633         self._lazy.pop(toplevel, None)
    634         return self.dispatch(cls)  # recurse
--> 635 raise TypeError(f"No dispatch for {cls}")

TypeError: No dispatch for <class 'numpy.float64'>
```

So far, I believe the reason for this is that the `min` is returning a scalar float (the type that is tried to be dispatched) instead of a `Series` with one element. This then fails in the concat operation when computing the series.

phofl:
[...]

PR #626, opened 2023-12-22
Use _concat for repartition

fjetter:
Closes https://github.com/dask-contrib/dask-expr/issues/625

I don't really like `_concat`. Starting with the name of it and it's relatively magical rv. It does fix the problem.

FWIW I encountered this in https://github.com/dask-contrib/dask-expr/pull/294 during division calculation since the optimization is now different and includes a postcompute, i.e. a repartition

### DIFF
--- a/dask_expr/_repartition.py
+++ b/dask_expr/_repartition.py
@@ -7,7 +7,7 @@ import numpy as np
 import pandas as pd
 from dask.base import tokenize
 from dask.dataframe import methods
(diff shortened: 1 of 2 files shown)


fjetter:
I suspect that we'd have to use this in other places as well if this is what we want do to

phofl:
See my comment in the issue, this solves this particular case, but having integers in partitions is a general issue that is bad

fjetter:
Closing in favor of https://github.com/dask/dask/pull/10739

PR #627, opened 2024-01-03
Fix column setter for midx columns

phofl:
ref #525


milesgranger (approved):
Nice. Thank you! 
(Assuming tests happy after resolving conflicts.)

Issue #501, opened 2023-12-08
Add split_every to reductions like sum


phofl:
Keyword is missing, but it's already supported in the base implementation

PR #631, opened 2024-01-03
Implement Series.describe

phofl:
ref #525

### DIFF
--- a/dask_expr/_collection.py
+++ b/dask_expr/_collection.py
@@ -35,7 +35,11 @@ from dask.utils import (
     typename,
 )
 from fsspec.utils import stringify_path
-from pandas.core.dtypes.common import is_datetime64_any_dtype
+from pandas.core.dtypes.common import (
+    is_datetime64_any_dtype,
+    is_numeric_dtype,
+    is_timedelta64_dtype,
+)
 from tlz import first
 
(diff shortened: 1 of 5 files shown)


dask_expr/_collection.py
```diff
@@ -695,14 +700,20 @@ def prod(self, skipna=True, numeric_only=False, min_count=0, split_every=False):
...
+        return new_collection(
+            self.expr.std(axis, skipna, ddof, numeric_only, split_every=split_every)
+        )
 
-    def mean(self, skipna=True, numeric_only=False, min_count=0):
-        return new_collection(self.expr.mean(skipna, numeric_only))
+    def mean(self, skipna=True, numeric_only=False, min_count=0, split_every=False):
+        return new_collection(
+            self.expr.mean(skipna, numeric_only, split_every=split_every)
+        )
```
milesgranger, line 716: Does this close https://github.com/dask-contrib/dask-expr/issues/501 then?

milesgranger (approved):
Interesting read. Looks goodie, with one clarifying comment. Thanks!

dask_expr/_collection.py
phofl: Probably, will close for now

PR #635, opened 2024-01-03
Implement categorize

hendrikmakait:


dask_expr/_collection.py
```diff
@@ -1363,6 +1363,72 @@ def iloc(self):
...
+            if is_categorical_dtype(meta.index):
+                index = not has_known_categories(meta.index)
+            elif index is None:
+                index = str(meta.index.dtype) in ("object", "string")
+
+        # Nothing to do
+        if not len(columns) and index is False:
+            return df
+
+        if split_every is None:
```
phofl, line 1410: ApplyConcatApply handles the split_every semantics, you can remove this here

dask_expr/_collection.py
```diff
@@ -1363,6 +1363,72 @@ def iloc(self):
...
+            GetCategories(
+                self.expr, columns=columns, index=index, split_every=split_every
+            )
+        ).compute()
+
+        # Some operations like get_dummies() rely on the order of categories
+        categories = {k: v.sort_values() for k, v in categories.items()}
+
+        # Categorize each partition
+        return new_collection(Categorize(self.expr, categories, index))
```
phofl, line 1430: the .expr is no longer necessary, we now handle this globally in the constructor

dask_expr/_categorical.py
```diff
@@ -127,3 +134,31 @@ class AsUnknown(Elemwise):
...
+
+    @functools.cached_property
+    def _meta(self):
+        meta = _categorize_block(
+            self.frame._meta, self.operand("categories"), self.operand("index")
+        )
+        return meta
+
+
+class GetCategories(ApplyConcatApply):
```
phofl, line 152: This should implement simpliy_down, so that we inject a Projection before the ``GetCategories`` so that we can drop unnecessary columns

dask_expr/_collection.py
```diff
@@ -1363,6 +1363,72 @@ def iloc(self):
...
+            split_every = 16
+        elif split_every is False:
+            split_every = df.npartitions
+        elif not isinstance(split_every, Integral) or split_every < 2:
+            raise ValueError("split_every must be an integer >= 2")
+
+        from dask_expr._collection import new_collection
+
+        # Eagerly compute the categories
+        categories, index = new_collection(
```
phofl, line 1420: We should probably warn that this triggers a compute?

dask_expr/_categorical.py
phofl: You can also do this before calling ``GetCategories``, that might make things a little bit easier

dask_expr/_collection.py
```diff
@@ -1363,6 +1363,72 @@ def iloc(self):
 
         return ILocIndexer(self)
 
+    def categorize(self, columns=None, index=None, split_every=None, **kwargs):
+        """Convert columns of the DataFrame to category dtype.
```
hendrikmakait, line 1367: I've noticed we seldom add docstrings. What's the current approach here?

phofl: We haven't generated docs yet, so I didn't bother with docstrings. We have to figure docs out soonish though, so feel free to add them

dask_expr/_collection.py
hendrikmakait: Added a warning to the docstring.

dask_expr/_categorical.py
hendrikmakait: Done.

dask_expr/_categorical.py
```diff
@@ -127,3 +134,39 @@ class AsUnknown(Elemwise):
...
+    def chunk_kwargs(self):
+        return {"columns": self.operand("columns"), "index": self.operand("index")}
+
+    @functools.cached_property
+    def _meta(self):
+        return ({}, pd.Series())
+
+    def _simplify_down(self):
+        return GetCategories(
+            Projection(self.frame, self.operand("columns")),
```
phofl, line 168: This should protect against loops, e.g. if 

```
set(self.frame.columns) ==  set(self.operand("columns"))
```

then don't do anything

hendrikmakait: Ah, right! 🤦

phofl (approved):
thx, we need a test that tests if the optimization works, but not blocking can be a follow up

phofl:
thx

PR #645, opened 2024-01-05
Add chunksize to from_pandas

phofl:
ref #525

### DIFF
--- a/dask_expr/io/tests/test_io.py
+++ b/dask_expr/io/tests/test_io.py
@@ -541,3 +541,10 @@ def test_from_dict():
     expected = lib.DataFrame(data)
     assert_eq(result, expected)
     assert_eq(DataFrame.from_dict(data), expected)
+
+
+def test_from_pandas_chunksize():
+    pdf = lib.DataFrame(np.random.randn(10, 5), columns=list("abcde"))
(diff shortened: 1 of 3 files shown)


dask_expr/io/tests/test_io.py
```diff
@@ -541,3 +541,10 @@ def test_from_dict():
     expected = lib.DataFrame(data)
     assert_eq(result, expected)
     assert_eq(DataFrame.from_dict(data), expected)
+
+
+def test_from_pandas_chunksize():
+    pdf = lib.DataFrame(np.random.randn(10, 5), columns=list("abcde"))
+    df = from_pandas(pdf, chunksize=4)
+    assert df.npartitions == 3
+    assert_eq(df, pdf)
```
milesgranger, line 550: Did we want to also check the invariants for chunksize and npartitions being None creating npartitions==1 and both not None raising a TypeError? Feel free to ignore if you think that's not worth it.

milesgranger (approved):

dask_expr/io/tests/test_io.py
phofl: Added, the tests make sense and don't think we have them in dask/dask

milesgranger:
Thanks @phofl!

PR #646, opened 2024-01-05
Fix value_counts with normalize and split_out

phofl:
ref #525

dask/dask currently calculates the length eagerly, this defers the length computation properly, but is kind of ugly, not sure


phofl:
green now

milesgranger (approved):
Thank you!

PR #653, opened 2024-01-05
Align defaults for `split_out` for all `aggregate` methods

hendrikmakait:
Align defaults for `split_out` for all `aggregate` methods and makes sure it matches with `dask/dask`.


phofl:
thx

PR #654, opened 2024-01-05
Fix numeric edge case in repartition

phofl:
ref #525

We are currently losing data, injecting another partition changes the number of partitions in lower which opens pandoras box in other places, so lets just make the last partition a little bit larger


milesgranger (approved):
Seems good for me. :+1:

PR #658, opened 2024-01-05
is_monotonic_increasing/decreasing

crusaderky:
Rewrite unit test and logic.


phofl:
Keep it lightweight here as much as possible.

I am porting tests over if I am fixing specific bugs that are covered by them, but not everything else

dask_expr/tests/test_reductions.py
```diff
@@ -41,30 +41,29 @@ def test_median(pdf, df):
...
+    if reverse:
+        series = series[::-1]
+    pds = lib.Series(series, index=series)
+    ds = from_pandas(pds, 2, sort=False)
+    if cls == "Index":
+        pds = pds.index
+        ds = ds.index
+
+    assert ds.is_monotonic_increasing.compute() == pds.is_monotonic_increasing
+    assert ds.is_monotonic_decreasing.compute() == pds.is_monotonic_decreasing
```
phofl, line 66: Is the type correct now? e.g. I think we got numpy.bool previously instead of a regular python bool

crusaderky: no, it's still mismatched

phofl: Can we reflect this in the test? A comment maybe?

dask_expr/tests/test_reductions.py
```diff
@@ -41,30 +41,29 @@ def test_median(pdf, df):
...
+def test_monotonic(series, reverse, cls):
+    if reverse:
+        series = series[::-1]
+    pds = lib.Series(series, index=series)
+    ds = from_pandas(pds, 2, sort=False)
+    if cls == "Index":
+        pds = pds.index
+        ds = ds.index
+
+    assert ds.is_monotonic_increasing.compute() == pds.is_monotonic_increasing
```
crusaderky, line 65:
```suggestion
    # assert_eq fails due to numpy.bool vs. plain bool mismatch
    # See https://github.com/dask/dask/pull/10671
    assert ds.is_monotonic_increasing.compute() == pds.is_monotonic_increasing
```

dask_expr/tests/test_reductions.py
crusaderky: Done

phofl (approved):

phofl:
thx

PR #659, opened 2024-01-05
Adjust shuffle method for drop_duplicates and smaller changes

phofl:
ref #525

### DIFF
--- a/dask_expr/_reductions.py
+++ b/dask_expr/_reductions.py
@@ -131,12 +131,14 @@ class ShuffleReduce(Expr):
         "split_out",
         "sort",
         "shuffle_by_index",
+        "shuffle_method",
     ]
     _defaults = {
         "split_every": 8,
         "split_out": True,
         "sort": None,
         "shuffle_by_index": None,
+        "shuffle_method": None,
     }
 
     @property
@@ -206,6 +208,7 @@ class ShuffleReduce(Expr):
                 shuffle_npartitions,
                 ignore_index=ignore_index,
                 index_shuffle=not split_by_index and self.shuffle_by_index,
+                backend=self.shuffle_method,
             )
 
         # Unmap column names if necessary
@@ -383,6 +386,8 @@ class ApplyConcatApply(Expr):
             split_out = self.operand("split_out")
             if isinstance(split_out, Callable):
                 split_out = split_out(self.frame.npartitions)
+            if split_out is None:
+                raise ValueError("split_out can't be None")
             return split_out
         else:
             return 1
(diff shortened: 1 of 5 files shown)


dask_expr/tests/test_collection.py
```diff
@@ -1921,7 +1941,7 @@ def test_quantile(df):
         df.x.quantile(q=[]).compute()
 
     ser = from_pandas(lib.Series(["a", "b", "c"]), npartitions=2)
-    with pytest.raises(TypeError, match="on non-numeric"):
+    with pytest.raises(ValueError, match="object series"):
```
milesgranger, line 1944: Seems like we ought to check the other methods `_raise_if_object_series` is applied to. Which itself sorta seems like maybe it could be a decorator to make it more visible for which methods it's applied to? (no strong opinions on this though)

milesgranger (approved):
One comment/suggestion, take it or leave it. Otherwise seems good. :+1:

dask_expr/_reductions.py
```diff
@@ -472,12 +477,13 @@ def _lower(self):
...
             shuffle_by_index=getattr(self, "shuffle_by_index", None),
+            shuffle_method=getattr(self, "shuffle_method", None),
         )
 
 
 class Unique(ApplyConcatApply):
-    _parameters = ["frame", "split_every", "split_out"]
-    _defaults = {"split_every": None, "split_out": True}
+    _parameters = ["frame", "split_every", "split_out", "shuffle_method"]
+    _defaults = {"split_every": None, "split_out": True, "shuffle_method": "tasks"}
```
hendrikmakait, line 486: Why did you choose `tasks` as default? From what I see, other methods default to `None` and decide at a later point (e.g., during `_lower`).

dask_expr/_util.py
```diff
@@ -184,6 +186,28 @@ def _maybe_from_pandas(dfs):
...
 
 
+def _get_shuffle_preferring_order(shuffle):
+    if shuffle is not None:
+        return shuffle
+
+    if get_default_shuffle_method() == "disk":
+        return "tasks"
+
+    return get_default_shuffle_method()
```
hendrikmakait, line 196: Why are we diverging from the "usual" defaults here?

nit:
```suggestion
    shuffle = get_default_shuffle_method()
    if shuffle == "disk":
        return "tasks"

    return shuffle
```

dask_expr/_reductions.py
phofl: Yeah but I want a different default here, tasks preserves order while disk does not, which is important for drop_duplicates and unique

dask_expr/tests/test_collection.py
phofl: We have tests in dask/dask, that's why I didn't bother with additional tests.

Yeah it needs a review probably that we do it for all functions, but for now we are consistent with dask/dask, so that's good enough for me for now

dask_expr/_reductions.py
hendrikmakait: We should probably:
* expand the warnings we raise for `p2p` regarding order stability to also trigger for `disk`
* add comments explaining why we diverge from the standard defaults (otherwise, I can see myself "fixing" this in a few weeks).

phofl: Fixing would result in a bad experience since tests will complain loudly :)

Agree with the warnings, will do this as a follow up

PR #661, opened 2024-01-05
Fix sort_values divisions and set_index append cases

phofl:
ref #525

### DIFF
--- a/dask_expr/_collection.py
+++ b/dask_expr/_collection.py
@@ -1234,6 +1234,7 @@ class DataFrame(FrameBase):
         shuffle_method=None,
         upsample: float = 1.0,
         partition_size: float = 128e6,
+        append: bool = False,
         **options,
     ):
         if isinstance(other, list) and len(other) == 1:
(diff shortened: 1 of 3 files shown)


dask_expr/_collection.py
```diff
@@ -1242,9 +1243,17 @@ def set_index(
             if any([isinstance(c, FrameBase) for c in other]):
                 raise TypeError("List[FrameBase] not supported by set_index")
             elif not sorted:
-                raise NotImplementedError("Multi-column set_index requires sorted=True")
+                raise NotImplementedError(
+                    "Dask dataframe does not yet support multi-indexes.\n"
+                    f"You tried to index with this index: {other}\n"
+                    "Indexes must be single columns only."
+                )
```
milesgranger, line 1250: I'm a bit confused how this message relates to `sorted=False` and this new message. Previous message being "multi-column set_index requires stored=True" implies something different. The following error for `isinstance(other, DataFrame)` makes sense in that context.

milesgranger (approved):
Looks good, I think I only need a bit of guidance on my one comment. Thanks!

dask_expr/_collection.py
phofl: The new message is copied from dask/dask to stay consistent, the general idea is the following:

We support multi indexes only if we don't have to shuffle, e.g. if sorted=True, having a list of other would result in a MultiIndex

PR #662, opened 2024-01-05
Avoid using ``map_partitions`` within internal dask-expr logic

rjzamora:
We should avoid using `map_partitions` for `isna`. It blocks column projection and doesn't work for cudf.

### DIFF
--- a/dask_expr/_collection.py
+++ b/dask_expr/_collection.py
@@ -54,10 +54,12 @@ from dask_expr._describe import DescribeNonNumeric, DescribeNumeric
 from dask_expr._expr import (
     BFill,
     Diff,
+    DiffColumns,
     Eval,
     FFill,
     Query,
     Shift,
+    ShiftColumns,
     ToDatetime,
     ToNumeric,
     ToTimedelta,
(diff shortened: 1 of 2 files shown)


mrocklin:
I don't disagree necessarily, but I'm curious about your thoughts on why it's not good to use `map_partitions` internally?  Are you getting some value from the terms like ShiftColumns and DiffColumns?  FWIW different systems do different things.  Some systems like naming everything explicitly.  Others like having only a few general purpose terms.  This is a matter of taste I think.  

It's totally ok for us to have the taste of "make classes for everything" but I don't think it's an obvious choice (nor is it obviously wrong).

phofl (commented):
I would rather not create boilerplate classes for everything that don‘t add any value at the moment

we can still move into this direction later on, but this seems unnecessary at the moment

rjzamora:
>I would rather not create boilerplate classes for everything that don‘t add any value at the moment

Yeah, I agree with this. I now realize that it is fine to use `map_partitions` for `diff`/`shift`. I was originally thinking that we are probably missing out on some optimizations by using `map_partitions`, but I suppose we cannot project columns in this case (my mistake).

Sorry - I will roll back the `diff`/`shift` changes. I should have qualified my statement as: "(map_partitions) is not yet a good way to implement expressions that can benefit from common optimizations (e.g. column projection)."

I still think the `isna` change makes sense.  The `map_partitions` approach doesn't work for cudf anyway.

rjzamora:
Okay - I stripped out everything besides the `isna` fix (which was my only real concern here). Sorry for the other noise, but I appreciate the feedback.

phofl (approved):

phofl:
thx

yeah axis=1 can't be optimised (mostly), we can push length calculations through, but that's not super important now imo and can be done with one general class as well

PR #664, opened 2024-01-05
Add io connectors through rountripping with dask/dask

phofl:
cc @mrocklin 

This is probably the area where we can save time without losing much in the sense of performance and UX for users. We can switch them out one after another after we are happy with the other parts of the API, compromising here makes the most sense imo. Wdyt?

### DIFF
--- a/dask_expr/__init__.py
+++ b/dask_expr/__init__.py
@@ -1,7 +1,13 @@
 from dask_expr import _version, datasets
 from dask_expr._collection import *
 from dask_expr.io._delayed import from_delayed
+from dask_expr.io.bag import to_bag
+from dask_expr.io.csv import to_csv
+from dask_expr.io.hdf import read_hdf, to_hdf
 from dask_expr.io.json import read_json, to_json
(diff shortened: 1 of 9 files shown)


mrocklin:
I don't have strong thoughts.  In principle this isn't great, but it's not worse than status quo and doesn't bind us to any bad decisions.  The best counterargument I can come up with is "if it's this straightforward to cheat then it's probably not much harder to just do the work.  This is probably just a tricky copy paste job.  There aren't any complex optimizations or anything.".  You're on the ground though and so I figure you have a more nuanced understanding of the costs here, so I'm more than happy to defer to your judgement.

phofl:
The transfer shouldn't be hard generally speaking but I suspect that it would still take at least 3-4 days of work until everything works smoothly. There are optimisations available, read_sql comes to mind, but that's probably not a big deal for now.

I don't want to keep this implementation forever, but if we want to get to a point where we are confident that dask-expr works reliable as fast as possible, I would rather prioritise making sure that the non-io stuff works great and do this after we are at that point.

PR #669, opened 2024-01-08
Fix datetimeindex in set_index

phofl:
ref #525

sits on https://github.com/dask/dask/pull/10757

### DIFF
--- a/dask_expr/_accessor.py
+++ b/dask_expr/_accessor.py
@@ -94,16 +94,5 @@ class FunctionMap(Elemwise):
 
 class FunctionMapIndex(FunctionMap):
     def _divisions(self):
-        try:
-            divisions = self.operation(
-                type(self.frame._meta)(self.frame.divisions),
-                self.accessor,
-                self.attr,
-                self.args,
-                self.kwargs,
-            )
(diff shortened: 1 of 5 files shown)


phofl:
this one should be ready now after the dask/dask pr is in

dask_expr/_accessor.py
```diff
@@ -94,4 +94,5 @@ def operation(obj, accessor, attr, args, kwargs):
 
 class FunctionMapIndex(FunctionMap):
     def _divisions(self):
+        # TODO: We can do better here
```
milesgranger, line 97: I've seen these todo's, kinda assuming you have a method for dealing with them or do we want to be strict about making follow up issues?

milesgranger (approved):
Just some more questions, still looks good to me though!

dask_expr/_accessor.py
phofl: grepping when we are happy with the state of the test suite,

this specific occurrence isn't super important, so I avoided creating an issue, it's a tiny annoyance for now

phofl:
I did not know that we were caching environments....

PR #672, opened 2024-01-08
Fix nulls that are introduced during index partitioning

phofl:
cc @hendrikmakait 

Same as what we fixed in dask/dask before Christmas

ref #525

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -31,15 +31,15 @@ jobs:
       - uses: actions/checkout@v2
         with:
           fetch-depth: 0  # Needed by codecov.io
+
       - name: Install Environment
-        uses: mamba-org/provision-with-micromamba@main
+        uses: mamba-org/setup-micromamba@v1
         with:
           environment-file: ${{ matrix.environment-file }}
-          environment-name: dask-match
-          extra-specs: |
-            python=${{ matrix.python-version }}
+          extra-specs: python=${{ matrix.python-version }}
+          cache-environment: true
 
-      - name: Install Dask-Match
+      - name: Install dask-expr
         run: python -m pip install -e . --no-deps
 
       - name: Run tests
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -25,7 +25,7 @@ jobs:
       fail-fast: false
       matrix:
         python-version: ["3.9", "3.10", "3.11", "3.12"]
-        environment-file: [ci/environment.yml, ci/environment_released.yml]
+        environment-file: [ci/environment.yml]
 
     steps:
       - uses: actions/checkout@v2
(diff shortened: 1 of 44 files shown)


dask_expr/_shuffle.py
```diff
@@ -244,7 +246,10 @@ def from_abstract_shuffle(cls, expr: Shuffle) -> Expr:
                 # Don't need to assign "_partitions" column
                 # if we are shuffling on a list of columns
                 nset = set(partitioning_index)
-                if nset & set(frame.columns) == nset and not index_shuffle:
+                if (
+                    nset & nset.intersection(set(frame.columns)) == nset
```
hendrikmakait, line 250: This condition feels overly complicated. If I'm not missing anything, this simplifies to `nset.issubset(set(frame.columns))`.

dask_expr/_merge.py
```diff
@@ -284,12 +284,14 @@ def _lower(self):
                             left,
                             shuffle_left_on,
                             npartitions_out=left.npartitions,
+                            backend="tasks",
```
hendrikmakait, line 287: Why does this rely on `tasks`?

phofl: To keep the hashing consistent, not ideal but everything else loses data at the moment

phofl: Cleaned this up now after we removed the bug in dask/dask

dask_expr/_shuffle.py
phofl: Removed altogether similar to dask/dask

phofl:
Turns out this gave me so much trouble because dask/dask had a bug, I am disabling the release ci until the next release because we will get a bunch of failures there. Won't cut a new dask-expr release until dask/dask releases. We will have to bump the min dask version then as well

.github/workflows/test.yaml
```diff
@@ -25,7 +25,7 @@ jobs:
       fail-fast: false
       matrix:
         python-version: ["3.9", "3.10", "3.11", "3.12"]
-        environment-file: [ci/environment.yml, ci/environment_released.yml]
```
hendrikmakait, line 28: Is this intended?

hendrikmakait (approved):
Generally LGTM; one question that needs to be clarified before merging, but there's no need to go through another review. Thanks, @phofl!

.github/workflows/test.yaml
phofl: Yes

We are currently incompatible with released dask dask

hendrikmakait: Makes sense, in an ideal world we should have split that into a separate PR.

phofl: I actually merged this already in another PR before this one, not sure why the diff is still the way it is

PR #674, opened 2024-01-08
Improve partition calculation for recursive join

phofl:
ref #525

### DIFF
--- a/dask_expr/_merge.py
+++ b/dask_expr/_merge.py
@@ -683,15 +683,10 @@ class BlockwiseMerge(Merge, Blockwise):
         if self.left.npartitions == self.right.npartitions:
             return super()._divisions()
         is_unknown = any(d is None for d in super()._divisions())
-        if self.left.npartitions > self.right.npartitions:
-            # Broadcast blockwise merge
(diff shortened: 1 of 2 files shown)


dask_expr/_merge.py
```diff
@@ -679,6 +679,20 @@ class BlockwiseMerge(Merge, Blockwise):
...
+        is_unknown = any(d is None for d in super()._divisions())
+        if self.left.npartitions > self.right.npartitions:
+            # Broadcast blockwise merge
+            if is_unknown:
+                return (None,) * (self.left.npartitions + 1)
+            return self.left.divisions
+        else:
+            if is_unknown:
+                return (None,) * (self.right.npartitions + 1)
+            return self.right.divisions
```
milesgranger, line 694:
```suggestion
        frame = self.left if self.left.npartitions > self.right.npartitions else self.right
        if is_unknown:
            return (None,) * (frame.npartitions + 1)
        return frame.divisions
```
nit, take it or leave it. :)

milesgranger (approved):
Looks good to me, just a small suggested refactor.

PR #686, opened 2024-01-09
Respect culling in set_index if divisions are given

phofl:
ref #525

### DIFF
--- a/dask_expr/_shuffle.py
+++ b/dask_expr/_shuffle.py
@@ -1122,9 +1122,26 @@ class _SetIndexPost(Blockwise):
         df.columns = df.columns.astype(column_dtype)
         return df
 
+    def _get_culled_divisions(self, divisions):
+        if self.frame.npartitions < len(divisions) - 1:
+            part_filter = list(self.frame.find_operations(PartitionsFiltered))
+            if len(part_filter) > 0:
+                return tuple(
+                    [
+                        div
+                        for i, div in enumerate(divisions)
+                        if i in part_filter[0]._partitions
+                    ]
+                    + [divisions[-1]]
+                )
+            else:
+                return self.frame.divisions
+
+        return divisions
+
     def _divisions(self):
         if self.operand("user_divisions") is not None:
-            return self.operand("user_divisions")
+            return self._get_culled_divisions(self.operand("user_divisions"))
         kwargs = self.key_kwargs
         key = (
             kwargs["other"],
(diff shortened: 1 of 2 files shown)


dask_expr/_shuffle.py
```diff
@@ -1122,9 +1122,26 @@ def operation(df, index_name, drop, set_name, column_dtype):
...
+        if self.frame.npartitions < len(divisions) - 1:
+            part_filter = list(self.frame.find_operations(PartitionsFiltered))
+            if len(part_filter) > 0:
+                return tuple(
+                    [
+                        div
+                        for i, div in enumerate(divisions)
+                        if i in part_filter[0]._partitions
+                    ]
+                    + [divisions[-1]]
```
milesgranger, line 1135: So this `part_filter` is always length==1 or do we only ever care about the first element in the list?

dask_expr/tests/test_categorical.py
```diff
@@ -48,3 +48,15 @@ def test_get_categories_simplify_adds_projection(df):
...
+
+
+def test_categorical_set_index():
+    df = lib.DataFrame({"x": [1, 2, 3, 4], "y": ["a", "b", "b", "c"]})
+    df["y"] = lib.Categorical(df["y"], categories=["a", "b", "c"], ordered=True)
+    a = from_pandas(df, npartitions=2)
+
+    b = a.set_index("y", divisions=["a", "b", "c"], npartitions=a.npartitions)
+    d1, d2 = b.get_partition(0), b.get_partition(1)
+    print(d1.index.optimize(fuse=False).pprint())
```
milesgranger, line 60: left over debug print?

milesgranger (approved):
Looks generally okay to me, only a couple things/clarifications

dask_expr/_shuffle.py
phofl: We care about the first child that is filtered, you could imagine somtethong like this:


expr -> restrict to partitions 1-5 -> another expr -> another expr -> restrict to partitions 2 -> expr -> us

We only care about the filter to 2, the fact that 1-5 was filtered earlier is not relevant for us

dask_expr/tests/test_categorical.py
phofl: Yep, thx for catching

dask_expr/_shuffle.py
milesgranger: ah, I see. Thanks!

PR #688, opened 2024-01-09
Enable the "disk" shuffle method for cudf

rjzamora:
Enables `shuffle_method="disk"` for cudf, and modifies several tests to work with both pandas and cudf.

### DIFF
--- a/dask_expr/_shuffle.py
+++ b/dask_expr/_shuffle.py
@@ -477,13 +477,16 @@ class DiskShuffle(SimpleShuffle):
             p.append(d, fsync=True)
 
     def _layer(self):
+        from dask.dataframe.dispatch import partd_encode_dispatch
+
         column = self.partitioning_index
         df = self.frame
 
         always_new_token = uuid.uuid1().hex
 
         p = ("zpartd-" + always_new_token,)
-        dsk1 = {p: (maybe_buffered_partd(),)}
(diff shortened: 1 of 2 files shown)


dask_expr/_shuffle.py
```diff
@@ -1118,7 +1121,8 @@ def _args(self) -> list:
 
     @staticmethod
     def operation(df, index_name, drop, set_name, column_dtype):
-        df = df.set_index(set_name, drop=drop).rename_axis(index=index_name)
+        df = df.set_index(set_name, drop=drop)
+        df.index.name = index_name
```
rjzamora, line 1125: I know it's annoying, but cudf doesn't have a `rename_axis` API yet.

dask_expr/tests/test_shuffle.py
```diff
@@ -55,7 +55,9 @@ def test_disk_shuffle(ignore_index, npartitions, df):
         npartitions=npartitions,
         ignore_index=ignore_index,
     ).partitions[1]
-    assert set(a["x"].compute()).issubset(b["y"].compute())
+    assert set(a["x"].compute().values.tolist()).issubset(
+        b["y"].compute().values.tolist()
+    )
```
rjzamora, line 60: I know these changes are also annoying, but cudf tries to be loud about device -> host transfers than may be unintentional. We typically need to convert to cupy/numpy first before converting to a list.

PR #693, opened 2024-01-10
Raise error for invalid aggfunc in pivot_table

phofl:
ref #525

### DIFF
--- a/dask_expr/_collection.py
+++ b/dask_expr/_collection.py
@@ -2518,7 +2518,7 @@ def pivot_table(df, index, columns, values, aggfunc="mean"):
     if not methods.is_categorical_dtype(df._meta[columns]):
         raise ValueError("'columns' must be category dtype")
     if not has_known_categories(df._meta[columns]):
-        raise ValueError("'columns' categories must be known")
+        raise ValueError("'columns' must have known categories")
 
     if not (
         is_scalar(values)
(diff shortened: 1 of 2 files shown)


dask_expr/_collection.py
```diff
@@ -2528,6 +2528,13 @@ def pivot_table(df, index, columns, values, aggfunc="mean"):
     ):
         raise ValueError("'values' must refer to an existing column or columns")
 
+    available_aggfuncs = ["mean", "sum", "count", "first", "last"]
+
+    if not is_scalar(aggfunc) or aggfunc not in available_aggfuncs:
+        raise ValueError(
+            "aggfunc must be either " + ", ".join(f"'{x}'" for x in available_aggfuncs)
+        )
```
milesgranger, line 2536:
```suggestion
        raise ValueError(
            f"aggfunc must be a scalar or one of: {', '.join(available_aggfuncs)}"
        )
```
mostly b/c I read "either" to mean one of two.

dask_expr/tests/test_reshape.py
```diff
@@ -62,10 +62,14 @@ def test_pivot_table_fails(df):
...
-    with pytest.raises(ValueError, match="'columns' categories must be known"):
+    with pytest.raises(ValueError, match="'columns' must have known categories"):
         df2.pivot_table(index="a", columns="y", values="z")
 
     with pytest.raises(
         ValueError, match="'values' must refer to an existing column or columns"
     ):
         df.pivot_table(index="x", columns="y", values="aaa")
+
+    msg = "aggfunc must be either 'mean', 'sum', 'count', 'first', 'last'"
```
milesgranger, line 73:
```suggestion
    msg = "aggfunc must be a scalar or one of: mean, sum, count, first, last"
```
...should you choose the suggestion above.

milesgranger (approved):
Small nit is all, looks great to me otherwise.

dask_expr/_collection.py
phofl: The correction is not completely correct, lists are generally not allowed but we also only allow the list that is hardcoded.

This would need an update in dask/dask as well to handle the testing suite

phofl:
lets figure out the wording a little bit later

PR #694, opened 2024-01-10
Implement ufunc protocol

phofl:
ref #525

Starts with adding the ufunc protocol, next step will be to align the DataFrames, but don't want to make the PR too big

### DIFF
--- a/dask_expr/_collection.py
+++ b/dask_expr/_collection.py
@@ -26,7 +26,12 @@ from dask.dataframe.core import (
 )
 from dask.dataframe.dispatch import is_categorical_dtype, make_meta, meta_nonempty
 from dask.dataframe.multi import warn_dtype_mismatch
-from dask.dataframe.utils import has_known_categories, index_summary
+from dask.dataframe.utils import (
+    has_known_categories,
+    index_summary,
+    meta_frame_constructor,
+    meta_series_constructor,
+)
 from dask.utils import (
     IndexCallable,
     get_default_shuffle_method,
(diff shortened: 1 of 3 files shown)


dask_expr/_collection.py
```diff
@@ -1791,6 +1851,25 @@ def nbytes(self):
...
+                import inspect
+
+                method_name = f"`{inspect.stack()[3][3]}`"
+            except IndexError:
+                method_name = "This method"
+            raise NotImplementedError(
+                f"{method_name} is not implemented for `dask.dataframe.Series`."
+            )
+
+        return meta_series_constructor(self)(array, index=index, name=self.name)
```
milesgranger, line 1871: Hmm, two very similar `__array_wrap__` methods. What do you think about moving this whole thing into `FrameBase` dispatching to a `meta_constructor` which routes to the appropriate `meta_series/frame_constructor`? Or maybe a class attr in Series/DataFrame for the correct constructor, dependency injection style.

dask_expr/_collection.py
```diff
@@ -2649,3 +2731,41 @@ def isna(arg):
...
+
+    args = _maybe_from_pandas(args)
+
+    # TODO: Add align partitions
+
+    dfs = [df for df in args if isinstance(df, FrameBase)]
+
+    return new_collection(
+        expr.UFuncElemwise(dfs[0], op, meta, transform_divisions, kwargs, *args)
+    )
```
milesgranger, line 2771: It doesn't need `*args[:-1]`? 
```suggestion
    df = next(df for df in args if isinstance(df, FrameBase))

    return new_collection(
        expr.UFuncElemwise(df, op, meta, transform_divisions, kwargs, *args[:-1])
    )
```

milesgranger (approved):
Gave it the best look could, some things here I'm not very confident with, but looks good. Just a couple small comments/nits. Take them or leave them. :)

dask_expr/_collection.py
phofl: Yeah makes sense as a follow up although I am not sure if I am missing something, I took this mostly as is from dask/dask

dask_expr/_collection.py
phofl: This would change argument order, which is not good when we hand them over to the actual method, the current pattern isn't great but it's probably the best we can do

PR #695, opened 2024-01-10
Add some more keywords to read_csv

phofl:
ref #525

cc @scharlottej13 I would recommend setting engine=pyarrow and dtype_backend=pyarrow as well


milesgranger (approved):
Beautiful stuff. Thanks!

scharlottej13:
Thanks @phofl!

PR #697, opened 2024-01-10
Disable released ci and bump dask min version

phofl:
ref #525 

for now until the next release happens


phofl:
the other PR will fix ci

phofl:
#697 

Ah I forgot that the pin in version will kill the ci, will update

PR #699, opened 2024-01-10
Implement ufunc with non-aligned inputs

phofl:
ref #525

### DIFF
--- a/dask_expr/_collection.py
+++ b/dask_expr/_collection.py
@@ -274,38 +274,6 @@ class FrameBase(DaskMethodsMixin):
     def __array__(self, dtype=None, **kwargs):
         return np.array(self.compute())
 
-    def __array_ufunc__(self, numpy_ufunc, method, *inputs, **kwargs):
-        out = kwargs.get("out", ())
-        for x in inputs + out:
-            # ufuncs work with 0-dimensional NumPy ndarrays
-            # so we don't want to raise NotImplemented
-            if isinstance(x, np.ndarray) and x.shape == ():
-                continue
-            elif not isinstance(
-                x, (Number, Scalar, FrameBase, Array, pd.DataFrame, pd.Series, pd.Index)
-            ):
-                return NotImplemented
-
-        if method == "__call__":
-            if numpy_ufunc.signature is not None:
-                return NotImplemented
-            if numpy_ufunc.nout > 1:
-                # ufuncs with multiple output values
-                # are not yet supported for frames
-                return NotImplemented
-            else:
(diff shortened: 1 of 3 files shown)


phofl:
Scalar inheriting from FrameBase is becoming a problem, we should revisit this at some point

dask_expr/_collection.py
```diff
@@ -1851,6 +1856,31 @@ def nbytes(self):
...
+                return NotImplemented
+            if numpy_ufunc.nout > 1:
+                # ufuncs with multiple output values
+                # are not yet supported for frames
+                return NotImplemented
+            else:
+                return elemwise(numpy_ufunc, *inputs, **kwargs)
+        else:
+            # ufunc methods are not yet supported for frames
+            return NotImplemented
```
milesgranger, line 1882: I am a bit confused why it's a problem to not have this in FrameBase, it's exactly the same method copied to both classes, no?

dask_expr/_collection.py
```diff
@@ -2772,7 +2807,53 @@ def elemwise(op, *args, meta=no_default, out=None, transform_divisions=True, **k
...
-    return new_collection(
-        expr.UFuncElemwise(dfs[0], op, meta, transform_divisions, kwargs, *args)
-    )
+
+def handle_out(out, result):
+    """Handle out parameters
+
+    If out is a dask.DataFrame, dask.Series or dask.Scalar then
+    this overwrites the contents of it with the result
+    """
```
milesgranger, line 2824: I feel like this could use a bit more explanation. I'm at least confused. :sweat_smile:

dask_expr/_collection.py
```diff
@@ -2772,7 +2807,53 @@ def elemwise(op, *args, meta=no_default, out=None, transform_divisions=True, **k
...
+    if isinstance(out, tuple):
+        if len(out) == 1:
+            out = out[0]
+        elif len(out) > 1:
+            raise NotImplementedError("The out parameter is not fully supported")
```
milesgranger, line 2829:
```suggestion
        elif len(out) > 1:
            raise NotImplementedError("The `out` parameter with length > 1 is not supported")
```

milesgranger (approved):
Seems okay to me, things in here I'm not wholly competent with so will defer to your judgment. Thank you! :+1:

dask_expr/_collection.py
phofl: Scalar inherits from FrameBase as well and can't have access to it. There are hasattr checks involved unfortunately

PR #701, opened 2024-01-10
Implement alignment for map_overelap

phofl:
ref #525

sits on #700 and #699


milesgranger:
Will review once #699 is in. :)

phofl:
Rebased after the merge, so this should be ready to review as well

PR #704, opened 2024-01-10
Fix some parquet related things

phofl:
ref #525

most of the tests live over in dask/dask

### DIFF
--- a/dask_expr/_collection.py
+++ b/dask_expr/_collection.py
@@ -8,12 +8,12 @@ from collections.abc import Callable, Hashable, Mapping
 from numbers import Integral, Number
 from typing import Any, ClassVar, Literal
 
+import dask.dataframe.methods as methods
 import numpy as np
 import pandas as pd
 from dask import compute
 from dask.array import Array
 from dask.base import DaskMethodsMixin, is_dask_collection, named_schedulers
-from dask.dataframe import methods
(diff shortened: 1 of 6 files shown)


dask_expr/_collection.py
```diff
@@ -2555,6 +2555,25 @@ def from_map(
...
+            lengths.add(len(iterable))
+        except (AttributeError, TypeError):
+            iterables[i] = list(iterable)
+            lengths.add(len(iterables[i]))
+    if len(lengths) == 0:
+        raise ValueError("`from_map` requires at least one Iterable input")
+    elif len(lengths) > 1:
+        raise ValueError("All `iterables` must have the same length")
+    if lengths == {0}:
+        raise ValueError("All `iterables` must have a non-zero length")
```
milesgranger, line 2575: Couldn't this be `lengths = {0, 1}` and thus still need to fail?
```suggestion
    if 0 in lengths:
        raise ValueError("All `iterables` must have a non-zero length")
```

milesgranger (approved):
LGTM, w/ a small clarification. :)

dask_expr/_collection.py
phofl: That is covered by the previous condition that raises if you have more than one length

PR #710, opened 2024-01-11
Remove axis from idxmin

phofl:
ref #525

### DIFF
--- a/dask_expr/_groupby.py
+++ b/dask_expr/_groupby.py
@@ -1234,25 +1234,23 @@ class GroupBy:
         return self._single_agg(ValueCounts, **kwargs)
 
     def idxmin(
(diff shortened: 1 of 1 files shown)


milesgranger (approved):
Looks good, but why is it removed? Still in dask/dask and not deprecated there. Do we need to deprecate its use?

phofl:
I think it's deprecated over in dask/daks, e.g. 

```
    def idxmin(
        self,
        split_every=None,
        split_out=1,
        shuffle=None,
        axis=no_default,
        skipna=True,
        numeric_only=no_default,
    ):
        if axis != no_default:
            warnings.warn(
                "`axis` parameter is deprecated and will be removed in a future version.",
                FutureWarning,
            )
```

PR #714, opened 2024-01-11
Implement cov and corr

### LINKED ISSUE #713: Implement cov and corr
And autocorr

phofl:
ref #525

closes #713


milesgranger (approved):
Sublime. We could have tested for the invariants checked here, but in the name of progress, 'tis for another day! :+1:

PR #715, opened 2024-01-11
Actually test on different Python versions

crusaderky:
The action was quietly testing everything on Python 3.12


phofl (approved):
thx

its confusing, I got some internal pytest errors consistently on the 3.11 build and nowhere else, did this sneak in with your change a couple of days ago maybe=

PR #718, opened 2024-01-12
Implement loc indexer completely

phofl:
ref #525

sorry it's a lot but mostly copy past


milesgranger (approved):
Let's gooOOooOooo! Nice work! :+1:

PR #719, opened 2024-01-12
Fix more groupby related tests

phofl:
ref #525


milesgranger (approved):
Nothing stands out to me, looks good. Thank you.

PR #723, opened 2024-01-12
Use try block for ``is_any_real_numeric_dtype``

rjzamora:
https://github.com/dask-contrib/dask-expr/pull/705 added an `is_any_real_numeric_dtype` import. This makes sense, because dask-expr requires pandas>=2. However, I'd like to be able to develop with pandas-1.5 for a bit longer (even if many tests fail). This PR makes a minimal change so I can continue experimenting with ways to add/support backend dispatching.

### DIFF
--- a/dask_expr/_collection.py
+++ b/dask_expr/_collection.py
@@ -97,19 +97,13 @@ from dask_expr._util import (
     _BackendData,
     _convert_to_list,
     _get_shuffle_preferring_order,
+    _is_any_real_numeric_dtype,
     _maybe_from_pandas,
     _raise_if_object_series,
     _validate_axis,
     is_scalar,
 )
 from dask_expr.io import FromPandasDivisions, FromScalars
-
-# Temporary/soft pandas<2 support to enable cudf dev
-try:
-    from pandas.api.types import is_any_real_numeric_dtype
-except ImportError:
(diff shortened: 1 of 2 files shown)


dask_expr/_collection.py
```diff
@@ -2654,7 +2660,9 @@ def from_pandas(data, npartitions=None, sort=True, chunksize=None):
     if not has_parallel_type(data):
         raise TypeError("Input must be a pandas DataFrame or Series.")
 
-    if data.index.isna().any() and not is_any_real_numeric_dtype(data.index):
+    if is_any_real_numeric_dtype and (
```
phofl, line 2663: I don't like this pattern, can you create a function in ``_util`` that just returns True/False if is_any_real_numeric_dtype is not available? We should keep those checks out of the actual logic as much as possible

rjzamora: Yeah, sounds good. Thanks for being open to that. I appreciate your help with this.

phofl (approved):

dask_expr/_collection.py
phofl: As long as we can keep the number of these things very very low, dask/dask ci is a pain at the moment because a different upstream version might screw you over on every new pr

PR #725, opened 2024-01-12
Implement fillna_check for ffill and bfill

### LINKED ISSUE #529: Add fillna_check to bfill and ffill
see test_bfill in dask/dask, specifically the ValueError that is raised there

phofl:
ref #525

closes #529


milesgranger (approved):
Super, thank you!

PR #728, opened 2024-01-12
Add basic backend dispatching to dask-expr

rjzamora:
It seems like we will temporarily need to allow external libraries (i.e. cudf) to register a `DaskBackendEntrypoint` for both `dask.dataframe` and `dask_expr`. I think the easiest approach is to let cudf use a `"dask-expr.dataframe.backend"` entrypoint path for dask-expr (and continue using `"dask.dataframe.backend"` in the legacy `dask.dataframe` API).

This PR also introduces a simple `get_collection_type` dispatch function (the dask-expr version of `get_parallel_type`). This completely breaks cudf support for now, but dask-expr + cudf is already broken, and this is probably the only way to **fix** it anyway.

NOTE: This PR does not introduce any *new* dispatching mechanisms (like https://github.com/dask-contrib/dask-expr/pull/321). It just adds enough code to support the existing dispatching mechanisms used in `dask.dataframe`.

### DIFF
--- a/dask_expr/_backends.py
+++ b/dask_expr/_backends.py
@@ -4,11 +4,31 @@ import pandas as pd
 from dask.backends import CreationDispatch
 from dask.dataframe.backends import DataFrameBackendEntrypoint
 
-dataframe_creation_dispatch = CreationDispatch(
+from dask_expr._dispatch import get_collection_type
+
+
+class DXCreationDispatch(CreationDispatch):
+    """Dask-expressions version of CreationDispatch"""
+
+    # TODO Remove after https://github.com/dask/dask/pull/10794
+    def dispatch(self, backend: str):
+        from dask.backends import detect_entrypoints
+
+        try:
+            impl = self._lookup[backend]
+        except KeyError:
+            entrypoints = detect_entrypoints(f"dask-expr.{self._module_name}.backends")
+            if backend in entrypoints:
+                return self.register_backend(backend, entrypoints[backend].load()())
+        else:
+            return impl
+        raise ValueError(f"No backend dispatch registered for {backend}")
+
+
+dataframe_creation_dispatch = DXCreationDispatch(
     module_name="dataframe",
     default="pandas",
     entrypoint_class=DataFrameBackendEntrypoint,
-    entrypoint_root="dask_expr",  # Differs from `dask.dataframe`
     name="dataframe_creation_dispatch",
 )
 
@@ -35,3 +55,41 @@ class PandasBackendEntrypoint(DataFrameBackendEntrypoint):
 
 
 dataframe_creation_dispatch.register_backend("pandas", PandasBackendEntrypoint())
+
+
+@get_collection_type.register(pd.Series)
+def get_collection_type_series(_):
+    from dask_expr._collection import Series
+
+    return Series
+
+
+@get_collection_type.register(pd.DataFrame)
+def get_collection_type_dataframe(_):
+    from dask_expr._collection import DataFrame
+
+    return DataFrame
+
+
+@get_collection_type.register(pd.Index)
+def get_collection_type_index(_):
+    from dask_expr._collection import Index
+
+    return Index
+
+
+@get_collection_type.register(object)
+def get_collection_type_object(_):
+    from dask_expr._collection import Scalar
+
+    return Scalar
+
+
+######################################
+# cuDF: Pandas Dataframes on the GPU #
+######################################
+
+
+@get_collection_type.register_lazy("cudf")
+def _register_cudf():
+    import dask_cudf  # noqa: F401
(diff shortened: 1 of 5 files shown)


rjzamora:
@phofl - I know you are busy with many things, but I'd be interested to know if you have any thoughts/concerns about this PR. My hope is that there is nothing controversial here: We are reusing the existing dispatching mechanisms and conventions from `dask.dataframe`, and nothing should get in the way of default/pandas development.

dask_expr/_backends.py
```diff
@@ -0,0 +1,95 @@
...
+    """Dask-expressions version of CreationDispatch"""
+
+    # TODO Remove after https://github.com/dask/dask/pull/10794
+    def dispatch(self, backend: str):
+        from dask.backends import detect_entrypoints
+
+        try:
+            impl = self._lookup[backend]
+        except KeyError:
+            entrypoints = detect_entrypoints(f"dask_expr.{self._module_name}.backends")
```
rjzamora, line 20: The only reason we cannot use `CreationDispatch` directly is the fact that `dask_cudf` is already using the `"dask.dataframe.backends"` path to expose the `"cudf"` backend entrypoint for `dask.dataframe`.

We can just use `CreationDispatch` directly after https://github.com/dask/dask/pull/10794 makes it possible to modify the entrypoint path upon initialization.

dask_expr/_backends.py
```diff
@@ -0,0 +1,95 @@
...
+            return impl
+        raise ValueError(f"No backend dispatch registered for {backend}")
+
+
+dataframe_creation_dispatch = DXCreationDispatch(
+    module_name="dataframe",
+    default="pandas",
+    entrypoint_class=DataFrameBackendEntrypoint,
+    name="dataframe_creation_dispatch",
+)
```
rjzamora, line 33: Note that we want to introduce a **new** `CreationDispatch` object for dask-expr, because we want to distinguish "legacy" dask-dataframe creation functions from newer expression-based creation functions (to be discovered at the `"dask_expr.dataframe.backend"` entrypoint path instead of `"dask.dataframe.backend"`).

In both legacy and expr-based code, the same `"dask.dataframe.backend"` config value is used to specify the default backend.

dask_expr/_collection.py
```diff
@@ -3098,17 +3105,9 @@ def __array__(self):
...
     expr._name  # Ensure backend is imported
-    if is_dataframe_like(meta):
-        return DataFrame(expr)
-    elif is_series_like(meta):
-        return Series(expr)
-    elif is_index_like(meta):
-        return Index(expr)
-    else:
-        return Scalar(expr)
+    return get_collection_type(meta)(expr)
```
rjzamora, line 3110: Note that we were have been using `new_collection` from the beginning, because we know we needed to make this change to support other backends (i.e. cudf) long term.

The behavior of `new_collection` is now consistent with the behavior of `new_dd_object` in `dask.dataframe`.

dask_expr/_collection.py
```diff
@@ -1529,9 +1531,14 @@ def to_backend(self, backend: str | None = None, **kwargs):
         -------
         DataFrame, Series or Index
         """
-        from dask.dataframe.io import to_backend
+        from dask_expr._backends import dataframe_creation_dispatch
 
-        return to_backend(self.to_dask_dataframe(), backend=backend, **kwargs)
```
rjzamora, line 1534: The previous code here was just converting to `dask.dataframe` - Definitely not what we want.

dask_expr/_dispatch.py
```diff
@@ -0,0 +1,5 @@
+from __future__ import annotations
+
+from dask.utils import Dispatch
+
+get_collection_type = Dispatch("get_collection_type")
```
rjzamora, line 5: I don't think we will need many dispatch functions here. So, we could also define this in `dask.datafame.dispatch`. For example, we could call it `get_expr_collection_class` in `dask.datafame` to avoid confusion.

dask_expr/_backends.py
rjzamora: Scratch that - I should probably wait for the next dask/dask release to remove DXCreationDispatch

dask_expr/_dispatch.py
rjzamora: This is now publicly exposed as `dask_expr.get_collection_type`. This can always be moved to `dask.datafame.dispatch` in the future.

phofl:
I've disabled released ci because we enabled a couple of other things that aren't in released dask yet. You can remove everything that depends on released here and wouldn't be necessary with main now. 

I don't intend to release before the next dask release anyway, so testing against the released version isn't very important

rjzamora:
>You can remove everything that depends on released here and wouldn't be necessary with main now.

Awesome. Thanks!

phofl (approved):

PR #731, opened 2024-01-15
Implement npartitions for head properly

phofl:
ref #525 

this addresses a bunch of dask/dask tests

### DIFF
--- a/dask_expr/_expr.py
+++ b/dask_expr/_expr.py
@@ -25,6 +25,7 @@ from dask.dataframe.core import (
     is_index_like,
     is_series_like,
     make_meta,
+    safe_head,
     total_mem_usage,
 )
 from dask.dataframe.dispatch import meta_nonempty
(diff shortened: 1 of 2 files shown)


dask_expr/_expr.py
```diff
@@ -1921,21 +1931,44 @@ def _simplify_up(self, parent, dependents):
...
+            frame = BlockwiseHead(
+                Partitions(self.frame, partitions), self.n, npartitions, safe
+            )
+            if npartitions != 1:
+                from dask_expr import Repartition
+
+                safe = (
+                    True
+                    if npartitions != self.frame.npartitions and npartitions != -1
+                    else False
```
milesgranger, line 1955: nit:
```suggestion
                safe = (
                    npartitions != self.frame.npartitions and npartitions != -1
```

dask_expr/_expr.py
```diff
@@ -1945,11 +1978,21 @@ class BlockwiseHead(Head, Blockwise):
...
     def _divisions(self):
-        return self.frame.divisions
+        return self.frame.divisions[: len(self._partitions) + 1]
 
     def _task(self, index: int):
-        return (M.head, (self.frame._name, index), self.n)
+        if self.safe:
+            op = safe_head
+        else:
+            op = M.head
```
milesgranger, line 1994: nit, compacts down slightly nicer.
```suggestion
        op = safe_head if self.safe else M.head
```

milesgranger (approved):
No pressure to take the nits, very minimal. Looks great otherwise. :+1:

dask_expr/_expr.py
phofl: I like spreading them out better here

PR #732, opened 2024-01-15
Rename backend in shuffle to method

phofl:
ref #525 

Mostly internal renaming after the keyword argument renaming


dask_expr/_shuffle.py
```diff
@@ -76,7 +76,7 @@ class Shuffle(Expr):
         Number of output partitions.
     ignore_index: bool
         Whether to ignore the index during this shuffle operation.
-    backend: str or Callable
+    "method": str or Callable
```
milesgranger, line 79:
```suggestion
    method: str or Callable
```

milesgranger (approved):
One suggestion. Thanks!

PR #733, opened 2024-01-15
Implement axis for clip

phofl:
ref #525

### DIFF
--- a/dask_expr/tests/test_collection.py
+++ b/dask_expr/tests/test_collection.py
@@ -609,6 +609,7 @@ def test_to_timestamp(pdf, how):
         lambda df: df.astype(int),
         lambda df: df.clip(lower=10, upper=50),
         lambda df: df.x.clip(lower=10, upper=50),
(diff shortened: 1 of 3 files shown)


dask_expr/tests/test_collection.py
```diff
@@ -609,6 +609,7 @@ def test_to_timestamp(pdf, how):
         lambda df: df.astype(int),
         lambda df: df.clip(lower=10, upper=50),
         lambda df: df.x.clip(lower=10, upper=50),
+        lambda df: df.clip(lower=3, upper=7, axis=1),
```
milesgranger, line 612: Do we want `None` and `0` cases for axis?

milesgranger (approved):
Good for me with one clarification.

dask_expr/tests/test_collection.py
phofl: Dask/dask has those, so not super concerned

PR #737, opened 2024-01-15
Fix metadata propagation in map_partition

phofl:
ref #525

### DIFF
--- a/dask_expr/_groupby.py
+++ b/dask_expr/_groupby.py
@@ -1068,6 +1068,7 @@ class GroupByCumulative(Expr, GroupByBase):
             True,
             False,
             True,
+            None,
             {"chunk": self.chunk, "columns": columns, **dropna, **self.numeric_only},
             *self.by,
         )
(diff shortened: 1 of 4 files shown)


dask_expr/_groupby.py
```diff
@@ -1095,6 +1096,7 @@ def _lower(self):
             True,
             False,
             True,
+            None,
```
milesgranger, line 1099: Oh man, what's the policy on named kwargs? Looking at a list of (True, True, False, True, None, ..) is not fantastic.

milesgranger (approved):
Only a general comment, but superb otherwise! Thank you!

dask_expr/_groupby.py
phofl: 100% agree

There is no real policy, its a technical necessity as long as we have the dynamic thing without constructors

If you need *args like here a couple lines below, you can't pass kwargs because the constructor can only deal with it like this

PR #738, opened 2024-01-15
Fix index preservation in drop duplicates

phofl:
ref #525


milesgranger (approved):
Short 'n sweet, sublime. Thanks!

PR #739, opened 2024-01-15
Use delayed for isin values

phofl:
ref #525

the solution isn't great but it's better than what we had before


milesgranger (approved):
> the solution isn't great but it's better than what we had before

A little better every day! :100:

PR #744, opened 2024-01-15
Remove simplification for decimal

phofl:
ref #525

this screwed up decimal ops and I don't see an objective benefit for this optimization


milesgranger (commented):
Can review again when tests are happy.

phofl:
updated

milesgranger (approved):
Thanks, @phofl!

PR #745, opened 2024-01-16
Emulate meta properly for apply and map

phofl:
ref #525


milesgranger (approved):
Most excellent, thanks!

PR #747, opened 2024-01-16
Implement random split

### LINKED ISSUE #652: Implement DataFrame.random_split?
Not sure what this is actually useful for, maybe we should just straight out deprecate?

cc @mrocklin do you know what the intention is for this method?

phofl:
ref #525

closes #652

### DIFF
--- a/dask_expr/_collection.py
+++ b/dask_expr/_collection.py
@@ -1170,6 +1170,44 @@ class FrameBase(DaskMethodsMixin):
     def isna(self):
         return new_collection(self.expr.isna())
 
+    def random_split(self, frac, random_state=None, shuffle=False):
+        """Pseudorandomly split dataframe into different pieces row-wise
+
+        Parameters
+        ----------
+        frac : list
+            List of floats that should sum to one.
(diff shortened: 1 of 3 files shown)


dask_expr/_collection.py
```diff
@@ -1170,6 +1170,44 @@ def to_timestamp(self, freq=None, how="start"):
...
+    def random_split(self, frac, random_state=None, shuffle=False):
+        """Pseudorandomly split dataframe into different pieces row-wise
+
+        Parameters
+        ----------
+        frac : list
+            List of floats that should sum to one.
+        random_state : int or np.random.RandomState
+            If int create a new RandomState with this as the seed.
+            Otherwise draw from the passed RandomState.
```
milesgranger, line 1182: What if `None`?

milesgranger (approved):
Looks goodie, w/ one docs nit. :+1:

dask_expr/_collection.py
phofl: same as with int, is passed to numpy. added to the docs

PR #748, opened 2024-01-16
Implement divmod

### LINKED ISSUE #685: Add __divmod__ and __rdivmod__ operators

phofl:
ref #525

closes #685

### DIFF
--- a/dask_expr/_collection.py
+++ b/dask_expr/_collection.py
@@ -735,6 +735,14 @@ class FrameBase(DaskMethodsMixin):
     def values(self):
         return self.to_dask_array()
 
+    def __divmod__(self, other):
+        result = self.expr.__divmod__(other)
+        return new_collection(result[0]), new_collection(result[1])
+
+    def __rdivmod__(self, other):
+        result = self.expr.__rdivmod__(other)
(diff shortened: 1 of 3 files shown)


dask_expr/_collection.py
```diff
@@ -735,6 +735,14 @@ def to_dask_array(
     def values(self):
         return self.to_dask_array()
 
+    def __divmod__(self, other):
+        result = self.expr.__divmod__(other)
+        return new_collection(result[0]), new_collection(result[1])
+
+    def __rdivmod__(self, other):
+        result = self.expr.__rdivmod__(other)
+        return new_collection(result[0]), new_collection(result[1])
```
milesgranger, line 744:
```suggestion
    def __rdivmod__(self, other):
        result = other.expr.__rdivmod__(self)
        return new_collection(result[0]), new_collection(result[1])
```
Is that right?

milesgranger (approved):
LGTM, w/ one clarification

dask_expr/_collection.py
milesgranger: Ah, nvm, You had it.

PR #749, opened 2024-01-16
Restructure shuffle architecture

phofl:
cc @fjetter 

This is roughly what we talked about

### DIFF
--- a/dask_expr/_shuffle.py
+++ b/dask_expr/_shuffle.py
@@ -63,28 +63,7 @@ from dask_expr._repartition import Repartition, RepartitionToFewer
 from dask_expr._util import LRU, _convert_to_list, is_valid_nth_dtype
 
 
-class Shuffle(Expr):
-    """Abstract shuffle class
-
-    Parameters
-    ----------
-    frame: Expr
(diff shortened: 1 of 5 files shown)


dask_expr/_shuffle.py
```diff
@@ -179,54 +141,71 @@ def _layer(self):
 
     @functools.cached_property
     def _meta(self):
-        # We will drop _partitions later on, so reflect this here
-        return self.frame._meta.drop(columns=["_partitions"], errors="ignore")
+        return self.frame._meta
```
fjetter, line 144: A little context for other potential reviewers

I noticed weird optimization behaviors where sometimes the `_partitions` column would not be dropped because the optimizer would think that this column wasn't there, i.e. this meta implementation is _wrong_.

This led us to adding explicit `Projection`s below. Since the `from_abstract_shuffle` is doing a similar thing but a little more sophisticated we had to touch this as well.

fjetter (approved):
LGTM assuming tests are green

phofl:
didn't rerun all of them, but the most relevant ones are green

PR #755, opened 2024-01-16
Partially address graph blowup in index based merges

phofl:
cc @fjetter

### DIFF
--- a/dask_expr/_shuffle.py
+++ b/dask_expr/_shuffle.py
@@ -241,7 +241,7 @@ class RearrangeByColumn(ShuffleBase):
                 ]
             drop_columns = partitioning_index.copy()
         elif index_shuffle:
-            dtypes = np.float64 if _is_numeric_cast_type(dtype) else None
+            dtypes = np.float64 if _is_numeric_cast_type(frame.index._meta.dtype) else None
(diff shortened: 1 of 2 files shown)


hendrikmakait (approved):
Thanks, @phofl! Looks reasonable to me

dask_expr/_shuffle.py
```diff
@@ -241,9 +241,7 @@ def _lower(self):
                 ]
             drop_columns = partitioning_index.copy()
         elif index_shuffle:
-            frame = Assign(frame, "_partitions_0", frame.index)
-            partitioning_index = ["_partitions_0"]
-            drop_columns = partitioning_index.copy()
+            dtypes = np.float64 if _is_numeric_cast_type(dtype) else None
```
hendrikmakait, line 244: dtype isn't defined here.

hendrikmakait (changes requested):

dask_expr/_shuffle.py
phofl: Thx, I should stop making non-trivial changes in the ui

hendrikmakait (approved):
Thanks for fixing!

PR #756, opened 2024-01-16
Add Various rolling fixes

### LINKED ISSUE #494: MapOverlap look at more than one partition
Currently, we only look at the previous partition and raise if this doesn't capture the whole window that is requested.

I am not convinced that looking at more than the previous partition is a real use-case, but we should make a decision

phofl:
ref #525

closes #494


milesgranger (approved):
From what I see, this is good stuff. Thank you!

PR #757, opened 2024-01-16
Change split_out default for value_counts

### LINKED ISSUE #670: Inconsistent default for split_out in unique methods
`drop_duplicates` and `unique` default to split_out=True, while `value_counts` defaults to split_out=1. 
As these three methods are conceptually very similar, I think they should have a coherent default.

phofl:
ref #525 

closes #670


milesgranger (approved):
Most excellent!

PR #758, opened 2024-01-16
Adjust tests for fixed bug

### LINKED ISSUE #485: Changing columns.dtype after dataframe creation may produce wrong output dtype
1. Update the columns of a DataFrame after creation, changing the dtype
2. Slice to select no columns
3. `df.columns.dtype` matches pandas, but `df.compute().columns.dtype` doesn't

This, among other things, breaks `assert_eq`.

```python
import pandas as pd
import dask_expr as dx

pdf = pd.DataFrame([[1,2],[3,4]], columns=["x", "y"])
df = dx.from_pandas(pdf)
pdf.columns = [10, 20]
df.columns = [10, 20]
pdf = pdf[[]]
df = df[[]]

print(pdf.columns.dtype)  # int
print(df.columns.dtype)  # int
print(df.compute().columns.dtype)  # object
```

phofl:
ref #525

closes #485 

this seems to work now

### DIFF
--- a/dask_expr/tests/test_indexing.py
+++ b/dask_expr/tests/test_indexing.py
@@ -44,12 +44,9 @@ def test_loc(df, pdf):
     assert_eq(df.loc[df.y == 20, ["x"]], pdf.loc[pdf.y == 20, ["x"]])
 
 
-def test_loc_slice(pdf):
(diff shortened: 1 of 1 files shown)


milesgranger (approved):
:tada:

PR #759, opened 2024-01-16
Fix broadcasting rules for arithmetic methods

phofl:
ref #525

### DIFF
--- a/dask_expr/_collection.py
+++ b/dask_expr/_collection.py
@@ -171,14 +171,31 @@ def _wrap_expr_method_operator(name, class_):
 
             axis = _validate_axis(axis)
 
+            if (
+                is_dataframe_like(other)
+                or is_series_like(other)
+                and axis in (0, "index")
+            ) and not is_dask_collection(other):
+                other = self._create_alignable_frame(other)
+
             if axis in (1, "columns"):
                 if isinstance(other, Series):
                     msg = f"Unable to {name} dd.Series with axis=1"
                     raise ValueError(msg)
+
+            frame = self
+            if isinstance(other, FrameBase) and not expr.are_co_aligned(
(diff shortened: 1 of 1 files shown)


dask_expr/_collection.py
```diff
@@ -194,10 +211,22 @@ def method(self, other, level=None, fill_value=None, axis=0):
...
+                other = self._create_alignable_frame(other)
+
+            frame = self
+            if isinstance(other, FrameBase) and not expr.are_co_aligned(
+                self.expr, other.expr, allow_broadcast=False
+            ):
+                divs = expr.calc_divisions_for_align(self.expr, other.expr)
+                frame, other = expr.maybe_align_partitions(
+                    self.expr, other.expr, divisions=divs
+                )
```
milesgranger, line 224: During these review, I notice it's not uncommon for duplicated (or very similar) blocks of code to be found. I'm rather assuming this is a product of driving up the pace and maybe in the future we can do a bit of reduction. Or is it a consequence of the overall design? I remember inheriting one method too far up caused problems somehow.

milesgranger (approved):
Only a general note/question; otherwise tip top condition!

dask_expr/_collection.py
phofl: Good catch, I don't like the current setup either, we have a bit too much duplication at the moment.

Ideally, I would like to capture this in an expression class that lowers itself into the alignment operation and the blockwise operation that does the actual work. I couldn't come up with a design that doesn't need one class per method where we need the alignment to happen yet, which is why I didn't push forward on this front.

I am very open to suggestions if you have any ideas

Some other things need a bit of deduplication as well, but there are mostly edge cases involved that make this a PITA...

PR #760, opened 2024-01-16
Fix groupby.apply with args

phofl:
ref #525

not the most elegant solution but way easier to maintain. The previous solution made my head spin every time I touched it


milesgranger (approved):
LGTM, thanks!

Issue #14, opened 2023-04-04
Distributed protocol


mrocklin:
We would like to send expressions up to the Dask scheduler have them optimized and lowered there and then enter the normal scheduler state.  To do this I think that we need a new route on client/scheduler communication.  Maybe that can be made more general and accept "something picklable that satisfies the `.__dask_foo__` protocols" or something similar.

Eventually I'd love it if we held on to the Expr structure in the scheduler to inform future optimizations, but as a first pass I think that we probably just try to end up with the same state that the current `update_graph` route achieves.

@fjetter I think that this would benefit from someone on your team

fjetter:
I'm not sure if this is required this early already. While I understand that we might want to have knowledge about the raw input on the scheduler eventually, hooking this up to distributed would be much easier if we did the optimization and materialization on client side first as long as this is still a moving target.

If the expressions can be materialized to an existing HLG or to a low level task / MaterializedGraph we wouldn't need an additional route on the scheduler. At this point I wouldn't know how a protocol like this would even look like and would prefer delaying changes to distributed until we know more about this

mrocklin:
> I'm not sure if this is required this early already.

Yeah, nothing is required yet, we're mostly trying to identify the hard parts of this problem, and make sure that we can solve them well.  Then, once we've done that demonstration it'll make sense to do a lot of the tedious work of moving over the bulk of the algorithms.

> as long as this is still a moving target

My hope is that by agreeing on some stable protocols we can make everyone happy.  

> At this point I wouldn't know how a protocol like

Yeah, let's think about this.  My recommendation would be that the scheduler receives a pickleable thing that satisfies various `__dask_foo__` protocols.  For example, it might look like the following:

```python
def update_graph(self, obj, ...):
    graph = obj.__dask_graph__()
    keys = obj.__dask_keys__()
    annotations = obj.__dask_annotations__()

    priority = dask.order(graph)
    ...
```

mrocklin:
If so, it would be great to have near-instantaneous execution.  I suspect that it will viscerally feel nice.

fjetter:
I poked at this briefly. This is unfortunately a bit of work. HLGs are hard coded in so many places. Introducing this protocol and allowing this to be accepted in `Scheduler.update_graph` is trivial but on client side there are so many verifications and graph mutations happening that all need to be thrown away / bypassed for a new object. All places that are not immediately assuming a HLG are falling back to materialization via `__dask_graph__`, `__dask_keys__`.

[...]

PR #765, opened 2024-01-18
add test that shows how lambda tokenization is broken

fjetter:
https://github.com/dask-contrib/dask-expr/pull/223 had good intention and introduced a utility function `_tokenize_deterministic` that is setting the `tokenize.ensure-deterministic` flag such that we can ensure that the tokens we're generating are indeed deterministic. 
Determinism in this context is mildly ambiguous and there are subtly different consistency guarantees we need if we want to pickle `Expr` objects in a way that is deterministic across python processes/interpreters

Essentially everything we'd need to fall back to cloudpickle to serialize is also susceptible to not being deterministically tokenized (therefore this loosely relates to https://github.com/dask-contrib/dask-expr/pull/331#discussion_r1357189975 even though those are separate issues) since the tokenization is using cloudpickle behind the scenes for those cases. In way that defeats the entire purpose of the tokenization engine (see point 3. below).

https://github.com/dask-contrib/dask-expr/pull/223 also implemented a "fix" for `LambdaType` functions which is using the `str` representation of those functions which looks like `<function func.<locals>.<lambda> at 0x107363be0>`. This is unique inside of an interpreter session since it includes the memory address where the lambda is stored but this changes once moved to another interpreter.

This effectively hard blocks https://github.com/dask-contrib/dask-expr/issues/14

There are a couple of options (I'm not really happy about either but here we go...)

1. Enforce pickle as a serialization protocol and forbid usage of local functions, types, etc. (effectively https://github.com/dask-contrib/dask-expr/pull/331 but this would expand to all user APIs. We could wrap user defined objects with some magical class but this would feel a little awkward I think)
1. A slightly better/milder version of 1). I believe we can define deterministic cross-interpreter tokens for lambdas/local functions. I had something like this in a version of https://github.com/dask/dask/pull/10808 but it turned out to not be python version stable so I ended up removing it again. I'd have to run a couple 
1. Abandon tokenization entirely. Inside of an interpreter this problem is easy since if an object doesn't implement a `__hash__` we could just use it's `id` for a unique identifier and hash (or just stick with the current engine although it feels slightly over engineered then). From what I understand the entire tokenization is merely for the distributed scheduler but I see a way how to avoid that. This would require us to rewrite how client and scheduler are communicating in `dask/distributed` but I believe this is doable with a reasonable amount of effort. The scheduler would de-facto be the source of truth and be responsible for key generation and not the other way round. Locally generated keys would no longer map to remote keys. That would be fine, I guess?

cc @rjzamora @mrocklin @phofl 

Note: This problem has been around forever. HLGs were not as much affected by this since their names are actually not unique (e.g. https://github.com/dask/dask/issues/9888). `Expr` are calculating names/tokens recursively which is objectively much better for key uniqueness but also means that if there is any object inside of the expr-tree that cannot be tokenized deterministically, this affects the entire tree.

### DIFF
--- a/dask_expr/tests/test_util.py
+++ b/dask_expr/tests/test_util.py
@@ -1,6 +1,43 @@
+import pickle
+
 import pytest
 
-from dask_expr._util import RaiseAttributeError
+from dask_expr._util import RaiseAttributeError, _tokenize_deterministic
+
+
+def _clear_function_cache():
+    from dask.base import function_cache, function_cache_lock
+
+    with function_cache_lock:
+        function_cache.clear()
+
+
+def tokenize(x, /, __ensure_deterministic=True, *args, **kwargs):
+    _clear_function_cache()
+    try:
+        before = _tokenize_deterministic(x, *args, **kwargs)
+        _clear_function_cache()
+        if __ensure_deterministic:
+            assert before == _tokenize_deterministic(x, *args, **kwargs)
+            _clear_function_cache()
+        try:
+            after = _tokenize_deterministic(
+                pickle.loads(pickle.dumps(x)), *args, **kwargs
+            )
+        except (AttributeError, pickle.PicklingError):
+            # If we go down this path we're almost certainly guaranteed to fail
+            # since cloudpickle dumps are not deterministic and the tokenization
+            # is relying on that
+            import cloudpickle
+
+            after = _tokenize_deterministic(
+                cloudpickle.loads(cloudpickle.dumps(x)), *args, **kwargs
+            )
+
+        assert before == after
+        return before
+    finally:
+        _clear_function_cache()
 
 
 def test_raises_attribute_error():
@@ -19,3 +56,36 @@ def test_raises_attribute_error():
         B.x
     with pytest.raises(AttributeError, match="'B' object has no attribute 'x'"):
         B().x
+
+
+def test_tokenize_lambda():
+    func = lambda x: x + 1
+    tokenize(func)
+
+
+import pandas as pd
+
+from dask_expr import from_pandas
+
+
+def test_tokenize_deterministic():
+    ddf = from_pandas(pd.DataFrame({"a": [1, 2, 3]}))
+    tokenize(ddf)
+
+    def identity(x):
+        return x
+
+    ddf2 = ddf.map_partitions(identity)
+    tokenize(ddf2)
+
+    def identity(x):
+        return x + 1
+
+    ddf3 = ddf.map_partitions(identity)
+
+    from distributed.protocol.pickle import dumps, loads
+
+    assert loads(dumps(ddf))._name == ddf._name
+
+    assert loads(dumps(ddf2))._name == ddf2._name
+    assert tokenize(ddf2) != tokenize(ddf3)
--- a/dask_expr/tests/test_util.py
+++ b/dask_expr/tests/test_util.py
@@ -70,13 +70,16 @@ from dask_expr import from_pandas
 
 def test_tokenize_deterministic():
     ddf = from_pandas(pd.DataFrame({"a": [1, 2, 3]}))
-    tokenize(ddf)
+    assert ddf._name == ddf._name
+    # This fails without https://github.com/dask/dask/pull/10808 already
+    # tokenize(ddf)
 
     def identity(x):
         return x
 
     ddf2 = ddf.map_partitions(identity)
-    tokenize(ddf2)
+    # This fails without https://github.com/dask/dask/pull/10808 already
+    # tokenize(ddf2)
 
     def identity(x):
         return x + 1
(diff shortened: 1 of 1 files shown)


fjetter:
Unless there are other options I am not seeing, 3) would be the most permissive thinking about users. Since this is also the one with the most uncertainty I'll probably invest some time and whip up a prototype outlining what I'm thinking.

Short term we may just postpone scheduler integration since I don't think this is critical for a release but I'd like to know what our options are. If we were forced to go with option 1) it would be good to know before the release

fjetter:
There is also the option of making cloudpickle actually deterministic. There has been a little progress on that front https://github.com/cloudpipe/cloudpickle/issues/453 https://github.com/cloudpipe/cloudpickle/pull/524

mrocklin:
Can someone help me understand the cost of non-deterministic hashing in
dask-expr?  Historically hashing was a nice way to avoid work, but if it
didn't work then 🤷‍♂️ we just did a bit more work.  I'm guessing that this
comes down to the fact that we recreate the same object repeatedly in the
optimization process, and need to see if the object is the same or not.  Is
that right?

phofl:
The assumption that the name attribute of an expression is the same if the expression has exactly the same operands is deeply baked into our optimization logic

fjetter:
Apart form the optimization logic itself, the way distributed is currently built is roughly as

### Client side

- Take expr and optimize it
- Generate the final output keys
- Use those keys to generate `Future`s we can fetch and listen to

### Scheduler side
- Take the expression and materialize it
- Compute stuff
- If one of the keys that the client asked for is ready, let the client know (If the keys generated on client side don't match with the actual graph keys we'll deadlock because the client is waiting for the results of keys that don't exist

Changing this procedure is what I meant with 3) but this is a bit of work. I was hoping to not have to do this, particularly not while still supporting HLGs

mrocklin:
One dumb approach (maybe not the approach that we want to take) is that we optionally store a finalized key on the expression before we serialize it.  So on the client side we do all of our optimizations, get a nice expr that we like, finalize the key on the expression, then serialize it up to the scheduler.  The `_name` property checks for that finalized key and if it exists uses it instead of tokenization.

This seems like it would be easy short term.  It would stop us from doing scheduler-side optimization (which I think we're all excited about medium-term) but might give us the fast client-scheduler communications we want in the meantime.

rjzamora:
I haven't looked carefully through this code or other linked work yet. However, just to check my rough understanding - It seems like there are two challenges here:

- Client-side dask-expr depends on deterministic tokenization to function. We identify an expression by its name, and its name depends directly on the tokenization of its operands. There are some cases (e.g. local functions) where the necessary "determinism" is broken and/or it limits what a user can do. Possible solutions:
    - Find a way to avoid tokenization (seems difficult)
    - Use something like `inspect.getsource` to automatically convert local/lambda objects to a deterministic strings (probably still has problems/limitations)
    - Something else...
- In order for us to materialize expressions on the scheduler, we need the expression names to be the same on the client and scheduler. However, even some objects that pass the `_tokenize_deterministic` test will not actually tokenize to the same value on the client and scheduler. Possible solutions:
    - We do something like Matts last suggestion. To unlock scheduler-side optimizations in the future, it may even be possible to "freeze" the keys on the client and then "remap" them on the scheduler. Things should be okay if the scheduler keeps track of the mapping between the new and original keys (maybe?).
    - Something else...

fjetter:
I'd be fine with a finalized name after optimization

rjzamora:
@fjetter - Just a note that I submitted https://github.com/dask-contrib/dask-expr/pull/767 to demonstrate how the tests added in this PR will pass if we use `inspect` to normalize local/lambda functions.

fjetter:
reposting here

the most recent commit on main of cloudpickle (not released yet) likely solves the lambda problem but there are still cases that are not handled well.

It's easy for us to verify on scheduler side if the tokenization was successful so we can abort with a clear exception. Maybe the new cloudpickle plus that scheduler side verification is good enough for us?

crusaderky:
# Status update

The PRs linked from https://github.com/dask/dask/issues/10905 make lambdas tokenization deterministic, with the current released version of cloudpickle, *on the same Python interpreter*. Nothing should be expected across processes or hosts.

So we'll either need tokens to be generated on the client and then shipped to the scheduler, or the scheduler to generate them first and ship them back to the client (the latter is more involved because as of today we use batched comms, which is one-directional, but the client would need to be informed of what keys they just gained in who_wants).

It would be trivial to add a single line to the dask tokenize method that bakes in an unique interpreter ID, so that we can see things fail hard and fast in trivial use cases instead of waiting for the edge cases of cloudpickle.

It of course remains possible to have some obscure third-party object (callable or not) that doesn't cloudpickle deterministically AND comes with neither `__dask_tokenize__` method nor `normalize_token.register` attached to it. Nothing we can do about that besides being as loud as possible when that happens (read: don't silently hang).

crusaderky:
Superseded by #https://github.com/dask/dask/pull/10883

PR #767, opened 2024-01-18
Use ``inspect`` to deterministically tokenize local/lambda functions

rjzamora:
Possible/partial solution to https://github.com/dask-contrib/dask-expr/pull/765

This is for demonstration purposes - Needs further testing/discussion.

### DIFF
--- a/dask_expr/_util.py
+++ b/dask_expr/_util.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import functools
+import inspect
 from collections import OrderedDict, UserDict
 from collections.abc import Hashable, Sequence
 from types import LambdaType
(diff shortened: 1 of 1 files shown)


fjetter:
I was using / trying to use this before over in https://github.com/dask/dask/pull/10808 but rejected the idea again. Sorry, I don't even remember why but something was missing.

One thing I remember is that inspect just fails at times

mrocklin:
Lambdas can also contain other important scope, like closed-over variables.

fjetter:
the most recent commit on main of cloudpickle (not released yet) likely solves the lambda problem but there are still cases that are not handled well.

rjzamora:
>Lambdas can also contain other important scope, like closed-over variables.

Totally agree with this. However, I'm still curious how much this matters for the tokenization purpose needed in dask-expr. (I'm sure we can come up with cases that demonstrate a problem, but I couldn't immediately imagine an obvious problem)

PR #768, opened 2024-01-18
Implement multi column assign

phofl:
ref #525

we can make this a little bit more efficient, but this is a good starting point


milesgranger (approved):
LGTM.  Thanks!

PR #769, opened 2024-01-18
Avoid assignment in shuffle

phofl:
ref #525 

next step towards ripping assign out of shuffles


milesgranger (approved):
Nice, thank you!

Issue #780, opened 2024-01-19
We don't seem to calculate divisions if ``calculate_divisions=True`` and filters are given


phofl:
See ``test_pyarrow_filter_divisions`` in dask/dask

optimise doesn't help, they are still None

cc @rjzamora if you have thoughts

rjzamora:
I'll be happy to investigate this (and #779) - Do you have specific reproducers in mind? (I can come up with something, but just wanted to check first)

phofl:
The test in the top post is from dask/dask and reproduces, that's something you can use

Issue #779, opened 2024-01-19
We don't seem to reduce the partition count when a read parquet filter is given


phofl:
See ``test_filters`` in ``test_parquet.py`` over in dask/dask

cc @rjzamora if you have thoughts

Calling optimise does not help here

rjzamora:
I don't see a failure when I run `DASK_DATAFRAME__QUERY_PLANNING=True py.test test_parquet.py::test_filters`. Am I looking at the wrong test? (note that my environment is a bit off, but just wanted to check before I started fresh)

phofl:
I disabled the checks that would fail so that ci passes

There is a block


```
    if not DASK_EXPR_ENABLED:
        # TODO: this doesn't reduce the partition count
        e = dd.read_parquet(tmp_path, engine=read_engine, filters=[("x", "in", (0, 9))])
        assert e.npartitions == 2
        assert ((e.x < 2) | (e.x > 7)).all().compute()
```

in there that you have to include to see the failures. I wanted to keep the other checks over just xfailing the whole test

PR #795, opened 2024-01-19
Align parquet filtering behavior with ``dask.dataframe``

rjzamora:
Closes https://github.com/dask-contrib/dask-expr/issues/780
Closes https://github.com/dask-contrib/dask-expr/issues/779

Makes a few small changes to `ReadParquet` to make divisions/npartitions the same as `dask.dataframe`.


hendrikmakait:
Do we have a sibling PR for `dask/dask` that adjusts the tests to accout for this change?

phofl:
thx

Issue #796, opened 2024-01-20
[Bug] Optimization is now much slower in TPCh benchmarks


rjzamora:
I tried running the TPCh benchmarks in https://github.com/coiled/benchmarks and noticed that dask-expr was much slower than before (both with "pandas" and "cudf" as the backend). After looking at a performance report, it was clear that the slow down was not in compute. I confirmed that the regression was coming from `optimize()` (a change from ~1s to >10s for query 1).

I did a cursory bisection, and found that the regression was definitely introduced in https://github.com/dask-contrib/dask-expr/pull/395

I haven't had time to figure out **why** 395 is slowing things down yet.

fjetter:
Haven't been looking for the cause but the slowness is there because the expression graph for Q1 is huge. We encountered a similar problem recently in (https://github.com/dask-contrib/dask-expr/pull/294; IIRC https://github.com/dask-contrib/dask-expr/pull/769 relates to a partial fix since it was removing a couple of redundant `Assign` expressions)

Printing the entire expr takes 500k lines so it's hard to follow what is happening in detail.

Here's one excerpt that show how we're duplicating multiple `Assign` and `Projection` columns on different levels (this is already dedented)

```python
Projection: columns='l_extendedprice'
  Assign: key='avg_qty'
    Assign: key='sum_base_price'
      Assign: key='sum_qty'
        Filter:
          ReadParquet: path='s3://coiled-runtime-ci/tpc-h/snappy/scale-100/lineitem' filesystem=None kwargs={'dtype_backend': None}
          LE: right=datetime.datetime(1998, 9, 2, 0, 0)
            Projection: columns='l_shipdate'
              ReadParquet: path='s3://coiled-runtime-ci/tpc-h/snappy/scale-100/lineitem' filesystem=None kwargs={'dtype_backend': None}
        Projection: columns='l_quantity'
          Filter:
            ReadParquet: path='s3://coiled-runtime-ci/tpc-h/snappy/scale-100/lineitem' filesystem=None kwargs={'dtype_backend': None}
            LE: right=datetime.datetime(1998, 9, 2, 0, 0)
              Projection: columns='l_shipdate'
                ReadParquet: path='s3://coiled-runtime-ci/tpc-h/snappy/scale-100/lineitem' filesystem=None kwargs={'dtype_backend': None}
      Projection: columns='l_extendedprice'
        Assign: key='sum_qty'
```

[...]

PR #797, opened 2024-01-22
Add caching to recursive ``simplify_once`` calls

rjzamora:
The `simplify` logic currently collects a `dependents` dictionary, and then recursively calls `simplify_once` on all `Expr` objects in the expression graph. This PR adds a simple caching mechanism to avoid unnecessary repetition of the same logic (which is particularly problematic for column-name reassignment).

Note that an earlier version of this PR stored the cache in the `Expr` instance itself. The current approach is a bit lighter weight, but still provides a 10x performance boost for TPCh Q1 optimization.

Closes https://github.com/dask-contrib/dask-expr/issues/796

~Possible mitigation for https://github.com/dask-contrib/dask-expr/issues/796 - There may be better ways to implement/achieve the caching we need long term, but this seems to work reasonably well.~

~With this PR, the optimization stage of TPCh Q1 is still slightly slower than it was before #395, but the overall runtime is back to something reasonable (down from roughly 26s to 10s on my machine)~


dask_expr/_core.py
```diff
@@ -279,6 +280,11 @@ def simplify_once(self, dependents: defaultdict):
         expr:
             output expression
         """
+        # Check if we've already simplified for these dependents
+        key = _tokenize_deterministic(sorted(dependents.keys()))
+        if key in self._simplified:
+            return self._simplified[key]
```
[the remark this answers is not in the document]
rjzamora: Interesting. Yes, it certainly speed things up a lot for me, but I only tried Q1.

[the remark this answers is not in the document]
fjetter: yeah, sorry. I had a bug in my cache :)

fjetter:
After profiling this I am confused why this kind of cache would help us anything. Everything that lights up the profile is essentially a computation of `_meta`. Some of them can be optimized (see also https://github.com/dask-contrib/dask-expr/issues/796#issuecomment-1906058812) but the fundamental problem is that we're generating a lot of intermediate expressions and we have to compute `_meta` for almost all of them which is unfortunately not always trivial.

How is this cache helping with this? I do indeed see fewer intermediates but I don't understand how this changes anything about the performance. I ran a test where I keep the cache but never access the element

```diff
diff --git a/dask_expr/_core.py b/dask_expr/_core.py
index 3388694..86ae404 100644
--- a/dask_expr/_core.py
+++ b/dask_expr/_core.py
@@ -282,8 +282,8 @@ class Expr:
         """
         # Check if we've already simplified for these dependents
         key = _tokenize_deterministic(sorted(dependents.keys()))
-        if key in self._simplified:
-            return self._simplified[key]
+        # if key in self._simplified:
+        #     return self._simplified[key]

         expr = self

@@ -315,6 +315,8 @@ class Expr:
                 if isinstance(operand, Expr):
                     new = operand.simplify_once(dependents=dependents)
                     if new._name != operand._name:
+                        if key in self._simplified:
+                            print("Already simplified but changed")
                         changed = True
                 else:
                     new = operand
``` 

and indeed... I get a lot of "Already simplified but changed" prints so even though the expression already passed once through this, there is still more work to do apparently. This cache prohibits that work which is why it's so much faster. However, will this yield the same results?

rjzamora:
@fjetter - Thanks for looking into this.

>and indeed... I get a lot of "Already simplified but changed" prints so even though the expression already passed once through this, there is still more work to do apparently. This cache prohibits that work which is why it's so much faster. However, will this yield the same results?

Right, my initial assumption here was that the new "dependents-informed" optimization loop was simply re-generating expressions that it didn't actually *need* to regenerate, and was therefore regenerating `_meta` that it had already generated before.

I'll admit that I didn't make a solid attempt to poke holes in my quick/simple "fix" yet. However, my assumption is that if we have our `dependents` graph has the same nodes, and our expression is the same, then we have already performed that "simplify" logic before. If something about the dependencies changed, then the node names would have changed.

fjetter:
I think this cache must only be populated if the expr is indeed identical in the end, so

```diff
diff --git a/dask_expr/_core.py b/dask_expr/_core.py
index 3388694..b3f70a7 100644
--- a/dask_expr/_core.py
+++ b/dask_expr/_core.py
@@ -324,8 +324,8 @@ class Expr:
                 expr = type(expr)(*new_operands)

             break
-
-        self._simplified[key] = expr  # Cache the result
+        if expr is self:
+            self._simplified[key] = expr  # Cache the result
         return expr

     def simplify(self) -> Expr:
```

with this I still get almost 3k cache hits

you could also check the name but you'd have to check if the expression is not modified. I think an equivalent would be 

```python
if expr._name == key:
    self._simplified[key] = expr
```

but we can't populate the cache immediately.

fjetter:
I tried the caching approach in https://github.com/dask-contrib/dask-expr/pull/798 as well and all changes combined makes everything pleasantly fast.
Over there, I am using an attribute that's set on the expression that remembers whether an expression is fully simplified or not. This feels less invasive than a mapping on every instance

mrocklin:
🎉

rjzamora:
>Over there, I am using an attribute that's set on the expression that remembers whether an expression is fully simplified or not. This feels less invasive than a mapping on every instance

I do like the feel of that better, but tests are failing. I think you really do need to keep track of whether the expression was simplified for the specific `dependents` in question (not just that the expression was simplified before and it didn't change).

rjzamora:
I know this PR wasn't the best solution in its original form. However, I think it should be a reasonable solution for https://github.com/dask-contrib/dask-expr/issues/796 in its revised form. The optimization time becomes negligible (<0.5s) for Q1 on my machine (a ~10x improvement).

dask_expr/_core.py
```diff
@@ -332,7 +341,8 @@ def simplify(self) -> Expr:
         expr = self
         while True:
             dependents = collect_dependents(expr)
-            new = expr.simplify_once(dependents=dependents)
+            simplified = {}
+            new = expr.simplify_once(dependents=dependents, simplified=simplified)
```
phofl, line 345: Can you pass the dict in directly? I want to avoid that someone takes this and forgets to clear it on every pass down the line. We avoid this if we don't have a variable

dask_expr/io/_delayed.py
```diff
@@ -23,6 +23,7 @@ class _DelayedExpr(Expr):
     _parameters = ["obj"]
 
     def __init__(self, obj):
+        self._simplified = {}
```
phofl, line 26: This confuses me, can you elaborate?

phofl (commented):
Yeah I intended to do the same, this makes sense performance wise

My other PR is still worthwhile btw, would appreciate if you could take a look there #842

dask_expr/io/_delayed.py
rjzamora: Oops - this was leftover from an earlier experiment. I'll remove it. Thanks!

dask_expr/_core.py
rjzamora: Yeah, good call. We can definitely do that.

rjzamora:
Thanks for the review @phofl !

>My other PR is still worthwhile btw, would appreciate if you could take a look there https://github.com/dask-contrib/dask-expr/pull/842

Yes, I agree that's worthwhile as well. I'll definitely take a look.

phofl (approved):

phofl:
thx

PR #798, opened 2024-01-23
Expr as singleton

fjetter:
Closes https://github.com/dask-contrib/dask-expr/issues/796
Closes https://github.com/dask-contrib/dask-expr/issues/800

1. the attribute checkup in `Expr.rewrite` is expensive since it checks for an element in a list. This https://github.com/dask-contrib/dask-expr/commit/cae4df19343a91c4ca26c18e163509fd684fe7f3 commit removes that check and simply ensures the two rewrites that are currently used exists. I think this would be a sensible change but a proper PR should do this in a clean fashion. This removes all issues with the rewrite stack in my testing
2. This one is a little more controversial. I'm ensuring that the `Expr` objects are in fact singletons by implementing a `__new__` method. This ensures that there can always ever by one `Expr` with a given `_name` (that is based on tokenization)

A couple of observations

1. The performance of optimize improves by roughly an order of magnitude down from 10-12s to 1.12s on my machine.
2. There are two tests that fail with this. I strongly suspect that this is due to improper tokenization since I haven't finished https://github.com/dask/dask/issues/10799 yet. I will use this test case to root out another broken token...
3. 1.12s is about as good as it gets with this but this still feels way too slow for such a small example... (I'll post more detailed measurements shortly)

### DIFF
--- a/dask_expr/_core.py
+++ b/dask_expr/_core.py
@@ -47,6 +47,18 @@ class Expr:
         Expr._instances[_name] = inst
         return inst
 
+    def _tune_down(self):
+        return None
+
+    def _tune_up(self, parent):
+        return None
+
+    def _cull_down(self):
+        return None
+
+    def _cull_up(self, parent):
+        return None
+
     def __str__(self):
         s = ", ".join(
             str(param) + "=" + str(operand)
@@ -211,28 +223,26 @@ class Expr:
             _continue = False
 
             # Rewrite this node
-            if down_name in expr.__dir__():
-                out = getattr(expr, down_name)()
+            out = getattr(expr, down_name)()
+            if out is None:
+                out = expr
+            if not isinstance(out, Expr):
+                return out
+            if out._name != expr._name:
+                expr = out
+                continue
+
+            # Allow children to rewrite their parents
+            for child in expr.dependencies():
+                out = getattr(child, up_name)(expr)
                 if out is None:
                     out = expr
                 if not isinstance(out, Expr):
                     return out
-                if out._name != expr._name:
+                if out is not expr and out._name != expr._name:
                     expr = out
-                    continue
-
-            # Allow children to rewrite their parents
-            for child in expr.dependencies():
-                if up_name in child.__dir__():
-                    out = getattr(child, up_name)(expr)
-                    if out is None:
-                        out = expr
-                    if not isinstance(out, Expr):
-                        return out
-                    if out is not expr and out._name != expr._name:
-                        expr = out
-                        _continue = True
-                        break
+                    _continue = True
+                    break
 
             if _continue:
                 continue
--- a/dask_expr/_core.py
+++ b/dask_expr/_core.py
@@ -36,7 +36,10 @@ class Expr:
             try:
                 operands.append(kwargs.pop(parameter))
             except KeyError:
-                operands.append(cls._defaults[parameter])
+                default = cls._defaults[parameter]
+                if callable(default):
+                    default = default()
+                operands.append(default)
         assert not kwargs, kwargs
         inst = object.__new__(cls)
         inst.operands = [_unpack_collections(o) for o in operands]
--- a/dask_expr/_core.py
+++ b/dask_expr/_core.py
@@ -28,17 +28,24 @@ def _unpack_collections(o):
 class Expr:
     _parameters = []
     _defaults = {}
+    _instances = weakref.WeakValueDictionary()
 
-    def __init__(self, *args, **kwargs):
+    def __new__(cls, *args, **kwargs):
         operands = list(args)
-        for parameter in type(self)._parameters[len(operands) :]:
+        for parameter in cls._parameters[len(operands) :]:
             try:
                 operands.append(kwargs.pop(parameter))
             except KeyError:
-                operands.append(type(self)._defaults[parameter])
+                operands.append(cls._defaults[parameter])
         assert not kwargs, kwargs
-        operands = [_unpack_collections(o) for o in operands]
-        self.operands = operands
+        inst = object.__new__(cls)
+        inst.operands = [_unpack_collections(o) for o in operands]
+        _name = inst._name
+        if _name in Expr._instances:
+            return Expr._instances[_name]
+
+        Expr._instances[_name] = inst
+        return inst
 
     def __str__(self):
         s = ", ".join(
@@ -129,7 +136,7 @@ class Expr:
 
     def dependencies(self):
         # Dependencies are `Expr` operands only
-        return [operand for operand in self.operands if isinstance(operand, Expr)]
+        return [operand for operand in operands if isinstance(operand, Expr)]
 
     def _task(self, index: int):
         """The task for the i'th partition
--- a/dask_expr/_core.py
+++ b/dask_expr/_core.py
@@ -151,7 +151,7 @@ class Expr:
 
     def dependencies(self):
         # Dependencies are `Expr` operands only
-        return [operand for operand in operands if isinstance(operand, Expr)]
+        return [operand for operand in self.operands if isinstance(operand, Expr)]
 
     def _task(self, index: int):
         """The task for the i'th partition
--- a/dask_expr/_core.py
+++ b/dask_expr/_core.py
@@ -36,10 +36,7 @@ class Expr:
             try:
                 operands.append(kwargs.pop(parameter))
             except KeyError:
-                default = cls._defaults[parameter]
-                if callable(default):
-                    default = default()
-                operands.append(default)
+                operands.append(cls._defaults[parameter])
         assert not kwargs, kwargs
         inst = object.__new__(cls)
         inst.operands = [_unpack_collections(o) for o in operands]
--- a/dask_expr/io/io.py
+++ b/dask_expr/io/io.py
@@ -36,7 +36,7 @@ class FromGraph(IO):
     conversion from legacy dataframes.
     """
 
-    _parameters = ["layer", "_meta", "divisions", "_name"]
+    _parameters = ["layer", "_meta", "divisions", "_dask_keys"]
 
     @property
     def _meta(self):
@@ -45,12 +45,13 @@ class FromGraph(IO):
     def _divisions(self):
         return self.operand("divisions")
 
-    @property
-    def _name(self):
-        return self.operand("_name")
-
     def _layer(self):
-        return dict(self.operand("layer"))
+        dsk = dict(self.operand("layer"))
+        # The name may not actually match the layers name therefore rewrite this
+        # using an alias
+        for part, k in enumerate(self.operand("_dask_keys")):
+            dsk[(self._name, part)] = k
+        return dsk
 
 
 class BlockwiseIO(Blockwise, IO):
--- a/dask_expr/io/io.py
+++ b/dask_expr/io/io.py
@@ -36,7 +36,7 @@ class FromGraph(IO):
     conversion from legacy dataframes.
     """
 
-    _parameters = ["layer", "_meta", "divisions", "_dask_keys"]
+    _parameters = ["layer", "_meta", "divisions", "keys", "name_prefix"]
 
     @property
     def _meta(self):
@@ -45,11 +45,17 @@ class FromGraph(IO):
     def _divisions(self):
         return self.operand("divisions")
 
+    @functools.cached_property
+    def _name(self):
+        return (
+            self.operand("name_prefix") + "-" + _tokenize_deterministic(*self.operands)
+        )
+
     def _layer(self):
         dsk = dict(self.operand("layer"))
         # The name may not actually match the layers name therefore rewrite this
         # using an alias
-        for part, k in enumerate(self.operand("_dask_keys")):
+        for part, k in enumerate(self.operand("keys")):
             dsk[(self._name, part)] = k
         return dsk
 
--- a/dask_expr/io/parquet.py
+++ b/dask_expr/io/parquet.py
@@ -1,7 +1,6 @@
 from __future__ import annotations
 
 import contextlib
-import functools
 import itertools
 import operator
 import warnings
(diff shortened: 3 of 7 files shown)


dask_expr/_core.py
```diff
@@ -28,23 +28,45 @@ def _unpack_collections(o):
...
             except KeyError:
-                operands.append(type(self)._defaults[parameter])
+                operands.append(cls._defaults[parameter])
         assert not kwargs, kwargs
         operands = [_unpack_collections(o) for o in operands]
-        self.operands = operands
-        if self._required_attribute:
-            dep = next(iter(self.dependencies()))._meta
-            if not hasattr(dep, self._required_attribute):
+        _dependencies = [operand for operand in operands if isinstance(operand, Expr)]
```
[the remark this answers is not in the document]
rjzamora: Agree that we shouldn't need to "attach" _dependencies to the Expr class here, but that information is being used to check `_required_attribute` below anyway.

I don't think we really *need* the `_required_attribute` check anymore if we update the `Expr.__getattr__` to avoid throwing an `AttributeError` when **any** metadata calculation (`._meta`, `._meta_chunk`, etc.) fails. I'll submit a simple PR to strip out the `_required_attribute` stuff.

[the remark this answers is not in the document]
fjetter: I was just lazy and don't really care what we do with `_dependencies`. Personally, I consider the cached property an indirection and a little more difficult to grasp but I don't have strong opinions here. We have to compute it either way at this point so it's just a matter of code organization.

Getting rid of the `_required_attribute` check would of course simplify this quite a bit

fjetter:
Pushed a version of the caching as I proposed in https://github.com/dask-contrib/dask-expr/pull/797#issuecomment-1906552923 but I'm using an attribute instead of a list with the key. With this, the TPC Q1 optimization is down to 200ms

dask_expr/_core.py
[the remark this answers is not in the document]
rjzamora: This PR removes `_required_atttribute`: https://github.com/dask-contrib/dask-expr/pull/799

(not that removing that attribute solves anything here - just nice to remove unnecessary code)

dask_expr/_core.py
```diff
@@ -279,9 +300,12 @@ def simplify_once(self, dependents: defaultdict):
         expr:
             output expression
         """
+        if self.fully_simplified:
+            return self
```
rjzamora, line 304: So, if this exact expression has every been simplified and it wasn't changed, then we return early here, correct? Even if the last time it was simplified, a different `dependents` argument was used?

I don't have any reason to believe this is "wrong" right now - Just want to make sure I understand. Is it possible for `dependents` to change whether or not a specific expression gets changed?

fjetter: This isn't true if any dependents changed unless you mutate the expression in place which I hope we're not doing anywhere. This is only set if the object already passed through simplify_once without any modifications

fjetter: I'll remove this again. The singleton is the important change and I don't want to get stuck on this caching issue

fjetter: I get why this exists but I'm confused why this has to be provided externally. Doesn't matter right now. This is a problem for another issue

fjetter:
Well, the singleton approach as suggested here will clash pretty hard with the way caching is implemented on the `ReadParquet` expression.

In essence, the tokenization of the expression class does not reliably tie the expression to the dataset to be read (the path is not sufficient, a modified timestamp or etag should be part of this as well). If we reuse instances, this will fail hard as soon as the dataset is mutated.
To a lesser extend, the same is of course true for all IO layers but it is particularly bad since the cached properties are also backed by a global cache for the case of parquet.

rjzamora:
>In essence, the tokenization of the expression class does not reliably tie the expression to the dataset to be read (the path is not sufficient, a modified timestamp or etag should be part of this as well). If we reuse instances, this will fail hard as soon as the dataset is mutated.

Ah, interesting. The **global** caching component of `ReadParquet` is actually not a problem at all. The problem is entirely caused by in the cached `_dataset_info` property. The global cache is always cleared when `to_parquet` is called, but the objects local `_dataset_info`  cache is **not** cleared. (sorry if this is obvious to you - just sharing my understanding)

dask_expr/io/parquet.py
```diff
@@ -536,13 +535,20 @@ def _dataset_info(self):
...
+            files_for_checksum = [self.path + fs.sep + "_metadata"]
+        else:
+            files_for_checksum = dataset_info["ds"].files
+
+        for file in files_for_checksum:
+            # The checksum / file info is usually already cached by the fsspec
+            # FileSystem dir_cache since this info was already asked for in
+            # _collect_dataset_info
+            checksum.append(fs.checksum(file))
+        dataset_info["checksum"] = tokenize(checksum)
```
fjetter, line 551: To deal with the cache consistency problem described in https://github.com/dask-contrib/dask-expr/issues/800 I am calculating a checksum here. For s3 this falls back to using the ETag provided in the listdir response. This should not add any overhead since this stuff is already cached by fsspec.
We're either taking the checksum of the metadata file or of all files that we iterate over. At this point this listdir operation is already done so the checksum identifies every dataset uniquely. Since adding this checksum to the dataset_info, this also guarantees that the cache for the plan is invalidated if the dataset changes.

dask_expr/io/parquet.py
```diff
@@ -432,6 +416,7 @@ class ReadParquet(PartitionsFiltered, BlockwiseIO):
         "kwargs": None,
         "_partitions": None,
         "_series": False,
+        "_dataset_info_cache": list,
```
fjetter, line 419: The new concept here is that I am moving off a global cache. The dataset_info is always calculated whenever a user calls `read_parquet(`foo`)` and will therefore always receive an accurate representation of the dataset at the time this is called.
This dataset_info is cached in this paramter. I am choosing a list as a container but this could be anything. I could also just set the operand and mutate the expression in place.

The benefit of using a paramter for this cache is that the cache will naturally propagate to all derived instances, e.g. whenever we rewrite the expression using `Expr.substitute_parameters`. This allows us to maintain the cache during optimization and it ties the lifetime of the cache to the lifetime of the expression ancestry removing any need for us to invalidate the cache ever.

dask_expr/io/parquet.py
```diff
@@ -474,7 +459,21 @@ def _simplify_up(self, parent, dependents):
                 return Literal(sum(_lengths))
 
     @cached_property
+    def _name(self):
+        return (
+            funcname(type(self)).lower()
+            + "-"
+            + _tokenize_deterministic(self.checksum, *self.operands)
+        )
```
fjetter, line 467: this checksum is part of the `_name` allowing us to differentiate expressions that point to modified states of the dataset. It also allows us to reuse already cached "plans / divisions" if the dataset did not change which is the most common case

dask_expr/io/parquet.py
```diff
@@ -474,7 +459,21 @@ def _simplify_up(self, parent, dependents):
...
+
+    @property
+    def checksum(self):
+        return self._dataset_info["checksum"]
+
+    @property
     def _dataset_info(self):
+        if rv := self.operand("_dataset_info_cache"):
+            return rv[0]
```
fjetter, line 476: When the `ReadParquet` expression is initialized first during a `read_parquet` this cache is empty and we'll fetch the dataset_info essentially during construction time of the expression object.

Subsequent expressions that are derived which are inheriting the cache will just access this making the `__new__` call instantaneous.

fjetter:
ok, this one requires two minor fixes in dask/dask, see https://github.com/dask/dask/pull/10880

fjetter:
Well, the performance is better but still not where I'd like it to be

image

image


We have a very consistent improvement with this PR but some of the queries appear to still scratch the 1s mark. 


The reason for this is that the current branch uses a weakvalue dictionary to determine whether an object already exists but it looks like the expr objects are going out of scope too quickly such that the instances are indeed being recreated frequently. Looking at Q1, the main branch creates about 20k new expression instances and with this branch I can half this. However, over the lifetime of the optimization there are only about 500 unique expressions. 
Switching from a weakvalue dictionary to something more persistent is currently also not easily possible since the [`collect_dependents`](https://github.com/dask-contrib/dask-expr/blob/fc92ed923ece233dfa6077f0feaf8f542632b536/dask_expr/_core.py#L683-L696) mechanism does rely on the weakref semantics.

phofl:
We should be able to remove the weak refs if we keep track of the "number of references" ourselves on top of this PR

fjetter:
If we combine this approach with https://github.com/dask-contrib/dask-expr/pull/831 we'll get down further

image


(higher is better here)
image


The one thing that is not budging is Q21. Looking into that, it appears that there is an insane amount of substitution happening during blockwise fusion
Screenshot 2024-02-01 at 17 55 54

phofl:
21 is because the 3 filters blow up the number of expressions, similar to what assign did in merge in the past. I am almost done with the filter pushdown in merges which will remove those problems entirely

fjetter:
With both changes, Q1 optimize is about as fast as it can get at around 500ms. Further speedup is possible by improving tokenization, simplifying the attribute lookup for Expr and speeding up convergence of the optimizer overall to require fewer steps. Even the meta computation is rather negligible at this point. I think 500ms is fine (for now) but I'd like to understand the substitude thing a little better.

fjetter:
> 21 is because the 3 filters blow up the number of expressions, similar to what assign did in merge in the past. I am almost done with the filter pushdown in merges which will remove those problems entirely

There are around ~700 unique expressions in that query during optimization. That's not that much (compared to ~500 for Q1)

dask_expr/io/io.py
```diff
@@ -36,7 +36,7 @@ class FromGraph(IO):
     conversion from legacy dataframes.
     """
 
-    _parameters = ["layer", "_meta", "divisions", "_name"]
+    _parameters = ["layer", "_meta", "divisions", "keys", "name_prefix"]
```
fjetter, line 39: The `FromGraph` thing got a little odd. For context, this thing is used in two ways

1. As a wrapper for futures when persist is called
2. As a wrapper for legacy dataframes

The previous implementation accepted the `_name` as an input argument. For the persisted dataframe, this was the name of the original expression. For the wrapped one, it is the name of the legacy dataframe.

The problem is that the legacy dataframe does not have the same strict uniqueness guarantees as the expressions do so setting duplicate names is very easy. In fact, our tests where doing just that! (and still are). This caused the `Expr.__new__` to deduplicate and effectively ignore the second dataframe... oops.

For persist it is also a little odd since if the `FromGraph` expression is inheriting the exact name of it's ancestor, there exist now two expressions of a different type with the same name. This is odd.


Effectively, with this chosen singleton implementation, setting the `_name` explicitly instead of calculating it using a hash is a cardinal sin and will cause all sorts of weird things to happen.

Now, this can be fixed but that has a rather big caveat. If I have to redefine the name of the expression, I actually also have to rewrite the graph! Many places in dask-expr are (imo incorrectly) assuming that output key names of a dataframe layer/expression are universally built as `(df._name, i)` and are hard coding this when implementing their own layer (instead of iterating over `i`, iterating over `df.__dask_keys__()` would maintain the abstraction). This rewrite adds effectively another layer of keys. In reality this is really ugly since when computing something on top of a persisted dataframe, there will always be this dummy key in between.

Alternatively, I could make the singleton deduplication type aware to give the `FromGraph` thing an excuse to overwrite the name. However, if we truly stick with singletons that are based on the name, I would prefer the name to actually be unique which required all implementations to stop hard coding keys of another expression/dataframe and iterate properly over the `__dask_keys__`

dask_expr/io/io.py
```diff
@@ -45,12 +45,19 @@ def _meta(self):
...
+        )
 
     def _layer(self):
-        return dict(self.operand("layer"))
+        dsk = dict(self.operand("layer"))
+        # The name may not actually match the layers name therefore rewrite this
+        # using an alias
+        for part, k in enumerate(self.operand("keys")):
+            dsk[(self._name, part)] = k
+        return dsk
```
fjetter, line 60: The intuitive fix for me would've been to overwrite `__dask_keys__` instead to point to the appropriate keys but as I explained above, the implementation of `Expr.__dask_keys__` is hard coded in many places.

dask_expr/io/io.py
phofl: > Now, this can be fixed but that has a rather big caveat. If I have to redefine the name of the expression, I actually also have to rewrite the graph! Many places in dask-expr are (imo incorrectly) assuming that output key names of a dataframe layer/expression are universally built as (df._name, i) and are hard coding this when implementing their own layer (instead of iterating over i, iterating over df.__dask_keys__() would maintain the abstraction). This rewrite adds effectively another layer of keys. In reality this is really ugly since when computing something on top of a persisted dataframe, there will always be this dummy key in between.

This is a good point, we should fix this instead of relying on names

fjetter:
I opened https://github.com/dask-contrib/dask-expr/pull/838 which points the dask/dask tests to https://github.com/dask/dask/pull/10880 such that we can see if CI is happy

dask_expr/io/io.py
fjetter: I will investigate if it is possible to just use `__dask_keys__` everywhere but I'd prefer doing this in a follow up

phofl: Yes totally agree, this should definitely be a follow up (also doesn't have to be you, I could pick this up as well)

fjetter: I got started on this, see

https://github.com/dask-contrib/dask-expr/commit/c3c01ed57d7986d1ad8c89f311a2bce2e453d08c

That commit replaces all occurrences that match `\((\w+(_\w+)?)(?<!self)\._name,` but something appears to be still missing. It is possible and not as ugly as I thought it would be

fjetter:
CI when pointing to dask/dask with the fixes https://github.com/dask-contrib/dask-expr/pull/838 is green

phofl:
thx @fjetter

PR #799, opened 2024-01-23
Remove ``Expr._required_attribute``

rjzamora:
The `_required_attribute` check was meant to avoid infinite `AttributeError` recursion. However, we can also avoid this recursion by expanding the current "_meta" check in `Expr.__getattr__` to throw a `RuntimeError` when the key is anything that starts with `"_meta"` (e.g. "_meta_chunk").

This PR also consolidates the redundant `dask_expr._expr.Expr.__getattr__` and `dask_expr._core.Expr.__getattr__` logic in once place. However, I can revert that change if there were two nearly-identical definitions for a reason.

### DIFF
--- a/dask_expr/_core.py
+++ b/dask_expr/_core.py
@@ -39,18 +39,6 @@ class Expr:
         assert not kwargs, kwargs
         operands = [_unpack_collections(o) for o in operands]
         self.operands = operands
-        if self._required_attribute:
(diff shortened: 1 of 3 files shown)


rjzamora:
Note that CI failures are unrelated: https://github.com/dask-contrib/dask-expr/issues/801

phofl (approved):

phofl:
thx

Issue #801, opened 2024-01-24
CI failing for ``test_repartition_quantiles``


rjzamora:
`dask_expr/tests/test_quantiles.py::test_repartition_quantiles` is failing locally for me in `main`. The failure is also showing up in CI:

```
E               AssertionError: Series.index are different
E               
E               Series.index values are different (16.66667 %)
E               [left]:  Index([0.0, 0.2, 0.4, 0.6000000000000001, 0.8, 1.0], dtype='float64')
E               [right]: Index([0.0, 0.2, 0.4, 0.6, 0.8, 1.0], dtype='float64')
```

rjzamora:
Looks like the cause is numpy:

```python
np.linspace(0, 1, 6)[3]  # Returns: 0.6000000000000001
```

rjzamora:
This is probably not new in numpy. This may only be failing now, because the `check_exact` default changed from `False` to `True` in `assert_frame_equal` for pandas 2.2 (https://pandas.pydata.org/docs/reference/api/pandas.testing.assert_frame_equal.html#pandas-testing-assert-frame-equal)

PR #802, opened 2024-01-24
Fix `test_repartition_quantiles` to unblock CI

rjzamora:
Pandas 2.2 changed the default for `check_exact` in `assert_frame_equal`. This PR sets it to `check_exact=False` to revert to the original behavior.

Closes https://github.com/dask-contrib/dask-expr/issues/801


crusaderky:
Superseded by #806

PR #808, opened 2024-01-25
Split out dask/dask tests in CI

crusaderky:
The dask/dask tests never got to run if there's a failure in the dask-expr tests.
This PR also reduces end-to-end runtime by 3 minutes.

### DIFF
--- a/.github/workflows/dask_test.yaml
+++ b/.github/workflows/dask_test.yaml
@@ -24,7 +24,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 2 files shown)


.github/workflows/dask_test.yaml
```diff
@@ -0,0 +1,57 @@
+name: dask/dask tests
```
hendrikmakait, line 1: nit: 
```suggestion
name: Tests / dask/dask
```
would IMO fit better into the existing hierarchy.

hendrikmakait (approved):
Thanks, @crusaderky!

PR #811, opened 2024-01-26
Fix `sort_values` simplification for `{head|tail}`

hendrikmakait:
For single partitions, `sort_values` supports sorting by multiple columns with each column sorted ascending or descending independently. When these values are not always the same, this cannot be simplified to `smallest`. Since the simplification does not save us much with a single partition, I avoid simplifying at all in this case.

### DIFF
--- a/dask_expr/tests/test_shuffle.py
+++ b/dask_expr/tests/test_shuffle.py
@@ -380,6 +380,14 @@ def test_sort_head_nlargest(df, pdf):
     b = df.nsmallest(10, columns=["x"]).expr
     assert a.optimize()._name == b.optimize()._name
 
+    a = df.sort_values(["x"], ascending=[False]).head(10, compute=False).expr
+    b = df.nlargest(10, columns=["x"]).expr
+    assert a.optimize()._name == b.optimize()._name
+
+    a = df.sort_values(["x"], ascending=[True]).head(10, compute=False).expr
+    b = df.nsmallest(10, columns=["x"]).expr
+    assert a.optimize()._name == b.optimize()._name
+
     a = df.sort_values(["x", "y"], ascending=[False]).head(10, compute=False).expr
     b = df.nlargest(10, columns=["x", "y"]).expr
     assert a.optimize()._name == b.optimize()._name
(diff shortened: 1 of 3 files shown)


dask_expr/tests/test_shuffle.py
```diff
@@ -372,6 +372,67 @@ def test_sort_head_nlargest(df):
...
+    assert a.optimize()._name == b.optimize()._name
+
+    a = df.sort_values(["x", "y"], ascending=[True]).head(10, compute=False).expr
+    b = df.nsmallest(10, columns=["x", "y"]).expr
+    assert a.optimize()._name == b.optimize()._name
+
+    with pytest.raises(
+        NotImplementedError,
+        match="Dask currently only supports a single boolean for ascending",
+    ):
```
fjetter, line 394: funny you should mention that... https://github.com/dask/dask/pull/10864

dask_expr/tests/test_shuffle.py
```diff
@@ -372,6 +372,83 @@ def test_sort_head_nlargest(df):
...
+    result = df.sort_values("x", ascending=[True]).head(10)
+    expected = pdf.sort_values("x", ascending=[True]).head(10)
+    assert_eq(result, expected, sort_results=False)
+
+    result = df.sort_values(["x"], ascending=[True]).head(10)
+    expected = pdf.sort_values(["x"], ascending=[True]).head(10)
+    assert_eq(result, expected, sort_results=False)
+
+    result = df.sort_values(["x"], ascending=[False]).head(10)
+    expected = pdf.sort_values(["x"], ascending=[False]).head(10)
```
hendrikmakait, line 435: This case also fails in the current implementation, which is why https://github.com/coiled/benchmarks/pull/1290/files#diff-ed9173e9cde3ccc498e25c8485d7c2060041c6b7cb0228f26e1c16603a3f3b10L478 works as a fix.

hendrikmakait:
CI failures are unrelated.

phofl (approved):

phofl:
thx

PR #815, opened 2024-01-29
Always install from main to circumvent caching

phofl:


hendrikmakait (approved):
Thanks, @phofl! This makes a lot of sense for me given the frequent adjustments we make to the Dask packages to accommodate for `dask-expr`.

PR #816, opened 2024-01-29
Actually circumvent the cache

crusaderky:
Follow-up to #815, which didn't actually achieve its intent.
See e.g. https://github.com/dask-contrib/dask-expr/actions/runs/7695594548/job/20968817200


hendrikmakait (approved):
I'm mildly confused that this works while the other approach doesn't, but the CI logs look good, so LGTM. Thanks, @crusaderky!

crusaderky:
...still doesn't work. I'm so confused.

crusaderky:
Misread the log. Fixed by https://github.com/dask/dask/pull/10870

PR #821, opened 2024-01-29
Avoid recursion error if filter condition can optimize itself

### LINKED ISSUE #807: RecurrsionError in TPC-H Query 9
https://github.com/coiled/benchmarks/pull/1275

This same query works with dask.dataframe. (2024.1.0)

Edit, just going to refactor that query which avoids the error in https://github.com/coiled/benchmarks/pull/1277, but still interesting the first implementation causes it in dask-expr but not in dask.dataframe.

phofl:
closes #807


hendrikmakait (approved):
Thanks, @phofl!

PR #823, opened 2024-01-30
Add docstrings

phofl:
Not sure if this is worth a proper review, 99% is copy paste from dask/dask for now


milesgranger (approved):
Thanks @phofl, know it's not the most exciting of work. :)

phofl:
> Thanks @phofl, know it's not the most exciting of work. :)

so true!

PR #824, opened 2024-01-31
Always avoid shuffling for `sort_values(...).{head|tail(...)`

hendrikmakait:


dask_expr/_reductions.py
```diff
@@ -1259,6 +1259,41 @@ def _nlargest_slow(df, columns, n):
...
 
+def _nfirst(df, columns, n, ascending):
+    return df.sort_values(by=columns, ascending=ascending).head(n)
+
+
+def _nlast(df, columns, n, ascending):
+    return df.sort_values(by=columns, ascending=ascending).tail(n)
+
+
+class NFirst(NLargest):
```
hendrikmakait, line 1270: I don't like having `NFirst`, `NLast`, `NLargestSlow` and `NSmallestSlow` that all do basically the same. I also don't like that current simplification hierarchy. We should refactor this in a follow-up.

dask_expr/_reductions.py
```diff
@@ -1259,6 +1259,41 @@ def _nlargest_slow(df, columns, n):
...
+    _defaults = {"n": 5, "_columns": None, "ascending": None, "split_every": None}
+    reduction_chunk = _nlast
+    reduction_aggregate = _nlast
+
+    @property
+    def chunk_kwargs(self):
+        return {"ascending": self.ascending, **super().chunk_kwargs}
+
+
+class NLargestSlow(NLargest):
```
phofl, line 1292: This got duplicated somehow

dask_expr/_shuffle.py
```diff
@@ -1042,6 +1044,10 @@ def _simplify_up(self, parent, dependents):
                     return NLargest(self.frame, n=parent.n, _columns=self.by)
                 else:
                     return NLargestSlow(self.frame, n=parent.n, _columns=self.by)
+            else:
```
phofl, line 1047: I think you can just remove Lines 1037-1046 or does this fail tests? Same for NLast Also ok as a follow up

dask_expr/tests/test_shuffle.py
```diff
@@ -447,23 +448,36 @@ def test_sort_tail_nsmallest(df, pdf):
 @pytest.mark.parametrize("npartitions", [1, 3])
 def test_sort_values_conflicting_ascending_head_tail(pdf, npartitions):
     df = from_pandas(pdf, npartitions=npartitions)
+
+    a = df.sort_values(by=["x", "y"], ascending=[True, False]).head(10, compute=False)
```
phofl, line 452: one more thing here: Could you assert that we didn't do a division calculation after optimise?

```
    divisions_lru.data = OrderedDict()
```

at the beginning of your test

and then 

```
assert len(divisions_lru) == 0
```

after optimise

hendrikmakait: Done!

phofl (approved):

phofl:
thx!

Issue #302, opened 2023-09-14
Optimizing Filters and Merges


mrocklin:
I was writing down some tests to help me think through swapping filters and merges

```python
def test_filter_merge():
    a = pd.DataFrame({"a": range(5), "b": range(5), "c": range(5)})
    b = pd.DataFrame({"c": [0, 2, 4, 6, 8], "x": range(5), "y": range(5)})

    a = dx.from_pandas(a)
    b = dx.from_pandas(b)

    # Some simple cases
    df = a.merge(b)
    df = df[df.x > 3]
    bb = b[b.x > 3]
    expected = a.merge(bb)
    assert df.optimize()._name == expected.optimize()._name

    df = a.merge(b)
    df = df[df.b > 3]
    aa = a[a.b > 3]
    expected = aa.merge(b)
    assert df.optimize()._name == expected.optimize()._name

    # Apply to both!
    df = a.merge(b)
    df = df[df.c > 3]
    aa = a[a.c > 3]
    bb = b[b.c > 3]
    expected = aa.merge(bb)
    assert df.optimize()._name == expected.optimize()._name

    # Works with more complex expressions, and multiple columns
    df = a.merge(b)
    df = df[df.a > df.b + 1]
    aa = a[a.a > a.b + 1]
    expected = aa.merge(b)
    assert df.optimize()._name == expected.optimize()._name

    # Only apply if all columns are in the table, not if only some are
    df = a.merge(b)
    df = df[df.c > df.x + 1]
    bb = b[b.c > b.x + 1]
    expected = a.merge(bb)
    assert df.optimize()._name == expected.optimize()._name

    # Bail if we engage non-elemwise expressions in the predicates
    df = a.merge(b)
    df = df[df.x > df.y.sum()]
    bb = b[b.x > b.y.sum()]
    not_expected = a.merge(bb)
    assert df.optimize()._name != not_expected.optimize()._name
```

I think that what I came to is that we want to push a filter into one side or the other if

1.   the predicate is valid when we replace the merged table with the left or right side.  Something like if `predicate.subs({merged: left})` doesn't raise an error (after we figure out renaming of columns).
2. The predicate has only elemwise expressions (nothing like `df.x < df.y.sum()` for example)

And so maybe an implementation here relies on ...

1.  Making a proxy object for `left` that has no expression history, but does have the same `._meta`
2. Calling `substitute` and seeing if it errs
[...]

PR #827, opened 2024-01-31
Filters through merge

fjetter:
Earlier I was talking with @hendrikmakait  about https://github.com/dask-contrib/dask-expr/issues/302 and was wondering what this was all about and why it was not supported yet.

The implementation was pretty straightforward and I ended up taking the test cases from the issue and adding to it. I'm sure there are more edge cases that are not handled well.

One thing that troubles me is that the way I ended up implementing `AND` concatenated conditions is not working because the optimizer drops one of the conditions. From what I can tell so far this is because the `dependents` at this point https://github.com/dask-contrib/dask-expr/blob/b588a9e15e90e0567061664ffc01374786686e20/dask_expr/_core.py#L293-L303 are not recomputed. the earlier simplify_up steps could alter the dependents such that they are no longer valid at this point. recomputing this (or breaking) fixes this but this causes all sorts of other things to break. Primarily the stuff with combine_similar in the test name


dask_expr/_merge.py
```diff
@@ -352,6 +357,45 @@ def _lower(self):
...
+            predicate = parent.predicate
+            if isinstance(predicate, (Projection, Unaryop)):
+                predicate_cols = set(predicate.columns)
+            elif isinstance(predicate, Binop):
+                # FIXME: This doesn't work yet. The optimizer drops the outer
+                # filter
+                # if isinstance(predicate, And):
+                #     new = Filter(self, predicate.left)
+                #     new_pred = predicate.right.substitute(self, new)
+                #     return Filter(new, new_pred)
```
fjetter, line 373: Maybe, I'm just doing it wrong but this breaks everything. Maybe @phofl can tell me why this is wrong

dask_expr/_merge.py
```diff
@@ -608,12 +652,14 @@ def _layer(self) -> dict:
...
-                    if self.how != "inner"
-                    else (other, part_out),
+                        (
+                            operator.getitem,
+                            (split_name, part_out),
+                            j,
+                        )
+                        if self.how != "inner"
+                        else (other, part_out)
+                    ),
```
fjetter, line 662: this isn't intentional. no idea what went wrong here...

dask_expr/_merge.py
phofl: Yeah I figured out what's wrong, I'll explain offline though, not sure if I can communicate this well this late

phofl:
Moving the filters is still too aggressive, here is a reproducer, it needs to take dependents into account before doing anything

```
df = pd.DataFrame({"a": [1, 2, 3], "b": 1})
ddf = from_pandas(df, npartitions=2)
df2 = pd.DataFrame({"a": [2, 3, 4], "c": 1})
ddf2 = from_pandas(df2, npartitions=2)
merged = ddf.merge(ddf2, on="a", how="left")
rhs = merged.c.sum()
result = merged[merged.a > 1].assign(c=rhs)
result.simplify().pprint()
```

this pushes too aggressively, which was the main reason that I didn't add this while we had combine_similar, that would have broken my brain and was one of the motivators that I wanted to get rid of it

```
 Assign: c=
  Merge: how='left' left_on='a' right_on='a'
    Filter:
      FromPandas: frame='<pandas>' npartitions=2
      GT: right=1
        Projection: columns='a'
          FromPandas: frame='<pandas>' npartitions=2
    Filter:
      FromPandas: frame='<pandas>' npartitions=2
      GT: right=1
        Projection: columns='a'
          FromPandas: frame='<pandas>' npartitions=2
  Sum:
    Projection: columns='c'
      Merge: how='left' left_on='a' right_on='a'
        Projection: columns=['a']
          FromPandas: frame='<pandas>' npartitions=2
        FromPandas: frame='<pandas>' npartitions=2
```

dask_expr/tests/test_collection.py
```diff
@@ -1971,15 +1971,10 @@ def test_contains(df):
 
 
 def test_filter_pushdown(df, pdf):
+    # TODO: This isn't technically correct, round shouldn't let filters through
     indexer = df.x > 5
-    result = df.replace(1, 5)[indexer].optimize(fuse=False)
-    expected = df[indexer].replace(1, 5)
-    assert result._name == expected._name
-
-    # Don't do anything here
```
phofl, line 1979: This part never made sense

PR #831, opened 2024-02-01
[DNM/RFC] Track dependents without weakrefs

fjetter:
@phofl this is an implementation that would track the "expression graph" on the instances themselves without needing weakref semantics.

In a nutshell, the expressions all track a dictionary that maps the names of the expresions to the names of their dependencies. When one needs the dependents, this can be reverted.

I haven't tested performance on this thing yet. The instance creation is a little more expensive now but I would be surprised if this would be a significant overhead. The dependents graph inversion should be the same cost as the weakref implementation we have right now.

Tests are currently failing because the DelayedExpr thingy is a little broken but the rest passes. I'll look at performance next and will see if this can be easily combined with https://github.com/dask-contrib/dask-expr/pull/798 This Draft PR is just an FYI


dask_expr/_core.py
```diff
@@ -11,34 +10,59 @@
...
+        # Graph instances is a mapping name -> Expr instance
+        # Graph itself is a mapping of dependencies mapping names to a set of names
+        self._graph_instances = merge(_graph_instances, *_subgraph_instances)
+        self._graph = merge(*_subgraphs)
+        self._graph[name] = children
+        # Probably a bad idea to have a self ref
+        self._graph_instances[name] = self
+
+    def __hash__(self):
+        raise TypeError("Don't!")
```
fjetter, line 65: "Funny" story: The above implementation is maintaining two dicts. One that maps names and dependencies and one that just maps names to the actual objects. Initially I just tried to use the Expr objects themselves in the mappings using a mapping from `Expr -> set[Expr}`. That triggered funny recursion errors and it took me a while to understand that...

Whenever there is a hash collision, the implementation of set falls back to use `__eq__` to compare the new object with the one that is stored under a given hash, i.e. `object.__eq__(old, new)`. However, this doesn't evaluate to a bool but defines another expression... 💥

phofl, line 65:
```suggestion
        raise TypeError("Expr objects can't be used in sets or dicts or similar, use the _name instead")
```

I went down the same rabbit hole in the past, lets make this very clear

phofl:
I have to take a closer look, but I like the solution. When I tried the same I only stored the dependents on the expression which caused me trouble while recreating the mapping. Simply building the mapping continuously and storing it on the expression should circumvent the issues I ran into. 

Thanks for taking a stab at this

dask_expr/_core.py
mrocklin: Yeah, definite `__eq__` is pretty hazardous in Python.  It would be reasonable, I think, to drop `Expr.__eq__` and use explicit `Eq(a, b)` calls instead, and then rely on `Frame.__eq__` for user comfort.

mrocklin: Or rather, use whatever class is defined in `collections.py` to hold the `__eq__` method

dask_expr/tests/test_shuffle.py
```diff
@@ -640,7 +641,7 @@ def test_set_index_sort_values_one_partition(pdf):
 
 def test_set_index_triggers_calc_when_accessing_divisions(pdf, df):
     divisions_lru.data = OrderedDict()
-    query = df.set_index("x")
+    query = df.fillna(random.randint(1, 100)).set_index("x")
```
phofl, line 644: This one is funny, the previous expression was cached, so we were never calculating divisions with this change. Using a random number here avoid this

phofl:
Fix for failing tests is here: https://github.com/dask-contrib/dask-expr/pull/880

Will merge into this branch

PR #832, opened 2024-02-01
Improve filter pushdown

phofl:
This makes Filter pushdown work more reliably and a bit more aggressive, the previous approach was pretty defensive


hendrikmakait (approved):
Thanks, @phofl!

PR #839, opened 2024-02-02
Make filter pushdown work for more complicated filters

phofl:
We were making decisions differently when a filter would modify its child, this wasn't reflected in th Filter class, which in turn let Projections through even though it shouldn't. Moving this logic to a method that can be overridden solves this inconsistency

Having multiple filters after each other will block projections if we kept it the way it was. Squashing them together again when we can't move them anymore solves this nicely

### DIFF
--- a/dask_expr/_expr.py
+++ b/dask_expr/_expr.py
@@ -1850,7 +1850,11 @@ class Filter(Blockwise):
     def _simplify_up(self, parent, dependents):
         if isinstance(parent, Filter) and not isinstance(self.frame, Filter):
             if not self.frame._filter_passthrough_available(self, dependents):
-                # collect filters if we can't move them anymore
+                # We want to collect filters again when we can't move them
+                # anymore. Otherwise, a chain of Filters might nlock Projections
+                # We start with the first Filter, e.g. if my child isn't a filter
+                # anymore. Filters above that don't know if the first Filter can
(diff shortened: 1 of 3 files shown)


dask_expr/_expr.py
```diff
@@ -1846,10 +1848,14 @@ class Filter(Blockwise):
     operation = operator.getitem
 
     def _simplify_up(self, parent, dependents):
+        if isinstance(parent, Filter) and not isinstance(self.frame, Filter):
```
hendrikmakait, line 1851: Why is the `and not isinstance(self.frame, Filter)` necessary here?

phofl:
I want to make sure that the first non filter starts collecting them,
because the last filter doesn’t know if the first one can still be pushed
down further

Hendrik Makait ***@***.***> schrieb am Fr. 2. Feb. 2024 um
17:16:

dask_expr/_expr.py
hendrikmakait: The explanation makes sense, a comment might be useful.

hendrikmakait (approved):
Thanks, @phofl! Consider the request for an additional comment non-blocking.

dask_expr/_expr.py
phofl: added a comment

PR #842, opened 2024-02-02
Improve column projection performance

phofl:
cc @rjzamora 

this cuts runtime in half on my machine, not super fancy but just avoids as much work as possible without changing the logic

### DIFF
--- a/dask_expr/_expr.py
+++ b/dask_expr/_expr.py
@@ -1758,9 +1758,6 @@ class Assign(Elemwise):
 
     def _simplify_down(self):
         if isinstance(self.frame, Assign):
-            # if len(self.vals) == 1 and isinstance(self.vals[0], Callable):
-            #     # UDFs can't move around
-            #     return
             if self._check_for_previously_created_column(self.frame):
                 # don't squash if we are using a column that was previously created
                 return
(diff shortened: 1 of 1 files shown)


dask_expr/_expr.py
```diff
@@ -3190,21 +3187,22 @@ def determine_column_projection(expr, parent, dependents, additional_columns=Non
...
 
+    seen = set()
     for p in parents:
-        if len(p.columns) > 0:
-            column_union.append(p._projection_columns)
+        if p._name in seen:
+            continue
+        seen.add(p._name)
+
+        column_union.extend(p._projection_columns)
```
rjzamora, line 3196: Is it worthwhile to have this in a `if len(p.columns) > 0:` block like before?

dask_expr/_expr.py
```diff
@@ -3190,21 +3187,22 @@ def determine_column_projection(expr, parent, dependents, additional_columns=Non
...
     # We can end up with MultiIndex columns from groupby ops, needs to be
     # accounted for in the sort
-    flattened_columns = set(flatten(column_union, container=list))
+    flattened_columns = set(column_union)
     try:
-        column_union = sorted(
-            flattened_columns,
-            key=lambda x: x[0] if isinstance(x, tuple) else x or MinType(),
-        )
+        column_union = sorted(flattened_columns)
```
rjzamora, line 3205: Did something change here that I'm not seeing, or was the `key=` never necessary?

rjzamora (commented):
Thanks @phofl ! Seems good to me, just some minor questions.

dask_expr/_expr.py
phofl: The logic changed a little bit, tuples now end up in the except branch. Since tuples are rare, it’s better to slow them down than to slow the default path down with the check

PR #844, opened 2024-02-05
Block predicate pushdown if they violate the join direction

phofl:
sits on the other pr

cc @hendrikmakait


milesgranger (approved):
:+1: Thanks!

PR #850, opened 2024-02-06
Avoid squashing filters where the condition depends on the previous one

phofl:
this addresses query 22


hendrikmakait (approved):
Thanks, @phofl!

Issue #840, opened 2024-02-02
Error when calling ``var`` or ``std`` on the group key in a ``SeriesGroupBy``


rjzamora:
```python
import dask_expr as dd
import pandas as pd

pdf = pd.DataFrame({"a": [1, 2] * 2, "b": [0] * 4})
df = dd.from_pandas(pdf , 2)

df.groupby("a")["a"].var()
```
Traceback:
```
---------------------------------------------------------------------------
IndexError                                Traceback (most recent call last)
Cell In[1], line 7
      4 pdf = pd.DataFrame({"a": [1, 2] * 2, "b": [0] * 4})
      5 df = dd.from_pandas(pdf , 2)
----> 7 df.groupby("a")["a"].var()

File ~/.../dask_expr/_groupby.py:1744, in GroupBy.var(self, ddof, split_every, split_out, numeric_only, shuffle_method)
   1725 result = new_collection(
   1726     Var(
   1727         self.obj.expr,
   (...)
   1737     )
   1738 )
   1739 if (
   1740     isinstance(self.obj, Series)
   1741     or is_scalar(self._slice)
   1742     and self._slice is not None
   1743 ):
-> 1744     result = result[result.columns[0]]
   1745 return result

File /.../site-packages/pandas/core/indexes/base.py:5366, in Index.__getitem__(self, key)
   5363 if is_integer(key) or is_float(key):
   5364     # GH#44051 exclude bool, which would return a 2d ndarray
   5365     key = com.cast_scalar_indexer(key)
-> 5366     return getitem(key)
   5368 if isinstance(key, slice):
   5369     # This case is separated from the conditional above to avoid
   5370     # pessimization com.is_bool_indexer and ndim checks.
   5371     return self._getitem_slice(key)

IndexError: index 0 is out of bounds for axis 0 with size 0
```
Note that I also see the same error for `std`.

phofl:
This doesn't work in dask/dask either and I am not really convinced that it should work, the result is kind of pointless since it will always be 0

rjzamora:
>This doesn't work in dask/dask either and I am not really convinced that it should work, the result is kind of pointless since it will always be 0

Yeah, I agree that the calculation seems pointless. I'm not sure if this is worth fixing at all, or if it's best to raise an error when someone tries to aggregate on the grouping key.

phofl:
Some aggregations can make kind of sense and work, it's just std and var that don't work

rjzamora:
[...]

PR #851, opened 2024-02-06
Raise ``NotImplementedError`` for unsupported ``SeriesGroupBy.var/std``

rjzamora:
Closes https://github.com/dask-contrib/dask-expr/issues/840

This change is not entirely necessary given that the "correct" result is trivial. However, I think it's best to provide a bit of information to the user since it's easy.

### DIFF
--- a/dask_expr/_groupby.py
+++ b/dask_expr/_groupby.py
@@ -1745,7 +1745,7 @@ class GroupBy:
         ):
             if len(result.columns) < 1:
                 raise NotImplementedError(
-                    "Cannnot call `SeriesGroupBy.var` on the key column. "
+                    "Cannot call `SeriesGroupBy.var` on the key column. "
                     "Please use `aggregate` if you really need to do this."
                 )
             result = result[result.columns[0]]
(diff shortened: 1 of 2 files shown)


dask_expr/_groupby.py
```diff
@@ -1743,6 +1743,11 @@ def var(
             or is_scalar(self._slice)
             and self._slice is not None
         ):
+            if len(result.columns) < 1:
+                raise NotImplementedError(
+                    "Cannnot call `SeriesGroupBy.var` on the key column. "
```
hendrikmakait, line 1748:
```suggestion
                    "Cannot call `SeriesGroupBy.var` on the key column. "
```

dask_expr/_groupby.py
```diff
@@ -1778,6 +1783,11 @@ def std(
             or is_scalar(self._slice)
             and self._slice is not None
         ):
+            if len(result.columns) < 1:
+                raise NotImplementedError(
+                    "Cannnot call `SeriesGroupBy.std` on the key column. "
```
hendrikmakait, line 1788:
```suggestion
                    "Cannot call `SeriesGroupBy.std` on the key column. "
```

dask_expr/_groupby.py
rjzamora: Wow. Thanks for catching that!

phofl:
thx

PR #853, opened 2024-02-06
Add list conversion to ``determine_column_projection``

rjzamora:
After https://github.com/dask-contrib/dask-expr/pull/842, I'm seeing a new test failure in `dask_cudf`. It seems that we need to convert the output of `_sort_mixed` to `list`.

### DIFF
--- a/dask_expr/tests/test_collection.py
+++ b/dask_expr/tests/test_collection.py
@@ -1373,13 +1373,17 @@ def test_columns_assignment():
 
 
 def test_parent_mixed_column_assignment():
-    pdf = pd.DataFrame(list(range(100)))
+    pdf = pd.DataFrame(list(range(20)))
     df = from_pandas(pdf, npartitions=4)
-    df["fold"] = 0
-    df["fold"] = df["fold"].map_partitions(
-        lambda x: pd.Series(np.random.randint(0, 4, len(x)))
+    df["a"] = 0
+    df["a"] = df["a"].map_partitions(
+        lambda x: pd.Series(
+            np.full(len(x), 1, dtype="int"),
+            index=x.index,
+        )
     )
-    df.simplify()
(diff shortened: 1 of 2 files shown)


dask_expr/tests/test_collection.py
```diff
@@ -1372,6 +1372,16 @@ def test_columns_assignment():
...
 
 
+def test_parent_mixed_column_assignment():
+    pdf = pd.DataFrame(list(range(100)))
+    df = from_pandas(pdf, npartitions=4)
+    df["fold"] = 0
+    df["fold"] = df["fold"].map_partitions(
+        lambda x: pd.Series(np.random.randint(0, 4, len(x)))
+    )
+    df.simplify()
```
rjzamora, line 1382: It does seem like the logic needed to reproduce the failure is a bit weird. I'm open to other ideas :)

phofl: This potentially converts from a series to a df if you do the conversion
Unconditionally, that needs a bit of care before merging

rjzamora: Are you saying there is an issue with the test or with the `tolist()` addition? (sorry for being slow)

phofl: Sorry I kind of messed this up in my head, your solution is good

dask_expr/tests/test_collection.py
```diff
@@ -1372,6 +1372,16 @@ def test_columns_assignment():
     assert_eq(df, ddf)
 
 
+def test_parent_mixed_column_assignment():
```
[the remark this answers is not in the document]
rjzamora: Ah good point. The test did "compute" before, but most of the result was NaN. I revised the logic a bit to represent something more realistic.

phofl:
thx

PR #857, opened 2024-02-08
Move all divisions checks into the lower stage

phofl:
sorry for the weird diff, but I had to move the classes a little bit

### DIFF
--- a/dask_expr/_expr.py
+++ b/dask_expr/_expr.py
@@ -316,8 +316,10 @@ class Expr(core.Expr):
         return Clip(self, lower=lower, upper=upper, axis=axis)
 
     def combine_first(self, other):
-        frame, other = self._align_divisions(other, axis=0)
-        return CombineFirst(frame, other=other)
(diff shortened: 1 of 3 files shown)


milesgranger (approved):
LGTM! Thanks @phofl!

dask_expr/_expr.py
```diff
@@ -3216,6 +3082,216 @@ def _kwargs(self) -> dict:
...
+    _expr_cls = Assign
+
+    def _simplify_up(self, parent, dependents):
+        # TODO: de-duplicate
+        if isinstance(parent, Projection):
+            columns = determine_column_projection(self, parent, dependents)
+            if not isinstance(columns, list):
+                columns = [columns]
+
+            cols = set(columns) - set(self.keys)
```
rjzamora, line 3184: This class has no `keys` property, so this PR causes a new failure in dask-cudf (I'll try to make a simple repro).

PR #858, opened 2024-02-08
Simplify sort_values simplify_up step

phofl:
This became unnecessary complex, the performance differences is 50 vs 100ms on a df with 1 million rows and 10 columns, which doesn't matter for us


hendrikmakait (approved):
Thanks, @phofl!

PR #859, opened 2024-02-08
Add how=leftsemi to merge

phofl:


dask_expr/tests/test_merge.py
```diff
@@ -852,3 +852,28 @@ def test_merge_scalar_comparison():
     expected = pdf.merge(pdf2)
     expected = expected[expected.a > pdf.a.mean()]
     assert_eq(result, expected, check_index=False)
+
+
+def test_merge_leftsemi():
+    pdf1 = pd.DataFrame({"a": [1, 2, 3, 4, 5, 6, 1, 2, 3], "b": 1})
+    pdf2 = pd.DataFrame({"a": [1, 2, 2, 4, 4], "c": 1})
```
hendrikmakait, line 859: We should also test the case where the right DF contains a value that's not in the left one.
```suggestion
    pdf2 = pd.DataFrame({"a": [1, 2, 2, 4, 4, 7], "c": 1})
```

phofl: added another value

PR #861, opened 2024-02-08
Minor change to ``ShuffleReduce._lower``

rjzamora:
I'm working on the "cudf" backend for dask-expr, and I noticed that a small piece of logic in `ShuffleReduce._lower` only works when the `DecomposableGroupbyAggregation` uses the index to store the group keys. The dask-cudf approach currently keeps everything in the columns (partly for historical reasons).

I know there are several workarounds I can pursue. However, the code "just works" with the small tweak in this PR. (and I'm pretty sure the change makes sense either way)


phofl (approved):

phofl:
thx

PR #863, opened 2024-02-09
Fix `leftsemi` merge on multi-letter columns

hendrikmakait:


phofl (approved):
Yikes this happened in the past as well, thx

PR #868, opened 2024-02-13
Move parquet specific tests into dedicated file

fjetter:
I'll probably start adding specific parquet tests such that it makes sense to have a dedicated test file. There are still parquet tests in `test_io.py` but I believe those are pretty universal optimization tests and most are either using parquet as a generic IO interface or are parametrizing over parquet and csv.

I think these are parquet only functionality

### DIFF
--- a/dask_expr/io/tests/test_parquet.py
+++ b/dask_expr/io/tests/test_parquet.py
@@ -1,0 +1,146 @@
+import os
+
+import pandas as pd
+import pytest
+from dask.dataframe.utils import assert_eq
+
+from dask_expr import from_pandas, optimize, read_parquet
+from dask_expr._expr import Lengths, Literal
+from dask_expr._reductions import Len
+from dask_expr.io import ReadParquet
+
+
+def _make_file(dir, df=None):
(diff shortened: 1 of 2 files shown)


dask_expr/io/tests/test_parquet.py
```diff
@@ -0,0 +1,146 @@
...
+from dask.dataframe.utils import assert_eq
+
+from dask_expr import from_pandas, optimize, read_parquet
+from dask_expr._expr import Lengths, Literal
+from dask_expr._reductions import Len
+from dask_expr.io import ReadParquet
+
+
+def _make_file(dir, df=None):
+    fn = os.path.join(str(dir), f"myfile.{format}")
```
phofl, line 14:
```suggestion
    fn = os.path.join(str(dir), f"myfile.parquet")
```

fjetter:
sorry, there are a couple of errors, I'll get to it soon

fjetter:
obsolete with https://github.com/dask-contrib/dask-expr/pull/882

PR #874, opened 2024-02-14
Fix some meta calculations and rest index projections

phofl:
Nonempty meta objects can actually screw us because pandas aligns under the hood


milesgranger (approved):
LGTM! :+1:

PR #877, opened 2024-02-14
Allow `columns` argument to be mandatory in `from_map`

fjetter:
`from_map` is performing some magic if the function that is being provided has a `columns` kwarg. However, if that columns `kwarg` is not optional, the magic breaks down. Status on main is that it raises this confusing error message

```
func() missing 1 required positional argument: 'columns'
```

this change adds functionality that allows the function to use a mandatory `columns` argument

Stumbled over this while trying to use it in https://github.com/dask-contrib/dask-deltatable/pull/69

Note: If the `meta` is not provided we have to raise because we don't have a way to generate the meta ourselves but in this case we're raising now a clean exception

### DIFF
--- a/dask_expr/_collection.py
+++ b/dask_expr/_collection.py
@@ -4890,14 +4890,15 @@ def from_map(
 
     # Check if `func` supports column projection
     allow_projection = False
+    columns_arg_required = False
     if param := inspect.signature(func).parameters.get("columns", None):
         allow_projection = True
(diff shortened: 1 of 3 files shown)


fjetter:
Lovely, another `RecursionError` 🔍

phofl (approved):

phofl:
thx

PR #878, opened 2024-02-14
Fix divisions for head

phofl:
closes https://github.com/coiled/benchmarks/issues/1393

cc @jrbourbeau


jrbourbeau (commented):
Nice, thanks @phofl

PR #880, opened 2024-02-14
Fix reads from local dir that changes directory

phofl:
Not adding this to the token makes us thing this is the same operation, silently reading the wrong data

The example might be a little bit too constructed, but if we start caching more this can happen in different scenarios

### DIFF
--- a/dask_expr/io/parquet.py
+++ b/dask_expr/io/parquet.py
@@ -426,6 +426,7 @@ class ReadParquet(PartitionsFiltered, BlockwiseIO):
         "filesystem",
         "engine",
         "kwargs",
+        "_cwd",  # needed for tokenization
         "_partitions",
         "_series",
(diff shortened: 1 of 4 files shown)


dask_expr/io/parquet.py
```diff
@@ -449,6 +450,7 @@ class ReadParquet(PartitionsFiltered, BlockwiseIO):
         "_partitions": None,
         "_series": False,
         "_dataset_info_cache": None,
+        "_cwd": None,
```
fjetter, line 453: I'm pretty sure the parquet reader does not need this since we're actually computing a checksum for all the files. if that doesn't work, we should make sure that the checksum is reliable and we may want the same/similar mechanism for csv instead of relying on CWD. CWD feels odd when working with remote storages

PR #881, opened 2024-02-15
Make column projections stricter

phofl:
No clue how to actually test those changes since they don't make a difference in how we optimises at the moment, but they would if we switch to breadth first


hendrikmakait:
> No clue how to actually test those changes since they don't make a difference in how we optimises at the moment, but they would if we switch to breadth first

Do you have any intuition about how things would change if we switched to BDF? If so, I'd suggest adding one or two tests that protect against those anticipated problems.

phofl:
Existing tests cover those cases, that's how I found those things in the first place

hendrikmakait (approved):
Thanks, @phofl! If we have sufficient test coverage already, this is good enough for me.

Issue #856, opened 2024-02-08
Parquet checksum calculation horribly slow with arrow FileSystem wrapper


fjetter:
We're calculating a checksum of the parquet file here https://github.com/dask-contrib/dask-expr/blob/d1c4ed1da01642df6802881d62998a5a81519b85/dask_expr/io/parquet.py#L550-L567 that relies on the fsspec dir_cache. This is implemented for the ordinary S3FS filesystem but not for the arrow wrapper. There are possibly other implementations where this fails as well.

Without this cache, this checksum is not feasible.

This was introduced in https://github.com/dask-contrib/dask-expr/pull/798

fjetter:
FWIW I'm currently running with the patch below to run tests

```diff
@@ -559,12 +677,12 @@ class ReadParquet(PartitionsFiltered, BlockwiseIO):
         else:
             files_for_checksum = dataset_info["ds"].files

-        for file in files_for_checksum:
-            # The checksum / file info is usually already cached by the fsspec
-            # FileSystem dir_cache since this info was already asked for in
-            # _collect_dataset_info
-            checksum.append(fs.checksum(file))
-        dataset_info["checksum"] = tokenize(checksum)
+        # for file in files_for_checksum:
+        #     # The checksum / file info is usually already cached by the fsspec
+        #     # FileSystem dir_cache since this info was already asked for in
+        #     # _collect_dataset_info
+        #     checksum.append(fs.checksum(file))
+        dataset_info["checksum"] = tokenize(files_for_checksum)

         # Infer meta, accounting for index and columns arguments.
         meta = self.engine._create_dd_meta(dataset_info)
```

PR #882, opened 2024-02-16
Parquet reader using Pyarrow FileSystem

fjetter:
Closes https://github.com/dask-contrib/dask-expr/issues/856

Primarily what isn't implemented is any sort of statistics collection, compaction or row group splitting.

The approach I'm taking here is pretty straight forward. I'm taking the pyarrow filesystem and am implementing the bare essentials to instantiate a parquet dataset and get the minimal metadata from the bucket and files (e.g. all paths, the schema). The expression then effectively tracks the `Fragment` objects that are being created by pyarrow. The Fragments expose all the API we truly care about
They can...
- load parquet metadata (not implemented yet)
- read the parquet file as pandas
- accept low level configuration options
- split themselves by row group (not implemented yet)

I think the most notable change besides relying directly and explicitly on the pyarrow filesystem (which is backed by C++ threads, i.e. no GIL/event loop blockage) is that I am configuring the `pre_buffer` cache options explicitly (this may actually also be possible with the existing implementation but I'm a little lost about how kwargs are passed through).

This is not really a cache but rathe a control for how many concurrent IO requests we are posting and how large/small those are allowed to become. The config options are briefly documented here https://github.com/apache/arrow/blob/a61f4af724cd06c3a9b4abd20491345997e532c0/python/pyarrow/io.pxi#L2139

The TLDR is primarily that the default `hole_size_limit` is a couple of orders of magnitude too small such that we get incredibly fragmented read requests.
Arrow actually ships with a utility that allows one to calculate appropriate config options easily [CacheOptions.from_network_metrics](https://github.com/apache/arrow/blob/a03d957b5b8d0425f9d5b6c98b6ee1efa56a1248/python/pyarrow/io.pxi#L2211-L2238)

Docs are in the C++ code here https://github.com/apache/arrow/blob/a61f4af724cd06c3a9b4abd20491345997e532c0/cpp/src/arrow/io/caching.cc#L49C28-L103 which also describes the math they are using and I think that makes sense. There is more documentation [here](https://github.com/apache/arrow/blob/a61f4af724cd06c3a9b4abd20491345997e532c0/cpp/src/arrow/io/caching.h#L44-L49) about the `lazy` and `prefetch_limit` options.

I started measuring S3 latencies for our benchmarking bucket (`coiled-runtime-ci`)

image

Depending on how hot/cold the bucket is, the FirstByteLatency fluctuates between 15ms and 60ms wheres total request latency is somewhere between 30ms and 200ms. Plugging this stuff into the equation and assuming 50MiB/s per connection, we'll get

```python
import pyarrow as pa
copts = pa.CacheOptions.from_network_metrics(
    100, # I think we should take TotalRequestLatency as a guiding point
    50,  # 50MiB/s per connection
)
print(f"hole_size_limit={format_bytes(copts.hole_size_limit)}")
print(f"range_size_limit={format_bytes(copts.range_size_limit)}")
```

```
hole_size_limit=5.00 MiB
range_size_limit=45.00 MiB
```

So, while the range limit is fine you'll notice that the hole size is `MiB` whereas the default is `kiB`. Effectively, we're crippling ourselves with column projections because we're introducing more latency to the requests than we're saving on data transfer.

---

How does this translate to benchmarks?

Well, I haven't run many yet and am about to setup a full tpch run. So far I looked at two queries

# Q1

On the pyarrow generated dataset, this change isn't doing that much. The files are reasonably large and the fragmentation doesn't hurt us that badly. Also, with current main only about 40-50% of this query is actual IO. With this branch I can reduce IO to 30% which effectively shaves off about 10% total runtime

On the DuckDB dataset which contains many small rowgroups (i.e. chances of producing holes are very large and we'll end up with many fragmented reads) this change without further tuning boosts us by 25% easily

# Q14

This one was pointed out to be to suffer from P2P and parquet cross interference. The query is about 40% faster on this branch using the arrow filesystem. I suspect this is fsspec vs arrow more than the caching but I haven't investigated.

I'll post full results once available.

### DIFF
--- a/dask_expr/_collection.py
+++ b/dask_expr/_collection.py
@@ -4641,7 +4641,7 @@ def read_parquet(
             raise NotImplementedError(
                 "split_row_groups is not supported when using the pyarrow filesystem."
             )
-        if blocksize != "default":
+        if blocksize is not None and blocksize != "default":
             raise NotImplementedError(
                 "blocksize is not supported when using the pyarrow filesystem."
             )
--- a/dask_expr/_collection.py
+++ b/dask_expr/_collection.py
@@ -4606,7 +4606,11 @@ def read_parquet(
     engine=None,
     **kwargs,
 ):
-    from dask_expr.io.parquet import ReadParquet, _set_parquet_engine
+    from dask_expr.io.parquet import (
+        ReadParquetFSSpec,
+        ReadParquetPyarrowFS,
+        _set_parquet_engine,
+    )
 
     if not isinstance(path, str):
         path = stringify_path(path)
@@ -4618,9 +4622,59 @@ def read_parquet(
             col, op, val = filter
             if op == "in" and not isinstance(val, (set, list, tuple)):
                 raise TypeError("Value of 'in' filter must be a list, set or tuple.")
+    from pyarrow import fs as pa_fs
+
+    if (
+        isinstance(filesystem, pa_fs.FileSystem)
+        or isinstance(filesystem, str)
+        and filesystem.lower() in ("arrow", "pyarrow")
+    ):
+        if calculate_divisions:
+            raise NotImplementedError(
+                "calculate_divisions is not supported when using the pyarrow filesystem."
+            )
+        if metadata_task_size is not None:
+            raise NotImplementedError(
+                "metadata_task_size is not supported when using the pyarrow filesystem."
+            )
+        if split_row_groups != "infer":
+            raise NotImplementedError(
+                "split_row_groups is not supported when using the pyarrow filesystem."
+            )
+        if blocksize != "default":
+            raise NotImplementedError(
+                "blocksize is not supported when using the pyarrow filesystem."
+            )
+        if aggregate_files is not None:
+            raise NotImplementedError(
+                "aggregate_files is not supported when using the pyarrow filesystem."
+            )
+        if parquet_file_extension != (".parq", ".parquet", ".pq"):
+            raise NotImplementedError(
+                "parquet_file_extension is not supported when using the pyarrow filesystem."
+            )
+        if engine is not None:
+            raise NotImplementedError(
+                "engine is not supported when using the pyarrow filesystem."
+            )
+
+        return new_collection(
+            ReadParquetPyarrowFS(
+                path,
+                columns=_convert_to_list(columns),
+                filters=filters,
+                categories=categories,
+                index=index,
+                storage_options=storage_options,
+                filesystem=filesystem,
+                ignore_metadata_file=ignore_metadata_file,
+                kwargs=kwargs,
+                _series=isinstance(columns, str),
+            )
+        )
 
     return new_collection(
-        ReadParquet(
+        ReadParquetFSSpec(
             path,
             columns=_convert_to_list(columns),
             filters=filters,
--- a/dask_expr/_collection.py
+++ b/dask_expr/_collection.py
@@ -4651,7 +4651,7 @@ def read_parquet(
         or isinstance(filesystem, str)
         and filesystem.lower() in ("arrow", "pyarrow")
     ):
-        if parse_version(pa.__version__) <= parse_version("15.0.0"):
+        if parse_version(pa.__version__) < parse_version("15.0.0"):
             raise ValueError(
                 "pyarrow>=15.0.0 is required to use the pyarrow filesystem."
             )
--- a/dask_expr/_collection.py
+++ b/dask_expr/_collection.py
@@ -12,6 +12,7 @@ import dask.array as da
 import dask.dataframe.methods as methods
 import numpy as np
 import pandas as pd
+import pyarrow as pa
 from dask import compute, delayed
 from dask.array import Array
 from dask.base import DaskMethodsMixin, is_dask_collection, named_schedulers
@@ -57,10 +58,12 @@ from dask.utils import (
 )
 from dask.widgets import get_template
 from fsspec.utils import stringify_path
+from packaging.version import parse as parse_version
 from pandas import CategoricalDtype
 from pandas.api.types import is_bool_dtype, is_datetime64_any_dtype, is_numeric_dtype
 from pandas.api.types import is_scalar as pd_is_scalar
 from pandas.api.types import is_timedelta64_dtype
+from pyarrow import fs as pa_fs
 from tlz import first
 
 import dask_expr._backends  # noqa: F401
@@ -4642,13 +4645,16 @@ def read_parquet(
             col, op, val = filter
             if op == "in" and not isinstance(val, (set, list, tuple)):
                 raise TypeError("Value of 'in' filter must be a list, set or tuple.")
-    from pyarrow import fs as pa_fs
 
     if (
         isinstance(filesystem, pa_fs.FileSystem)
         or isinstance(filesystem, str)
         and filesystem.lower() in ("arrow", "pyarrow")
     ):
+        if parse_version(pa.__version__) <= parse_version("15.0.0"):
+            raise ValueError(
+                "pyarrow>=15.0.0 is required to use the pyarrow filesystem."
+            )
         if calculate_divisions:
             raise NotImplementedError(
                 "calculate_divisions is not supported when using the pyarrow filesystem."
--- a/dask_expr/io/parquet.py
+++ b/dask_expr/io/parquet.py
@@ -50,7 +50,6 @@ from dask_expr._expr import (
     Projection,
     are_co_aligned,
     determine_column_projection,
-    is_filter_pushdown_available,
 )
 from dask_expr._reductions import Len
 from dask_expr._util import _convert_to_list, _tokenize_deterministic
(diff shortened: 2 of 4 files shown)


fjetter:
The hole size of `~5MiB` also matches with the default block size we chose in [`plateau`](https://github.com/data-engineering-collective/plateau/blob/e0f96d1c1664a57aeb7c36ac21006942e1f9685a/plateau/serialization/_parquet.py#L221) (4MiB there). The buffer there works a little differently but the block size there correlates pretty well with the hole size in this implementation. The buffer there worked great for me in the past.

fjetter:
> Note that there may be some value to concurrent reads to S3, even with appropriately sized reads. Anecdotally I've found value in 3x number of connections per core/task.

Of course. The `range_size_limit` parameter is there to ensure ranges are not fused too aggressively. I think there is still some experimentation necessary but these parameters are making a big difference already.

fjetter:
Preliminary results (I got hard failures for Q18 and Q20, therefore no data). This is A/B main w/ fsspec and this branch

image

mrocklin:
Nice.

dask_expr/io/parquet.py
```diff
@@ -379,25 +397,70 @@ def to_parquet(
...
+        for type_converter in type_mappers:
+            converted_type = type_converter(pyarrow_dtype)
+            if converted_type is not None:
+                return converted_type
+
+    if len(type_mappers) > 0:
+        return default_types_mapper
+
+
+class ReadParquetPyarrowFS(PartitionsFiltered, BlockwiseIO):
```
fjetter, line 454: I just noticed that this is not inheriting from `ReadParquet`. I intended this to inherit since there are many places in the code path where we're checking for `ReadParquet`. So.... this didn't use any of the optimizations defined in  `ReadParquet.simplify_up`. I guess those are not very relevant for TPCH? 🤔

fjetter:
I added filter pushdown to this based on https://github.com/dask-contrib/dask-expr/pull/886 with a couple of fixes as discussed with @phofl . Looks like some tests broke so I may remove this again to merge it.

I have good and bad news

First the good news: Pushing filters down appears to be worth it
(this is this branch with filters pushdown compared against no filter pushdown, i.e. 717ad0b872e0ae4c655b1671511a6af9d6084be9 vs 227557d20fee6bdceb58f6659741d5d55152c829

This makes sense since the filter pushdown can prune entire files. This is happening for Q1 for example. I will investigate in a follow up whether it is worthwhile pushing filters down to IO level. I assume it is on the DuckDB dataset with the granular rowgroups but not on the pyarrow dataset. TBD

image


Bad news: Rebasing seems to have broken something and I caught a regression. I'll need to investigate this (left 227557d20fee6bdceb58f6659741d5d55152c829 w/out filters; right 717ad0b872e0ae4c655b1671511a6af9d6084be9 with filters)

image

fjetter:
Oh well, it turns out that some of the performance kick came from dropped data :( https://github.com/dask/distributed/pull/8520 I could track this down for query_2 and I suspect this is also the case for the other regressions we're seeing above. Now that this is settled, I can dig into why they are slower. That shouldn't be the case either way.

fjetter:
I'd like to move forward with merging this. I have a couple more things to try and modify but would like to enter an ordinary review mode asap.

current benchmarks against main (with latest pandas) looks like 

image

The _regressions_ we're seeing for query2, for example, are P2P stragglers. Looking at the bar chars w/ error bars this is a little more obvious

image

image


Particularly with https://github.com/dask-contrib/dask-expr/issues/856 the pyarrow filesystem is horribly broken anyhow so this PR is a very, very clear improvement.

fjetter:
In case this is not abundantly obvious, this isn't feature complete. Particularly the parquet metadata scraping and all the goodies that come from it are not implemented yet. Unfortunately, the pyarrow Fragments loose their metadata upon serialization so we'll have to implement a similar collection layer as in dask/dask (if we choose to scrape all metadata). This is something I'd like to enter a proper review mode for.

There are two other possibly major performance changes that I would also like to separate
- Use `pyarrow` dtype backend such that the pyarrow tables are only wrapped and actually never converted to pandas. This gives us in my testing another solid 10% performance kick on top of the above (sometimes more). However, this will likely cause fallout in tests which is why a separate PR seems sensible
- Implement another concurrency layer in FusedIO. This is something I would love to not have to do, particularly since pyarrow appears to implement something similar (that's the fragment readahead foo) but I haven't found out how to tap into that, yet.

dask_expr/io/parquet.py
```diff
@@ -496,18 +552,295 @@ def _simplify_up(self, parent, dependents):
...
+            batch_size=10_000_000,
+            # batch_readahead=16,
+            # fragment_readahead=4,
+            fragment_scan_options=pa.dataset.ParquetFragmentScanOptions(
+                pre_buffer=True,
+                cache_options=pa.CacheOptions(
+                    hole_size_limit=parse_bytes("4 MiB"),
+                    range_size_limit=parse_bytes("32.00 MiB"),
+                ),
+            ),
```
fjetter, line 768: I'm open to exposing all of this via kwargs or options but I don't believe there are (m)any users that have the necessary know how and determination to fine tune this. I could be wrong, of course. I would prefer adding this once there is need for it.

dask_expr/io/parquet.py
```diff
@@ -496,18 +552,295 @@ def _simplify_up(self, parent, dependents):
...
+            # and to set this to something like inf but we have to set a finite,
+            # positive number.
+            # In the presence of row groups, the underlying array will still be
+            # chunked per rowgroup
+            batch_size=10_000_000,
+            # batch_readahead=16,
+            # fragment_readahead=4,
+            fragment_scan_options=pa.dataset.ParquetFragmentScanOptions(
+                pre_buffer=True,
+                cache_options=pa.CacheOptions(
```
rjzamora, line 764: It seems like `CacheOptions` requires pyarrow>=15, so dask-expr would need to pin pyarrow pretty aggressively to use this.

fjetter: yes, indeed

fjetter: is this a problem?

rjzamora: Well, I cannot build an environment with cudf and pyarrow>14 right now. I'm sure there are plenty of other packages that would like to use dask/dask-expr with a pyarrow version that is more than 4 weeks old.

mrocklin: All of this is already behind a big `if` check, right?  Maybe we can add the PyArrow version to that condition so that this only gets used if Arrow is pretty new.

rjzamora: Right, I should clarify that I don't think it's a bad idea to "use" it (setting appropriate data sieving properties can be huge).

dask_expr/_collection.py
```diff
@@ -4638,9 +4642,59 @@ def read_parquet(
             col, op, val = filter
             if op == "in" and not isinstance(val, (set, list, tuple)):
                 raise TypeError("Value of 'in' filter must be a list, set or tuple.")
+    from pyarrow import fs as pa_fs
+
+    if (
+        isinstance(filesystem, pa_fs.FileSystem)
```
phofl, line 4648: do you want to make those configs raise or should we rather fall back to the old implementation?

dask_expr/io/parquet.py
```diff
@@ -406,28 +436,63 @@ def to_parquet(
...
-        "ignore_metadata_file",
-        "metadata_task_size",
-        "split_row_groups",
-        "blocksize",
-        "aggregate_files",
-        "parquet_file_extension",
         "filesystem",
-        "engine",
         "kwargs",
-        "_partitions",
```
phofl, line 429: Can we remove all of those? This class doesn't seem to be used directly anymore?

dask_expr/io/parquet.py
```diff
@@ -496,18 +552,295 @@ def _simplify_up(self, parent, dependents):
...
+    _absorb_projections = True
+    _filter_passthrough = True
+
+    @cached_property
+    def fs(self):
+        fs_input = self.operand("filesystem")
+        if isinstance(fs_input, pa.fs.FileSystem):
+            return fs_input
+        else:
+            fs = pa_fs.FileSystem.from_uri(self.path)[0]
```
phofl, line 643: How expensive is the creation of the filesystem?

phofl (commented):
few comments

dask_expr/io/parquet.py
fjetter: no cost. This doesn't talk to S3 yet

dask_expr/_collection.py
fjetter: The old implementation is currently pretty broken so I don't want to fall back, see https://github.com/dask-contrib/dask-expr/issues/856

fjetter:
Given the lack of any objections and the fact that this is a dedicated code path that doesn't touch the fsspec filesystem code and the pyarrow FS code is currently broken anyhow (https://github.com/dask-contrib/dask-expr/issues/856) I'll move forward with merging

Issue #854, opened 2024-02-06
Reusing intermediate results causes memory issues


hendrikmakait:
**Problem**

Whenever we reuse intermediate results and there is a pipeline breaker (such as shuffles, joins, reductions, or groupby operations), it forces us to materialize the entire intermediate result (thus, it is breaking the pipelining that we could utilize for reuse for example between multiple element-wise operations.

This result materialization puts a hard limit on our ability to scale as I have observed in multiple TPC-H benchmark queries.

To illustrate this, run these two snippets on a cluster of your choice

**With full intermediate result materialization**
```python
from dask_expr.datasets import timeseries

from distributed import Client

if __name__ == "__main__":
    with Client() as client:
        print(client.dashboard_link)
        df = timeseries(start="2000-01-01", end="2020-12-31", freq="100ms", dtypes={"x": float})
        # To compute mean, we have to fully materialize df, and we won't free its data
        # until we have reused the chunks to compute a partial of the sum. 
        mean = df["x"].mean()
        df[df["x"] > mean].sum().compute()
```

**Without full intermediate result materialization**
```python
from dask_expr.datasets import timeseries

from distributed import Client

if __name__ == "__main__":
    with Client() as client:
        print(client.dashboard_link)
        df = timeseries(start="2000-01-01", end="2020-12-31", freq="100ms", dtypes={"x": float})
        # Compute the mean beforehand so that we don't have to keep all of `df` in memory
        mean = df["x"].mean()
        df[df["x"] > mean].sum().compute()
```

**Possible solution**

The easiest approach would be to _never_ reuse any intermediate results. This has a few downsides:
* Non-deterministic functions will lead to unexpected results
* We waste _a lot_ of computational resources on recomputations

...but it will allow us to scale. 

We can certainly get smarter about intermediate result materialization, but this will require some effort depending on how smart we want to be. (There's a body of (ongoing) research and implementations in the database world we could draw from.)

phofl:
This is concerning

fjetter:
The two code examples above are identical, aren't they?

mrocklin:
FWIW this is a long-standing issue.  https://github.com/dask/dask/issues/874

hendrikmakait:
> The two code examples above are identical, aren't they?

_Fixed, I messed up the copy+paste._

hendrikmakait:
Summary from multiple offline conversations with @phofl (and @fjetter):

[...]

PR #883, opened 2024-02-16
Add ``branch_id`` to distinguish between reusable branches and pipeline breakers

phofl:
ref #854

Couple of thoughts here:

- enabling / disabling reuse of branches seems like a completely different thing compared to what we do in simplify, so it felt wrong adding it in here and we would need more special casing on top of it. I want an easy way to opt out of this, so having a more central approach seems a lot better

- branch_id needs to be available on every expression, so that we can move this around freely without worrying if the expression supports it.

- it has to be available in __new__, because it influences how the expression tokenises. 

- I decided to bubble the expression step by step but not keep it on intermediate layers (e.g. non Reductions and non IO layers). We would have to consider the branch_id in every lower and simplify operation to avoid dropping it in between, which introduces a lot of uncertainty and mental lode which didn't seem to be worth the effort. The step by step approach gives us more flexibility how we handle other Expressions that can consume the branch_id (thinking about merges and shuffles at the moment)

This is not ready to merge, it needs more tests and support for merges and shuffles at least, maybe more. But it does what it's supposed to now

### DIFF
--- a/dask_expr/_core.py
+++ b/dask_expr/_core.py
@@ -225,7 +225,7 @@ class Expr:
 
         return {(self._name, i): self._task(i) for i in range(self.npartitions)}
 
-    def rewrite(self, kind: str):
+    def rewrite(self, kind: str, cache):
         """Rewrite an expression
 
         This leverages the ``._{kind}_down`` and ``._{kind}_up``
@@ -238,6 +238,9 @@ class Expr:
         changed:
             whether or not any change occured
         """
+        if self._name in cache:
+            return cache[self._name]
+
         expr = self
         down_name = f"_{kind}_down"
         up_name = f"_{kind}_up"
@@ -274,7 +277,8 @@ class Expr:
             changed = False
             for operand in expr.operands:
                 if isinstance(operand, Expr):
-                    new = operand.rewrite(kind=kind)
+                    new = operand.rewrite(kind=kind, cache=cache)
+                    cache[operand._name] = new
                     if new._name != operand._name:
                         changed = True
                 else:
--- a/dask_expr/_core.py
+++ b/dask_expr/_core.py
@@ -40,7 +40,7 @@ class Expr:
         if _branch_id is None and len(operands) and isinstance(operands[-1], BranchId):
             _branch_id = operands.pop(-1)
         elif _branch_id is None:
-            _branch_id = BranchId(None)
+            _branch_id = BranchId(0)
 
         for parameter in cls._parameters[len(operands) :]:
             try:
@@ -125,6 +125,10 @@ class Expr:
                     op = "<series>"
                 elif is_arraylike(op):
                     op = "<array>"
+                elif isinstance(op, BranchId):
+                    if op.branch_id == 0:
+                        continue
+                    op = f" branch_id={op.branch_id}"
                 header = self._tree_repr_argument_construction(i, op, header)
 
         lines = [header] + lines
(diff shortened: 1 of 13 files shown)


dask_expr/tests/test_groupby.py
```diff
@@ -107,6 +107,7 @@ def test_groupby_reduction_optimize(pdf, df):
     ops = [
         op for op in agg.expr.optimize(fuse=False).walk() if isinstance(op, FromPandas)
     ]
+    agg.simplify().pprint()
```
fjetter, line 110: debug print?

dask_expr/_expr.py
```diff
@@ -2758,24 +2767,37 @@ def optimize(expr: Expr, fuse: bool = True) -> Expr:
         Input expression to optimize
     fuse:
         whether or not to turn on blockwise fusion
+    common_subplan_elimination : bool
```
fjetter, line 2770: If I'm not mistaken, our default is actually a very radical form of "common subplan elimination". I think this feature is actually quite the opposite and rather something like "replicate common subplan"? I guess other engines are doing it the other way round, aren't they?

dask_expr/_core.py
```diff
@@ -30,23 +35,36 @@ class Expr:
...
         Expr._instances[_name] = inst
         return inst
 
+    @functools.cached_property
+    def argument_operands(self):
+        return self.operands[:-1]
+
+    @functools.cached_property
+    def _branch_id(self):
+        return self.operands[-1]
```
fjetter, line 66: How about just storing this as an attribute? Why does this need to be an operand?

phofl: Having it in the operands makes sure that it is used for ``_name``, which is what we mostly care about. If that is overridden and forgotten, then the whole thing won't work properly

dask_expr/tests/test_groupby.py
phofl: Oh yes, thx

dask_expr/_expr.py
phofl: I took the name from here: https://docs.pola.rs/py-polars/html/reference/lazyframe/api/polars.LazyFrame.explain.html

But I don't have strong feelings either way

fjetter: There is also https://en.wikipedia.org/wiki/Common_subexpression_elimination with a simple example =)

fjetter: Naming is hard. @hendrikmakait any thoughts?

dask_expr/_core.py
fjetter: I get that. I'm just concerned this is messing with other things since `operands` are everywhere. If it works I'm fine.

fjetter: I think to "avoid" the kind of confusion I'm thinking about you introduced argument_opearands but this may also be confusing to developers

dask_expr/_expr.py
hendrikmakait: Sorry that I'm late to the party. `dask-expr` approaches CSE from the other end, defaulting to as much CSE as possible. This makes it less intuitive. As mentioned in #896, `common_subplan_elimination=False` does not mean that we don't do any CSE but instead that we try to reduce it. I'm fine with the argument name for now that you've cleaned it up, let's see if this leads to some confusion in the future.

hendrikmakait: Regarding Common Sub**expression** Elimination vs Common Sub**plan** Elimination: I think the latter is more common for data systems and compilers/IR-based systems.

hendrikmakait: Interestingly. `polars` seems to have both sub**expression** (https://github.com/pola-rs/polars/blob/128803b237dc13d0522c22dbccae1257ae30477e/crates/polars-plan/src/logical_plan/optimizer/cse_expr.rs) as well as sub**plan** elimination (https://github.com/pola-rs/polars/blob/128803b237dc13d0522c22dbccae1257ae30477e/crates/polars-plan/src/logical_plan/optimizer/cse.rs).

I'm not _entirely_ sure what the difference is, but from what I understand `cse_expr` deals with eliminating duplication within a single expressions (`a.sum() + b.sum() + a.sum()`) and `cse` deals with what eliminating subgraphs across the entire graph.

That is, `cse_expr` corresponds to _local_ CSE and `cse` to _global_ CSE in the link Florian shared.

PR #884, opened 2024-02-16
Simplify again after lowering

phofl:
We removed this when we still had combine_similar, because simplifying again would have screwed over everything. This is not a concern anymore and we only needed 2 small changes to make this work again


hendrikmakait (approved):
Thanks, @phofl!

PR #887, opened 2024-02-20
Add shuffle if objects are not aligned and partitions are unknown in assign

### LINKED ISSUE #390: How should we treat the result of a GroupBy with a shuffle?
A relatively recent PR in dask/dask introduced a set_index after a ``GroupBy.transform`` (https://github.com/dask/dask/pull/10184), which is not ideal imo. This was done to ensure that in and output of the groupby have the same Index to make an assign possible.

This opened us up to a bunch of inconsistencies, e.g.

- transform now resets the initial index
- apply and shift also trigger a shuffle but they don't do this
- This only works if the original object has known divisions, otherwise the set_index is a waste because it wouldn't work anyway
 
```python
import dask.dataframe as dd

df = pd.DataFrame({"a": [1, 2, 3, 4, 5] * 10, "b": 1, "c": list(range(50))}, index=list(range(50, 0, -1)))
df = dd.from_pandas(df, npartitions=50, sort=False)
df["new"] = df.groupby("a").transform(lambda x: x)["c"]
```

- set_index is expensive because of the division calculation

How do we want to treat this in dask-expr?

We can be smarter, e.g. only inserting the set_index when we actually want to assign the result back to the original object, but this still leaves the set_index problems and known division problems from the original object. 

I think the way to go might be to make assign smarter so that it aligns the partitions if they aren't aligned already.

cc @fjetter @jrbourbeau

phofl:
- closes #390

### DIFF
--- a/dask_expr/_collection.py
+++ b/dask_expr/_collection.py
@@ -726,9 +726,15 @@ class FrameBase(DaskMethodsMixin):
         >>> df = df.shuffle(df.columns[0])  # doctest: +SKIP
         """
         if on is no_default and not on_index:
-            raise TypeError("Have to pass on or set index_shuffle to True")
+            raise TypeError(
+                "Must shuffle on either columns or the index; currently shuffling on "
+                "neither. Pass column(s) to 'onn' or set 'on_index' to True."
+            )
         elif on is not no_default and on_index:
-            raise TypeError("Can't pass on and set index_shuffle to True")
+            raise TypeError(
+                "Cannot shuffle on both columns and the index. Do not pass column(s) "
+                "to 'on' or set 'on_index' to False."
+            )
 
         # Preserve partition count by default
         npartitions = npartitions or self.npartitions
--- a/dask_expr/_collection.py
+++ b/dask_expr/_collection.py
@@ -679,7 +679,7 @@ class FrameBase(DaskMethodsMixin):
 
     def shuffle(
         self,
-        on: str | list,
+        on: str | list | no_default = no_default,
         ignore_index: bool = False,
         npartitions: int | None = None,
         shuffle_method: str | None = None,
@@ -703,6 +703,9 @@ class FrameBase(DaskMethodsMixin):
             be preserved by default.
         shuffle_method : optional
             Desired shuffle method. Default chosen at optimization time.
+        index_shuffle : bool, default False
+            Whether to shuffle the index. Needs to be set to True with no
+            on given to shuffle the index.
         **options : optional
             Algorithm-specific options.
 
@@ -715,6 +718,10 @@ class FrameBase(DaskMethodsMixin):
         --------
         >>> df = df.shuffle(df.columns[0])  # doctest: +SKIP
         """
+        if on is no_default and not index_shuffle:
+            raise TypeError("Have to pass on or set index_shuffle to True")
+        elif on is not no_default and index_shuffle:
+            raise TypeError("Can't pass on and set index_shuffle to True")
 
         # Preserve partition count by default
         npartitions = npartitions or self.npartitions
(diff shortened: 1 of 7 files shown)


phofl:
dask/dask update is here: https://github.com/dask-contrib/dask-expr/pull/887

hendrikmakait:
> dask/dask update is here: https://github.com/dask-contrib/dask-expr/pull/887

This links to this issue, not one on `dask/dask`.

phofl:
sorry 

https://github.com/dask/dask/pull/10944

dask_expr/_collection.py
```diff
@@ -721,6 +725,10 @@ def shuffle(
         --------
         >>> df = df.shuffle(df.columns[0])  # doctest: +SKIP
         """
+        if on is no_default and not index_shuffle:
+            raise TypeError("Have to pass on or set index_shuffle to True")
+        elif on is not no_default and index_shuffle:
+            raise TypeError("Can't pass on and set index_shuffle to True")
```
hendrikmakait, line 731: I'm a bit confused. IIUC, this states that 

IF `on is no_default`, THEN `index_shuffle` has to be `True`. 
OTHERWISE, (i.e, `on is not no_default`), THEN `index_shuffle` has to be `False`.

If that's correct, why not simply infer the value via `index_shuffle = on is no_default`?

phofl: Too implicit for my liking, especially since this is a new keyword

hendrikmakait: From an offline conversation: I'd suggest renaming `shuffle_index` to `on_index` which should make the relationship between `on` and `on_index` more explicit.

hendrikmakait: I'd also adjust the error message to something that makes the relationship clearer, e.g., 

"Must shuffle on either columns or the index; currently shuffling on neither. Pass column(s) to `on` or set `on_index` to `True`."

"Cannot shuffle on both columns and the index. Do not pass column(s) to `on` or set `on_index` to `False`.

dask_expr/_expr.py
```diff
@@ -3174,6 +3174,29 @@ def _lower(self):
...
         ):
             return self._expr_cls(*self.operands)
+        elif self.divisions[0] is None:
+            # We have to shuffle
+            npartitions = max(df.npartitions for df in dfs)
+            dtypes = {df._meta.index.dtype for df in dfs}
+            if not _are_dtypes_shuffle_compatible(dtypes):
+                raise TypeError(
+                    "DataFrames are not aligned. We need to shuffle to align partitions "
+                    "with each ohter. This is not possible because the indexes of the "
```
hendrikmakait, line 3184:
```suggestion
                    "with each other. This is not possible because the indexes of the "
```

dask_expr/_expr.py
```diff
@@ -3174,6 +3174,29 @@ def _lower(self):
...
+                    f"DataFrames have differing dtypes={dtypes}. Please ensure that "
+                    f"all Indexes have the same dtype or align manually for this to "
+                    f"work."
```
hendrikmakait, line 3187:
```suggestion
                    "all Indexes have the same dtype or align manually for this to "
                    "work."
```

dask_expr/_expr.py
```diff
@@ -3664,15 +3695,10 @@ def _check_dependents_are_predicates(
     return other_names.issubset(allowed_expressions)
 
 
-def calc_divisions_for_align(*exprs):
+def calc_divisions_for_align(*exprs, allow_shuffle=True):
```
hendrikmakait, line 3698: Do we use `allow_shuffle` anywhere?

hendrikmakait (commented):
Mostly nits, one question about an unused (?) kwarg.

dask_expr/_expr.py
phofl: Yeah we still do in map_overlap, this is also on my bucket list after that one is in

PR #888, opened 2024-02-21
Fix merge predicate pushdowns with weird predicates

phofl:
I thought a little about how we do predicate pushdowns and stumbled upon this, we currently don't cover weird predicates very well, we have to walk the predicate until we arrive at the merge to determine from which side they are from

### DIFF
--- a/dask_expr/_merge.py
+++ b/dask_expr/_merge.py
@@ -26,6 +26,8 @@ from dask_expr._expr import (  # noqa: F401
     PartitionsFiltered,
     Projection,
     Unaryop,
+    _DelayedExpr,
+    are_co_aligned,
     determine_column_projection,
     is_filter_pushdown_available,
 )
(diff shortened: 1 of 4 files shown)


dask_expr/_merge.py
```diff
@@ -108,6 +110,45 @@ def _filter_passthrough_available(self, parent, dependents):
...
+                return self._walk_predicates(predicate.left)
+            elif isinstance(predicate.right, Elemwise):
+                return self._walk_predicates(predicate)
+            else:
+                return None
+        else:
+            # Unsupported predicate type
+            return None
+
+    def _walk_predicates(self, predicate):
```
fjetter, line 130: nit: the method names doesn't really tell what's going on. To understand what's going on one has to read and understand the entire method which I think is not necessary with a better name.

Even `_predicate_columns_inner` or smth like this would be better I think

fjetter (approved):

Issue #770, opened 2024-01-18
Centralise alignment handling and push to later in the process


phofl:
There are currently a few places where we check for pandas alignment and add repartitions if necessary, these places should be centralised and ideally happen a little bit later. I deferred this until now for velocity in fixing tests.

I'll take care of this when I get back, just keeping track here.

PR #891, opened 2024-02-22
Remove eager divisions from indexing

phofl:
closes https://github.com/dask-contrib/dask-expr/issues/770


hendrikmakait (approved):
Thanks, @phofl!

PR #895, opened 2024-02-26
Use Aggregation from dask/dask

phofl:
This caused weird import errors and we use the same class anyway


milesgranger (approved):
Good for me! :+1:

hendrikmakait:
CI is really unhappy, is this related?

PR #896, opened 2024-02-26
DNM: Full branch_id implementation

phofl:


dask_expr/_expr.py
```diff
@@ -2758,18 +2764,29 @@ def normalize_expression(expr):
...
+) -> Expr:
     result = expr
     if stage == "logical":
         return result
 
-    # Simplify
-    expr = result.simplify()
+    while True:
+        if not common_subplan_elimination:
+            out = result.rewrite("reuse", cache={})
```
hendrikmakait, line 2776: While we have cleaned up the meaning of `common_subplan_elimination`, it's still weird to me that we execute the `reuse` rule to _avoid_ CSE. Maybe we should rename it to something like `avoid_common_subplan_elimination`?

dask_expr/_expr.py
```diff
@@ -2805,6 +2824,10 @@ def optimize(expr: Expr, fuse: bool = True) -> Expr:
         Input expression to optimize
     fuse:
         whether or not to turn on blockwise fusion
+    common_subplan_elimination : bool, default False
```
hendrikmakait, line 2827: Note that `common_subplan_elimination` does not mean _no_ CSE but rather _less_ CSE. We may want to reflect that in the docstring.

dask_expr/tests/test_distributed.py
```diff
@@ -354,3 +355,186 @@ def test_func(n):
...
+    _check_consumer_node(q, 1)
+    _check_consumer_node(q, 2, consumer_node=P2PShuffle)
+    x = c.compute(q)
+    x = await x
+
+    expected = pdf.fillna(100)
+    expected = expected.a + expected.a.sum()
+    pd.testing.assert_series_equal(x.sort_index(), expected)
+
+    # Check that we have 1 shuffle barrier but 20 p2pshuffle tasks for the output
```
hendrikmakait, line 377: IIUC, this PR introduces functionality that relies on being able to read shuffle outputs several times. The fact that this seems to work with disk-based P2P is a lucky implementation detail but not guaranteed to work, let alone tested. Before releasing, we should test this. (It will also not work with in-memory P2P, but that's currently not supported in `dask-expr` anyway.)

hendrikmakait:
One general comment: This PR introduces many different names for seemingly similar things, e.g., branch vs. subplan, reuse vs. elimination. We may want to clean this up before merging to make it easier to grasp concepts.

dask_expr/_shuffle.py
```diff
@@ -555,7 +571,8 @@ def _layer(self):
         )
 
         dsk = {}
-        token = self._name.split("-")[-1]
+        # Ensure that shuffles with different branch_ids have the same barrier
+        token = self._dep_name.split("-")[-1]
```
fjetter, line 575: I'm pretty strongly -1 for this. The fact that this works is purely coincidental. The barrier should be treated as an internal implementation detail since way too much logic depends on this. If we want/need this functionality, it should be supported as a proper API of the extension

hendrikmakait, line 575: From a conversation with @fjetter: This is severely leaking internals of the P2P implementation.

hendrikmakait: I see the benefit of reusing results that are written to disk, but I agree that the fact that this works is purely coincidental and very brittle wrt to changes.

From what I see, this might be useful if it were
* well-tested (also on the P2P side)
* not hidden as what looks like an implementation detail within the `_layer`

I'm not sure if this is something we can implement within the P2P extension. _Maybe_ we can do this after we have the scheduler integration? I could also see this become an optimization pass that makes this very explicit.

PR #899, opened 2024-02-26
P2P shuffle drops partitioning column early

hendrikmakait:
Closes https://github.com/dask/distributed/issues/8530

Sibling to https://github.com/dask/distributed/pull/8531.

I don't like that the column gets assigned outside of the shuffle but dropped inside the shuffle. We may want to restructure this in a follow-up.


phofl:
I'll do this in a follow up, we already do something similar in merge, so best to generalise the function that does this I guess

phofl (approved):

phofl:
thx

PR #906, opened 2024-02-29
Implement more efficient GroupBy.mean

phofl:
This actually made a bigger difference than expected, it cuts query 17 from 5 minutes to 3:20, shuffling twice is just expensive

I will reactor the layout a little bit in a follow-up, but didn't want to touch too many things at once


milesgranger (approved):
Looks good to me! :+1:

Issue #907, opened 2024-02-29
`are_co_aligned` slow and possibly dangerous


fjetter:
The utility function `are_co_aligned`, see

https://github.com/dask-contrib/dask-expr/blob/9334e062a7b41161977ca1c42176197629569cc5/dask_expr/_expr.py#L2863-L2874

is unfortunately rather slow due to the tokenization and lack of caching. in paruqet_reader benchmarks on larger datasets, I saw this slowing down the optimize step by almost a second (when using pyarrowFS such that filters are pushed down

![image](https://github.com/dask-contrib/dask-expr/assets/8629629/15cf2cef-48e3-4ebe-9468-eeaa51969e0a)

On top of this, I believe the implementation is unsafe since it is putting `Expr` objects into a `set`. Sets and dicts are requiring both `__hash__` and `__eq__` to be implemented and working as the stdlib protocol defines them. While this is true for hash (it hashes the name, this is not the case for `__eq__` since this just creates another `Expr` instance instead of returning a bool. I suspect this just tells how the set is redundant if there hasn't been ever a hash collision / duplicate object here.

phofl:
The unsafe thing is relatively easy to fix (see pr). The slowdown probably needs a bit of care

fjetter:
Also, I think the `are_co_aligned` (which is a bit of a misleading function name considering that we typically talk about alignment in the context of partitions) is buggy since it doesn't include `filters`

phofl:
Are you talking about filters in the tokenize partial call? Filters can change the number of partitions with pruning so we shouldn’t include them

I thought a little bit about the filters in read parquet, we shouldn’t use are_co_aligned there anyway, it’s too generous and allows expressions that we don’t want to allow, I have a solution for this soonish

fjetter:
> Are you talking about filters in the tokenize partial call?

yes. I haven't thought too deeply about this but it is confusing that it isn't included while columns are. No big deal, though

fjetter:
Looking into this a little more deeply, it appears that the `_tokenize_partial` is slow because it does tokenize too many things. The parquet reader for instance is reusing one of the paramters as a cache and tokenizing this cache is what makes it costly. This info is represented in `ReadParquet` by overwriting the `_name` method but this is obviously not respected by the `_tokenize_partial`

fjetter:
[...]

PR #909, opened 2024-02-29
Speedup init of `ReadParquetPyarrowFS`

fjetter:
This includes a couple of fixes to the `ReadParquetPyarrowFS` expr that speed up the initial kick off significantly. On the TPCH 1000 scale datasets for Q1 creating the query and optimizing it initially took 10-11s on my machine (with me sitting in europe and the dataset in us-east-2). With these fixes, we're down to 1-2s which is on par with the fsspec implementation.

### DIFF
--- a/dask_expr/_expr.py
+++ b/dask_expr/_expr.py
@@ -2867,7 +2867,7 @@ def are_co_aligned(*exprs, allow_broadcast=True):
     ancestors = [set(non_blockwise_ancestors(e)) for e in exprs]
     unique_ancestors = {
         # Account for column projection within IO expressions
-        _tokenize_partial(item, ["columns", "_series"])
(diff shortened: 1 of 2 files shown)


dask_expr/_expr.py
```diff
@@ -2867,7 +2867,7 @@ def are_co_aligned(*exprs, allow_broadcast=True):
     ancestors = [set(non_blockwise_ancestors(e)) for e in exprs]
     unique_ancestors = {
         # Account for column projection within IO expressions
-        _tokenize_partial(item, ["columns", "_series"])
+        _tokenize_partial(item, ["columns", "_series", "_dataset_info_cache"])
```
fjetter, line 2870: See https://github.com/dask-contrib/dask-expr/issues/907 for the root cause discussion. What I'm doing here is a pretty brittle patch imo

phofl:
thx

PR #910, opened 2024-02-29
Fix detection of parquet filter pushdown

phofl:
we don't want to use are_co_aligned here, it's way too permissive.

```
    # Don't push down if replace is in there
    x = df[df.replace(5, 50).a == 5]["b"]
    y = x.optimize(fuse=False)
    assert isinstance(y.expr, Filter)
    assert len(y.compute()) == 0

```

This shouldn't move the filters inside of read_parquet

Another lessons learned for me personally is that we shouldn't use it in hot-paths in the current implementation

### DIFF
--- a/dask_expr/io/parquet.py
+++ b/dask_expr/io/parquet.py
@@ -48,7 +48,6 @@ from dask_expr._expr import (
     Literal,
     Or,
     Projection,
-    are_co_aligned,
     determine_column_projection,
 )
 from dask_expr._reductions import Len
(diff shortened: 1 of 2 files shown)


fjetter:
is this still needed with https://github.com/dask-contrib/dask-expr/pull/909#discussion_r1507523471 ?

phofl:
PR title is bad (sorry), this is actually more of a bug fix that brings a performance improvement as well. Currently we are too eagerly pushing down filters into the read_parquet call (see the example in the top post)

fjetter (approved):
good catch

PR #911, opened 2024-02-29
Wrap fragments

fjetter:
This is not my preferred way of doing things but this effectively bypasses https://github.com/apache/arrow/issues/40279 and delays the initialization cost of the filesystem to the workers. This is only necessarily because we no longer have the "don't deserialize on scheduler" mechanics (at least not properly). The most important part is the delay of the filesystem deserialization. However, storing this as a bytestring comes at a cost since pickle can't seem to deduplicate this any longer such that the overall size of the serialized graph is larger. 

I still have to run a few more tests to see if this is all really worth it.

### DIFF
--- a/dask_expr/io/parquet.py
+++ b/dask_expr/io/parquet.py
@@ -130,7 +130,7 @@ class FragmentWrapper:
             if fs is None:
                 fs = pickle.loads(fs_raw)
                 FragmentWrapper._filesystems[fs_raw] = fs
-            # arrow doens't keep the python object alive so if we want to reuse
+            # arrow doesn't keep the python object alive so if we want to reuse
             # we need to keep a reference
             self._fs = fs
             self._fragment = pqformat.make_fragment(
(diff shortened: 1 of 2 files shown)


dask_expr/io/parquet.py
```diff
@@ -714,54 +840,51 @@ def fragments(self):
...
+            _fragment_to_pandas,
+            FragmentWrapper(self.fragments[index]),
             self.columns,
             self.filters,
-            self._dataset_info["schema"],
+            self._dataset_info["schema"].remove_metadata(),
         )
 
 
+def _fragment_to_pandas(fragment_wrapper, columns, filters, schema):
```
fjetter, line 853: making this a proper function was not strictly necessary but I wanted to be extra sure that there is no weird reference. Also, I find this a little cleaner but it doesn't really matter, it's just style

dask_expr/io/parquet.py
```diff
@@ -85,6 +87,128 @@ def _tokenize_fileinfo(fileinfo):
...
+    def pack(self):
+        if self._fragment_packed is None:
+            self._fragment_packed = (
+                self._fragment.format,
+                (
+                    self._fragment.path
+                    if self._fragment.buffer is None
+                    else self._fragment.buffer
+                ),
+                pickle.dumps(self._fragment.filesystem),
```
fjetter, line 114: This is truly odd. When I keep the filesystem as is, I get a memory leak on the scheduler I'm afraid this is a genuine C++ memory leak but I couldn't reproduce it locally. I can just tell that from manual testing the leak vanished when I pickle this explicitly.

dask_expr/io/parquet.py
```diff
@@ -85,6 +87,128 @@ def _tokenize_fileinfo(fileinfo):
...
+                pqformat,
+                path_or_buffer,
+                fs_raw,
+                partition_expression,
+                file_size,
+            ) = self._fragment_packed
+            fs = FragmentWrapper._filesystems.get(fs_raw)
+            if fs is None:
+                fs = pickle.loads(fs_raw)
+                FragmentWrapper._filesystems[fs_raw] = fs
```
fjetter, line 132: Note: even if I do not deduplicate but just `loads` the filesystem, the memory leak is gone. No idea what's going on but by caching this we avoid the initialization cost, even if it's just a ms

dask_expr/io/parquet.py
```diff
@@ -85,6 +87,128 @@ def _tokenize_fileinfo(fileinfo):
...
+        self.unpack()
+        return self._fragment
+
+    def __reduce__(self):
+        self.pack()
+        return FragmentWrapper, (None, None, self._fragment_packed)
+
+
+class FragmentWrapper:
+    _filesystems = weakref.WeakValueDictionary()
```
hendrikmakait, line 155: Am I missing something or is this the same as the class above? At the very least, it's the same name.

fjetter: yes, that's the same. odd. I was moving branches too frequently and I probably committed some garbage :/

dask_expr/io/parquet.py
```diff
@@ -85,6 +87,69 @@ def _tokenize_fileinfo(fileinfo):
 _CACHED_PLAN_SIZE = 10
 _cached_plan = {}
 
+import pickle
+import weakref
+
```
fjetter, line 92: out linter should find something like this

phofl (approved):

phofl:
thx

PR #917, opened 2024-03-01
[Parquet] Calculate divisions from statistics

fjetter:
This is a WIP/POC kind of things to see how to fetch statistics the best.

It's implementing the "calculate divisions for index" feature again but I'm mostly interested in playing with the stats collection a little

### DIFF
--- a/dask_expr/_collection.py
+++ b/dask_expr/_collection.py
@@ -400,6 +400,9 @@ class FrameBase(DaskMethodsMixin):
             other = other.copy()
         return new_collection(self.expr.__getitem__(other))
 
+    def __dask_tokenize__(self):
+        return type(self).__name__, self._expr._name
+
     def __bool__(self):
         raise ValueError(
             f"The truth value of a {self.__class__.__name__} is ambiguous. "
--- a/dask_expr/_collection.py
+++ b/dask_expr/_collection.py
@@ -4665,10 +4665,6 @@ def read_parquet(
             raise ValueError(
                 "pyarrow>=15.0.0 is required to use the pyarrow filesystem."
             )
-        if calculate_divisions:
-            raise NotImplementedError(
-                "calculate_divisions is not supported when using the pyarrow filesystem."
-            )
         if metadata_task_size is not None:
             raise NotImplementedError(
                 "metadata_task_size is not supported when using the pyarrow filesystem."
@@ -4701,6 +4697,7 @@ def read_parquet(
                 filters=filters,
                 categories=categories,
                 index=index,
+                calculate_divisions=calculate_divisions,
                 storage_options=storage_options,
                 filesystem=filesystem,
                 ignore_metadata_file=ignore_metadata_file,
--- a/dask_expr/io/parquet.py
+++ b/dask_expr/io/parquet.py
@@ -750,7 +750,12 @@ class ReadParquetPyarrowFS(ReadParquet):
             files = self._dataset_info["all_files"]
         if fragments is None:
             fragments = self.fragments_unsorted
-        token_stats = flatten(dask.compute(_collect_statistics_plan(files, fragments)))
+        # Collecting code samples is actually a little expensive (~100ms) and
+        # we'd like this thing to be as low overhead as possible
+        with dask.config.set({"distributed.diagnostics.computations.nframes": 0}):
+            token_stats = flatten(
+                dask.compute(_collect_statistics_plan(files, fragments))
+            )
         for token, stats in token_stats:
             _STATS_CACHE[token] = stats
 
--- a/dask_expr/io/parquet.py
+++ b/dask_expr/io/parquet.py
@@ -625,7 +625,7 @@ class ReadParquet(PartitionsFiltered, BlockwiseIO):
             return header
         return super()._tree_repr_argument_construction(i, op, header)
 
-    @property
+    @cached_property
     def _meta(self):
         meta = self._dataset_info["base_meta"]
         columns = _convert_to_list(self.operand("columns"))
@@ -745,17 +745,12 @@ class ReadParquetPyarrowFS(ReadParquet):
         stats = [_STATS_CACHE[tokenize(finfo)] for finfo in files_to_consider]
         return _combine_stats(stats)
 
-    def load_statistics(self, files=None, fragments=None, scheduler=None):
+    def load_statistics(self, files=None, fragments=None):
         if files is None:
             files = self._dataset_info["all_files"]
         if fragments is None:
             fragments = self.fragments_unsorted
-        token_stats = flatten(
-            dask.compute(
-                _collect_statistics_plan(files, fragments),
-                scheduler=scheduler,
-            )
-        )
+        token_stats = flatten(dask.compute(_collect_statistics_plan(files, fragments)))
         for token, stats in token_stats:
             _STATS_CACHE[token] = stats
 
@@ -788,11 +783,7 @@ class ReadParquetPyarrowFS(ReadParquet):
             ixs.append(sort_ix)
             finfos_sampled.append(finfos[sort_ix])
             frags_samples.append(frags[sort_ix])
-        # Note: Providing get explicitly avoids a warning that is raised if a
-        # distributed client is in context
-        from dask.local import get_sync
-
-        self.load_statistics(finfos_sampled, frags_samples, scheduler=get_sync)
+        self.load_statistics(finfos_sampled, frags_samples)
         return ixs
 
     @cached_property
@@ -1738,13 +1729,13 @@ def _aggregate_statistics_to_file(stats):
 
 @dask.delayed
 def _gather_statistics(frags):
+    @dask.delayed
     def _collect_statistics(token_fragment):
         return token_fragment[0], _extract_stats(token_fragment[1].metadata.to_dict())
 
-    from concurrent.futures import ThreadPoolExecutor
-
-    with ThreadPoolExecutor() as tpe:
-        return list(tpe.map(_collect_statistics, frags))
+    return dask.compute(
+        list(_collect_statistics(frag) for frag in frags), scheduler="threading"
+    )[0]
 
 
 def _collect_statistics_plan(file_infos, fragments):
--- a/dask_expr/io/parquet.py
+++ b/dask_expr/io/parquet.py
@@ -4,13 +4,15 @@ import contextlib
 import itertools
 import operator
 import pickle
+import statistics
 import warnings
 import weakref
 from abc import abstractmethod
 from collections import defaultdict
-from functools import cached_property
+from functools import cached_property, partial
 
 import dask
+import numpy as np
 import pandas as pd
 import pyarrow as pa
 import pyarrow.dataset as pa_ds
@@ -18,6 +20,7 @@ import pyarrow.fs as pa_fs
 import pyarrow.parquet as pq
 import tlz as toolz
 from dask.base import normalize_token, tokenize
+from dask.core import flatten
 from dask.dataframe.io.parquet.core import (
     ParquetFunctionWrapper,
     ToParquetFunctionWrapper,
@@ -65,6 +68,9 @@ def _tokenize_fileinfo(fileinfo):
         fileinfo.mtime_ns,
         fileinfo.size,
     )
+
+
+_STATS_CACHE = {}
 
 
 PYARROW_NULLABLE_DTYPE_MAPPING = {
@@ -654,6 +660,7 @@ class ReadParquetPyarrowFS(ReadParquet):
         "storage_options",
         "filesystem",
         "ignore_metadata_file",
+        "calculate_divisions",
         "kwargs",
         "_partitions",
         "_series",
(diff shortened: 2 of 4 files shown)


dask_expr/io/parquet.py
```diff
@@ -759,8 +929,23 @@ def _filtered_task(self, index: int):
...
+            return 1
+        approx_stats = self.approx_statistics()
+        total_uncompressed = 0
+        for col in approx_stats["columns"]:
+            total_uncompressed += col["total_uncompressed_size"]
+        after_projection = 0
+        for col in self.operands("columns"):
+            after_projection += col["total_uncompressed_size"]
+
+        return max(after_projection / total_uncompressed, 0.001)
```
fjetter, line 947: @phofl this now calculates the approx compression factor of the columns using parquet statistics.
The parquet statistics here originate from a sample of files. Currenlty I'm sampling the smallest, largest and median (based on file size) and am averaging them. This aggregation/approximation strategy can be adjusted of course but I figured I'll start somewhere and this isn't horribly, I hope.

With this sample on a scale 1k TPCH dataset this gives these statistics

image

Therefore, whenever comments are dropped, this will cause quite drastic compression. I'll shortly run some A/B tests

dask_expr/io/parquet.py
```diff
@@ -631,6 +641,125 @@ def fs(self):
...
+    @cached_property
+    def raw_statistics(self):
+        self.load_all_statistics()
+        return [
+            _STATS_CACHE[tokenize(finfo)] for finfo in self._dataset_info["all_files"]
+        ]
+
+    @cached_property
+    def aggregated_statistics(self):
+        return _aggregate_statistics_to_file(self.raw_statistics)
```
fjetter, line 755: TODO: docs

fjetter:
> > It's implementing the "calculate divisions for index" feature again but I'm mostly interested in playing with the stats collection a little
> 
> Based on this comment it sounds like you might be questioning the concept of divisions generally. FWIW if so I mostly agree. I think that query optimization should give us a richer ability to ask questions of statistics than we had before. Historically divisions was nice because it was simple-enough metadata that we could propogate it well. We have more of an ability to propogate today, so maybe divisions is too simplistic.

Well, I'm at the very least skeptical about their true value and am curious about what else is possible. 

With this specific comment I was mostly criticizing the usefulness of the "compute divisions from parquet statistics" feature since I suspect there are very few cases where this truly works. It's a rather complex feature that will only benefit few.

mrocklin:
> It's a rather complex feature that will only benefit few

Ah, interesting. My sense is that it's commonly useful for things like datetime columns or other values that might serve as a natural index.

dask_expr/io/parquet.py
```diff
@@ -1338,3 +1523,164 @@ def _normalize_and_strip_protocol(path):
...
+        last_max = file_max
+    divisions.append(last_max)
+    return tuple(divisions), argsort
+
+
+def _extract_stats(original):
+    """Take the raw file statistics as returned by pyarrow (as a dict) and
+    filter it to what we care about. The full stats are a bit too verbose and we
+    don't need all of it."""
+    # TODO: dicts are pretty memory inefficient. Move to dataclass?
```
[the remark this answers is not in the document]
fjetter: Storing this in a dict with rather long keynames is also not super efficient, of course.

dask_expr/io/parquet.py
fjetter: so, first prelim results show that the compaction is indeed much stronger which causes me to run into memory failures.

fjetter: The estimation is not too bad though.

above partition size calculated using memory_usage(deep=True), below the approximated size given the fusion compression factor

image

fjetter: the out of memory issues I run into seem to be a downstream memory blow up and not connected to the IO itself https://github.com/coiled/benchmarks/issues/1430

fjetter:
> > It's a rather complex feature that will only benefit few
> 
> Ah, interesting. My sense is that it's commonly useful for things like datetime columns or other values that might serve as a natural index.

Yeah, timeseries datasets are probably the best candidates, that's true. I think for other column types it is much less likely to line up to build proper divisions. Mostly when we have these kind of strict partitioning conditions we'll encounter hive-like tables but a hive partitioning is often not suited for a division index due to cardinality / partition sizes.
However, if we knew that partitions are hive-like, i.e. we knew that there is an injective mapping from values to partitions, that'd be powerful knowledge (even without knowing the mapping itself). This knowledge alone would suffice to perform efficient groupby aggregations, for instance.

anyhow, to the actual PR...

---

I think the PR is mostly ready to go. The statistics collection as implemented also suffers from the fragment serialization overhead but the collection itself is otherwise very fast since I'm using a threadpool during the collection.
As far as I can tell, this isn't triggered anywhere we don't want it to. The compaction factor is possibly problematic though. Also, I haven't written (m)any tests, yet 🤷‍♂️

dask_expr/io/parquet.py
```diff
@@ -693,6 +703,111 @@ def fs(self):
...
+        return [
+            _STATS_CACHE[tokenize(finfo)] for finfo in self._dataset_info["all_files"]
+        ]
+
+    @cached_property
+    def aggregated_statistics(self):
+        return _aggregate_statistics_to_file(self.raw_statistics)
+
+    def _get_lengths(self):
+        # TODO: Fitlers that only filter partition_expr can be used as well
```
phofl, line 806:
```suggestion
        # TODO: Filters that only filter partition_expr can be used as well
```

dask_expr/io/parquet.py
```diff
@@ -693,6 +703,111 @@ def fs(self):
...
+        -------
+        dict
+        """
+        idxs = self.sample_statistics()
+        files_to_consider = np.array(self._dataset_info["all_files"])[idxs]
+        stats = [_STATS_CACHE[tokenize(finfo)] for finfo in files_to_consider]
+        return _combine_stats(stats)
+
+    def load_statistics(self, files=None, fragments=None, scheduler=None):
+        if files is None:
```
phofl, line 752: Can this potentially load stats twice if called from different places?

fjetter: the statistics are cached in `_STATS_CACHE` by the fileinfo token. This includes the path, last modified, etc.
The `_collect_statistics_plan` is only considering collection of those files which are not already stored in the cache.

This approach would even allow us to store those stats on disk since cache invalidation should be straight forward if not trivial. This would make life easier for folks who are reading the same dataset over and over again.

dask_expr/io/parquet.py
```diff
@@ -667,12 +675,14 @@ class ReadParquetPyarrowFS(ReadParquet):
         "storage_options": None,
         "filesystem": None,
         "ignore_metadata_file": True,
+        "calculate_divisions": False,
+        "statistics": None,
```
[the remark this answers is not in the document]
fjetter: that's actually unused. I'll remove it again

phofl:
Is #906 in on both sides? I assume so, but that one would be an obvious culprit. It reduced runtime by roughly 40s, so kind of corresponds to what you are seeing here

fjetter:
It should be but I'll double check

fjetter:
Ok, I redid the benchmarks and this looks both concerning but also almost expected

image

The first thing we'll notice is that the entire bottom which measures the optimization time is red. This is the optimization time which also includes the statistics collection.
Looking at one of the bad offenders we see that this moving from 700ms to 1050ms which isn't great but fine. This is the additional request to S3 for the parquet stats.
image

I'm a little more concerned about the query runtimes and have to look into this a little. Particularly, I will rerun benchmarks once again after https://github.com/dask-contrib/dask-expr/pull/928 is benchmarked, reviewed and merged

fjetter:
I ran again some benchmarks against 67e2ec5953142fa81d85128e14d5d810a37540d6 and the results are not very encouraging. I think the more aggressive compaction of partitions is not playing nicely with other optimizations, e.g. https://github.com/dask-contrib/dask-expr/pull/940

I'll have to investigate a little more

image

fjetter:
If I can't stabilize the performance, we can hide this behind the collect_statistics flag and fall back to the trivial compaction ratio. maybe we'll need more functionality for this stats collection overhead to truly shine

dask_expr/_collection.py
```diff
@@ -400,6 +400,9 @@ def __getitem__(self, other):
             other = other.copy()
         return new_collection(self.expr.__getitem__(other))
 
+    def __dask_tokenize__(self):
+        return type(self).__name__, self._expr._name
```
fjetter, line 404: The collection is also being tokenized on compute. Don't know exactly why but it's been using the repr so far which helped me find https://github.com/dask-contrib/dask-expr/pull/941

dask_expr/io/parquet.py
```diff
@@ -693,6 +700,123 @@ def fs(self):
...
+        if files is None:
+            files = self._dataset_info["all_files"]
+        if fragments is None:
+            fragments = self.fragments_unsorted
+        # Collecting code samples is actually a little expensive (~100ms) and
+        # we'd like this thing to be as low overhead as possible
+        with dask.config.set({"distributed.diagnostics.computations.nframes": 0}):
+            token_stats = flatten(
+                dask.compute(_collect_statistics_plan(files, fragments))
+            )
```
fjetter, line 758: I moved this from using a sync executor to using the default scheduler. initially, I figured it is cheaper to just run this locally since the collect statistics thing internally will also use a threading executor. However, dispatching to a cluster, if available, makes sense since the cluster is more likely to be closer to the data than the client.

This entire thing adds an overhead of about 300-400ms to the optimize. This is happening in the tune_up step

fjetter:
I optimized a couple of client side things that slowed things down a couple of seconds in total. This overhead could easily explain why we're seeing those poor benchmarks. I'm running a new AB on this vs 60075aac335470987c617de6c4dbbc5dd0c8c85d

fjetter:
For full transparency, we're not resetting any caches in these benchmarks so particularly later queries are benefitting from this. A stats lookup takes 200-300ms per dataset so multi dataset queries would be slowed down by maybe a second with a cold cache. For query 11 (the one that is ~35% faster) this would be a slowdown of about 5% or less so with or without cache warmup, this is a great result.

I'll poke briefly at Q1 and Q18 but I think I'm happy with the results as is for now, particularly considering that this may unlock quite interesting optimizations down the line.

phofl:
I suspect that we are giving some of the gains from #940 back with this since we are now operating on bigger partitions (probably), reducing the factor 20 to something like 10 might help us speed things up again if the average partition is now larger

Q1 could be a similar root cause, but I haven't looked into this at all so this is only guesswork

fjetter:
I ran Query 11 manually and it is returning an empty result. This doesn't sound right. Haven't checked against main, yet, but if we're dropping too much data that'd explain the extreme speedup

fjetter:
main is also returning an empty result 🤔

hendrikmakait:
> main is also returning an empty result 🤔
That's odd, the correctness tests in our benchmark suite haven't failed 🤔 

With what parameters do you run the queries?

fjetter:
> That's odd, the correctness tests in our benchmark suite haven't failed 🤔 With what parameters do you run the queries?

what do you mean with parameters?

phofl:
I can't reproduce on scale 1 locally but I __can__ reproduce on scale 10, so that's probably why we don't fail on the correctness tests

phofl:
The query looks like an empty result might be expected. The data are dropped in the filter step right at then end and the way the filter is written makes an empty result with increasing scale more likely. (DuckDB doesn't have answers for scale 10 unfortunately)

fjetter:
Well, I'm lacking a row count or anything like this but the byte size suggest that the first merge already drops much of the data. This is scale 10

analyze-operation-ed0fe2c74514a0016e409b0ae4b7a0d0

fjetter:
Either way, this PR is generating _nothing_ more quickly than main so I guess we can move forward and merge. If this is indeed an issues we should track it somewhere else

phofl:
> Well, I'm lacking a row count or anything like this but the byte size suggest that the blockwise merge already drops much of the data. This is scale 1000

The nation set is restricted to 1/24 countries, so that looks fine

dask_expr/io/parquet.py
```diff
@@ -1399,3 +1594,183 @@ def _normalize_and_strip_protocol(path):
...
+
+
+@dask.delayed
+def _gather_statistics(frags):
+    @dask.delayed
+    def _collect_statistics(token_fragment):
+        return token_fragment[0], _extract_stats(token_fragment[1].metadata.to_dict())
+
+    return dask.compute(
+        list(_collect_statistics(frag) for frag in frags), scheduler="threading"
```
phofl, line 1742: Just double checking my understanding here: This operates on the stats cache that is stored locally, so running it locally is what we want, correct?

fjetter: `_gather_statistics` (the outer function) is being dispatched to a cluster if one is available while the inner function `_collect_statistics` is executed on that cluster inside of a worker and is being parallelized using a threadpool.

typing this now makes me wonder how well dask handles nested concurrency that is accessing the same threadpool. I could see this deadlock if it's using the same global pool. I'll double check.

fjetter: nope, everything good with the threading thing. We may spawn many threads but that's fine. Dask is handling this properly and ensures that the nested call get it's own threadpool

fjetter:
Re Q18 we're indeed now using very large partitions. At one point we're using 700-800MB partitions. I also see https://github.com/dask-contrib/dask-expr/pull/942 impacting this

phofl:
#940 not #942, we reduced the constant to 20 there, having bigger partitions in the beginning makes it reasonable to reduce the constant even further

fjetter:
What I meant is that #942 is also showing up in the analyze plots even though not necessary

image

PR #922, opened 2024-03-01
Implement array inference in new_collection

phofl:
Tests are here: https://github.com/dask-contrib/dask-expr/pull/921

Needs: https://github.com/dask/dask/pull/10973


milesgranger (approved):
Super, thanks @phofl!

PR #912, opened 2024-02-29
Add support for convert string option

phofl:
xref #691

this will probably break a bunch of tests in dask/dask

dask/dask pr: https://github.com/dask/dask/pull/10968

PR #923, opened 2024-03-02
Add string conversion to from_map

phofl:


phofl:
merged into #912, tests cross-depend, so that should be easier

PR #926, opened 2024-03-04
Re-run versioneer install to fix version number (tag_prefix)

jorisvandenbossche:
The `tag_prefix` as specified in pyproject.toml (`'v'`, which is also what is being used in practice) didn't match what's in the generated/committed `_version.y` file. so re-running `versioneer install --no-vendor` to fix this.

I used the latest version 0.29, which also seems to include some other formatting changes

Probably this is only needed for editable dev installs

### DIFF
--- a/dask_expr/_version.py
+++ b/dask_expr/_version.py
@@ -1,4 +1,3 @@
-
 # This file helps to compute a version number in source trees obtained from
 # git-archive tarball (such as those provided by githubs download-from-tag
 # feature). Distribution tarballs (built by setup.py sdist) and build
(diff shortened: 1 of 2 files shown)


dask_expr/_version.py
```diff
@@ -50,8 +51,8 @@ def get_config() -> VersioneerConfig:
     cfg = VersioneerConfig()
     cfg.VCS = "git"
     cfg.style = "pep440"
-    cfg.tag_prefix = ""
-    cfg.parentdir_prefix = "dask-"
+    cfg.tag_prefix = "v"
```
jorisvandenbossche, line 54: This is the actual relevant change to fix the "unknown" version number for editable installs

phofl:
Could you exclude it from precomit or run pre-commit over it?

jorisvandenbossche:
> which also seems to include some other formatting changes

ah, that's probably for running black afterwards. will do that here as well

phofl (approved):

phofl:
thx!

PR #927, opened 2024-03-04
Ensure that columns are copied in projection

phofl:
closes https://github.com/dask/dask/issues/10957

This seems like a sensible thing to protect against for regular lists


milesgranger (approved):
Thanks, @phofl!

PR #928, opened 2024-03-04
Concat arrow tables before converting to pandas

fjetter:
Apologies, this is building on top of https://github.com/dask-contrib/dask-expr/pull/917. I'll try to split apart

The last commit https://github.com/dask-contrib/dask-expr/commit/1e7cf9fa7c4739745b1e843bd7cb1ea069690ac4 introduces functionality that will load the individual chunks of a `FusedIO` block as arrow tables, concats those tables and only afterwards converts them to pandas.

With this I'm slowly approaching CPU saturation. The following plots show TPCH Q1 and Q6 which are both very IO heavy (6 more than 1)

Note: Those graphs also include https://github.com/dask-contrib/dask-expr/pull/917

image

and also nice network saturation

image

### DIFF
--- a/dask_expr/io/io.py
+++ b/dask_expr/io/io.py
@@ -177,7 +177,7 @@ class FusedParquetIO(FusedIO):
             )
             for frag, filter in frag_filters
         )
-        table = pa.concat_tables(tables, promote_options="permissive"))
+        table = pa.concat_tables(tables, promote_options="permissive")
         return ReadParquetPyarrowFS._table_to_pandas(table, *to_pandas_args)
 
     def _task(self, index: int):
--- a/dask_expr/io/io.py
+++ b/dask_expr/io/io.py
@@ -183,7 +183,7 @@ class FusedParquetIO(FusedIO):
             )
             for frag, filter in frag_filters
         )
-        table = pa.concat_tables(tables)
+        table = pa.concat_tables(tables, promote_options="permissive"))
         return ReadParquetPyarrowFS._table_to_pandas(table, *to_pandas_args)
 
     def _task(self, index: int):
--- a/dask_expr/io/io.py
+++ b/dask_expr/io/io.py
@@ -5,6 +5,7 @@ import math
 import operator
 
 import numpy as np
+import pyarrow as pa
 from dask.dataframe import methods
 from dask.dataframe._pyarrow import to_pyarrow_string
 from dask.dataframe.core import apply_and_enforce, is_dataframe_like, make_meta
@@ -145,6 +146,64 @@ class FusedIO(BlockwiseIO):
 
     def _tune_up(self, parent):
         return
+
+
+class FusedParquetIO(FusedIO):
+    _parameters = ["_expr"]
+
+    @functools.cached_property
+    def _name(self):
+        return (
+            funcname(type(self.operand("_expr"))).lower()
+            + "-fused-parq-"
+            + _tokenize_deterministic(*self.operands)
+        )
+
+    @staticmethod
+    def _load_multiple_files(
+        frag_filters,
+        columns,
+        schema,
+        *to_pandas_args,
+    ):
+        from dask_expr.io.parquet import ReadParquetPyarrowFS
+
+        # Note: Ideally we'd build a pyarrow fragment that points to multiple
+        # tables since pyarrow handles IO and CPU threading separately and we're
+        # just overcomitting threads here. However, this isn't exposed at time
+        # of writing
+        # TODO: We may want to use a gloal pool with sufficient threads since
+        # executing this task in a local executor could be a little too much
+        tables = (
+            ReadParquetPyarrowFS._fragment_to_table(
+                frag,
+                filter,
+                columns,
+                schema,
+            )
+            for frag, filter in frag_filters
+        )
+        table = pa.concat_tables(tables)
+        return ReadParquetPyarrowFS._table_to_pandas(table, *to_pandas_args)
+
+    def _task(self, index: int):
+        expr = self.operand("_expr")
+        bucket = self._fusion_buckets[index]
+        fragments_filters = []
+        assert bucket
+        to_pandas_args = ()
+        for i in bucket:
+            _, frag_to_table, *to_pandas_args = expr._filtered_task(i)
+            fragments_filters.append((frag_to_table[1], frag_to_table[2]))
+            columns = frag_to_table[3]
+            schema = frag_to_table[4]
+        return (
+            self._load_multiple_files,
+            fragments_filters,
+            columns,
+            schema,
+            *to_pandas_args,
+        )
 
 
 class FromMap(PartitionsFiltered, BlockwiseIO):
--- a/dask_expr/io/io.py
+++ b/dask_expr/io/io.py
@@ -168,12 +168,6 @@ class FusedParquetIO(FusedIO):
     ):
         from dask_expr.io.parquet import ReadParquetPyarrowFS
 
-        # Note: Ideally we'd build a pyarrow fragment that points to multiple
-        # tables since pyarrow handles IO and CPU threading separately and we're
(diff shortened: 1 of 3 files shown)


mrocklin:
Cool.

> With this I'm slowly approaching CPU saturation

At some point (no urgency) I'll be curious to learn what is taking up CPU time in these cases.  My intuition says that we should still be IO bound.

dask_expr/io/io.py
```diff
@@ -147,6 +154,58 @@ def _tune_up(self, parent):
...
+        # executing this task in a local executor could be a little too much
+        tables = (
+            ReadParquetPyarrowFS._fragment_to_table(
+                frag,
+                filter,
+                columns,
+                schema,
+            )
+            for frag, filter in frag_filters
+        )
```
mrocklin, line 186: In a future where we tried using python threads for download concurrency, this is where we would use a ThreadPoolExecutor?  Is that correct?

fjetter: yes

fjetter:
> At some point (no urgency) I'll be curious to learn what is taking up CPU time in these cases. My intuition says that we should still be IO bound.


Well, CPU ~ memcpy. either arrow->pandas or pandas internal foo. Q1 for instance is not bound by IO at all any more

fjetter:
Also, the above plots were measured with COW enabled in case anybody is wondering

mrocklin:
>> At some point (no urgency) I'll be curious to learn what is taking up CPU time in these cases. My intuition says that we should still be IO bound.

> Well, CPU ~ memcpy. either arrow->pandas or pandas internal foo. Q1 for instance is not bound by IO at all any more

That surprises me given that IO is only running at 200 MB/s.  My intuition says that memcopies and pandas internal foo are much much faster than that.  Does your intuition disagree?  If you're ever around to share some profiles live I'd be curious (but again, not urgent)

fjetter:
I'm happy to talk about this but in the interest of sharing publicly, this is an annotated screenshot of the profile of Q1 (Heap view). I'll share the same for Q6 shortly

Q1_annotated_profile

[tpchq1.zip](https://github.com/dask-contrib/dask-expr/files/14484735/tpchq1.zip)

mrocklin:
Hrm, that is surprising.  Probably you're right then that concurrent reads would help a bit but not a ton.  (we could squeeze down that 30% a little, but not a lot).  

I don't suppose that using pure arrow dtypes for everything helps? (maybe it hurts)

mrocklin:
@phofl I'd be curious if this matches your expectations as well from a Pandas+Arrow perspective.  200 MB/s seems like a reasonable bottleneck given these operations, or is something likely off here?

phofl:
I am not super surprised about the operations that flare up there, groupby uses kahan summation to counteract numerical instabilities but you pay a price in performance. The ``dictionary_encode`` in arrow is something that could potentially be improved (we found bottlenecks in operations like this before in arrow), especially for high cardinality stuff.

table_to_pandas will be faster with arrow dtypes since it is zero copy and just a very thin wrapper, but chances are that you pay in performance later on since it's not as well tested as numpy dtypes

Does this answer your questions somehow?

mrocklin:
> Does this answer your questions somehow

I think I'm mostly thinking higher level.  I'll state my previous thinking:

-  Memory copies run at 20GB/s (Maybe 5-10 GB's on cloud containers)
- Inefficient-but-single-pass numpy/pandas code runs at 1-2GB/s
- S3 runs at 200 MB/s

If this is true then it's hard for pandas to be the bottleneck in cloud-based workflows.

What we're seeing here though is probably something more like "a couple of 500 MB/s steps (factorize and groupby-sum) which when added to solid S3 performance yield a total bandwidth of 200 MB/s".  Mostly I'm checking "does 500 MB/s seem right to you for these kinds of operations?"  If so then I'll update my mental model.

fjetter:
As promised, here is Q6 annotated. In my tests this is the query that is dominated the most by IO and in this example run above we're not saturating it fully, yet. As you can see, the CPU still has a little room. Another concurrency layer may help there but I'm still running tests. Particularly the pyarrow dtype has in previous tests shown a solid 10-20% perf gain and I could see this pushing us to the limit here (less "wasted" CPU time means the IO threads are also less idle since we're not able to pipeline perfectly)

To the query itself, we see that it breaks down to

- 63% IO and parquet deserialization (Unfortunately pyspy doesn't give me vision into the arrow threads)
- 18% arrow table to pandas (looks like we're converting a nullable float column here, i.e. this is not just memcpy and will go away if we use native dtypes)
- 17% evaluation of boolean masks. I suppose this can be optimized. This strikes me as a lot
- rest is a little take and the sum


Q6_annotated

mrocklin:
Cool.  That's fun to see.  This is the ~250MB/s read from the image in the top post?  I'd be curious to see what the limit is here at some point.

dask_expr/io/parquet.py
```diff
@@ -56,6 +57,15 @@
...
 from dask_expr.io import BlockwiseIO, PartitionsFiltered
 
+_CPU_COUNT_SET = False
+
+
+def _maybe_set_pacpu_count():
+    global _CPU_COUNT_SET
+    if not _CPU_COUNT_SET:
+        pa.set_cpu_count(os.cpu_count())
+    _CPU_COUNT_SET = True
```
fjetter, line 67: I'll factor this out

dask_expr/io/io.py
```diff
@@ -132,6 +133,12 @@ def _task(self, index: int):
         bucket = self._fusion_buckets[index]
         return (methods.concat, [expr._filtered_task(i) for i in bucket])
 
+    def _simplify_down(self):
```
phofl, line 136: I would rather overwrite the tune_up method in ReadParquetPyarrowFS, but that's not blocking

fjetter:
This is surprisingly bad. I'll look into this a little more since I got way better results with manual tests

image

dask_expr/io/io.py
```diff
@@ -147,6 +148,58 @@ def _tune_up(self, parent):
...
+        tables = (
+            ReadParquetPyarrowFS._fragment_to_table(
+                frag,
+                filter,
+                columns,
+                schema,
+            )
+            for frag, filter in frag_filters
+        )
+        table = pa.concat_tables(tables)
```
hendrikmakait, line 181: We may need to deal with type promotion (`promote_options="permissive")` here if we want to accommodate varying types in the input files.

hendrikmakait: See https://github.com/hendrikmakait/distributed/blob/c6ce3bb52d7c41437efc54b294dd38891fa89b97/distributed/shuffle/_arrow.py#L52-L62 for the corresponding code snippet from P2P.

fjetter: this all is guaranteed to use pyarrow>=15 so this compat things isn't necessary

hendrikmakait: We'd still need the `promote_options` kwarg if I'm not mistaken.

fjetter: yes, just not the compat foo

dask_expr/io/io.py
```diff
@@ -147,6 +148,58 @@ def _tune_up(self, parent):
...
+    def _task(self, index: int):
+        expr = self.operand("_expr")
+        bucket = self._fusion_buckets[index]
+        fragments_filters = []
+        assert bucket
+        to_pandas_args = ()
+        for i in bucket:
+            _, frag_to_table, *to_pandas_args = expr._filtered_task(i)
+            fragments_filters.append((frag_to_table[1], frag_to_table[2]))
+            columns = frag_to_table[3]
```
phofl, line 192: The fact that this get's overwritten on every pass is a little bit confusing, but can easily be a follow up

phofl (approved):

phofl:
thx

PR #936, opened 2024-03-05
Accept user arguments for arrow_to_pandas

fjetter:
This forwards the kwargs dtype_backend and types_mapper.

Re: types_mapper, this is a bit awkward. I'm not even sure if we need this. dask/dask has the possibility to accept a user defined type mapper. In fact, it allows passing all possible options to `arrow_to_pandas`. This is a very open approach. I implemented the same API for the new reader but I don't think many options are actually worth being piped through

### DIFF
--- a/dask_expr/io/parquet.py
+++ b/dask_expr/io/parquet.py
@@ -32,7 +32,6 @@ from dask.dataframe.io.parquet.core import (
 )
 from dask.dataframe.io.parquet.utils import _split_user_options
 from dask.dataframe.io.utils import _is_local_fs
-from dask.dataframe.utils import pyarrow_strings_enabled
 from dask.delayed import delayed
 from dask.utils import apply, funcname, natural_sort_key, parse_bytes, typename
 from fsspec.utils import stringify_path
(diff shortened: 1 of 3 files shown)


fjetter:
I ended up setting the dtype_backend to pyarrow here. I'll run an A/B but last time I checked this was better overall. I haven't tested against dask/dask but I suspect this will not flare up since the tests there are not parametrized over fs (yet)

fjetter:
This is scale 1000. Results are not as great as I was hoping

image

phofl:
I assume that the to_pandas call doesn't matter much for the more complex queries?

fjetter:
I'll revert the default but will keep the setting forwarding. this way we can merge and reassess the situation later

dask_expr/io/parquet.py
```diff
@@ -516,7 +515,7 @@ def pyarrow_type_mapper(pyarrow_dtype):
         type_mappers.append(user_types_mapper)
 
     # next in priority is converting strings
-    if convert_string:
+    if pyarrow_strings_enabled():
```
phofl, line 518: This is problematic, this would pull whatever is set on the cluster at that point in time. dask/dask preserves whatever is set when generating the collection. The fsspec implementation does this too but it's hidden very deep inside the legacy implementation

phofl (commented):
small comment

dask_expr/io/parquet.py
fjetter: good point. that's easily fixed/reverted

phofl (approved):
thx

PR #938, opened 2024-03-05
Avoid fusing from_pandas ops to avoid duplicating data

phofl:
fusing from_pandas expressions is problematic, in broadcasting expressions we fuse them into every partition, duplicating the data times the number of partitions. This created a 1.4GB task graph in our benchmarks compared to 67MB previously.

This gets us back to that number

### DIFF
--- a/dask_expr/_expr.py
+++ b/dask_expr/_expr.py
@@ -2925,6 +2925,10 @@ def are_co_aligned(*exprs):
 ## Utilites for Expr fusion
 
 
+def is_valid_blockwise_op(expr):
+    return isinstance(expr, Blockwise) and not isinstance(expr, (FromPandas, FromArray))
+
+
 def optimize_blockwise_fusion(expr):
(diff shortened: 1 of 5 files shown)


dask_expr/_expr.py
```diff
@@ -2925,6 +2925,10 @@ def are_co_aligned(*exprs):
 ## Utilites for Expr fusion
 
 
+def is_valid_blockwise_op(expr):
+    return isinstance(expr, Blockwise) and not isinstance(expr, (FromPandas, FromArray))
```
fjetter, line 2929: Wouldn't it be better to make `FromPandas` inherit from `IO` instead of `BlockwiseIO`?

phofl: This would require us to patch other places (culling for example), we have to swallow the extra logic somewhere unfortunately

fjetter (approved):

PR #940, opened 2024-03-06
Adjust automatic split_out parameter

phofl:
I tried avoiding fragmentation in the beginning, but I think we were too aggressive in squashing partitions during groupby. 

AB Test: https://github.com/coiled/benchmarks/actions/runs/8164285327

Query 10, 18 and 20 got faster (18 and 20 by a lot), everything else seems mostly noise except 13 where there might be a small slowdown, but not really noticeable.


hendrikmakait (approved):
Thanks, @phofl!

PR #941, opened 2024-03-06
Revert "Add _repr_html_ and updated __repr__ for FrameBase (#930)"

fjetter:
This reverts commit 226f5da06889a2a19db9d2cef9252a5e0ad2299a.

The `self.dask` triggers a graph materialization which is something that must not happen during a `__repr__`.

I suspect this code was copy pasted from the dask/dask DataFrame since it explicitly mentions "layers" but that's no longer valid in dask-expr. Instead, this is probably supposed to use `Expr.walk` or smth similar

cc @milesgranger


milesgranger (approved):
Thanks, will adjust tomorrow.

phofl:
Walk should be fine yes and just count the unique names of the expressions

phofl:
thx

PR #942, opened 2024-03-06
Remove repartition from P2P shuffle

phofl:


dask_expr/_shuffle.py
```diff
@@ -187,10 +187,11 @@ def _lower(self):
         # Reduce partition count if necessary
         frame = self.frame
         npartitions_out = self.npartitions_out
-        if npartitions_out < frame.npartitions:
+        method = self.method or get_default_shuffle_method()
+
+        if npartitions_out < frame.npartitions and method != "p2p":
             frame = Repartition(frame, new_partitions=npartitions_out)
```
fjetter, line 193: I'm confused about this overall I think. Aren't all the shuffle methods accepting an npartitions_out?
I know for a fact that simple, tasks and p2p does support those. If disk doesn't we should only add this to the disk shuffle

phofl: I don't know the internal workings of task shuffle well enough. I took this from dask/dask, e.g.

```
    if (
        shuffle_method != "p2p"
        and npartitions is not None
        and npartitions < df.npartitions
    ):
        df = df.repartition(npartitions=npartitions)
```

This is the logic in ``rearrange_by_column`` over in dask/dask

fjetter: ok, I don't care deeply about this. I was just surprised

fjetter (approved):

PR #946, opened 2024-03-07
Use implicit knowledge about divisions for efficient grouping

fjetter:
There are operations like Groupby+Apply that do not require the knowlege of the exact order of rows of the columns but are already happy knowing that a given value is only occurring in a single partition. If this was known we can avoid a shuffle.

This property can be calculated from parquet statistics (if one collects them) but it also occurs very naturally whenever we shuffle data. This is a POC on how such a property could be handled and utilized.

Note: There are even stronger properties we can pipe through here but this is a simple one to get started

### DIFF
--- a/dask_expr/_expr.py
+++ b/dask_expr/_expr.py
@@ -407,6 +407,9 @@ class Expr(core.Expr):
         except Exception:
             raise
 
+    def injective_mapping_columns(self):
+        return set()
+
     @property
     def _projection_columns(self):
         return self.columns
--- a/dask_expr/_expr.py
+++ b/dask_expr/_expr.py
@@ -2194,10 +2194,11 @@ class ResetIndex(Elemwise):
             predicate = None
             if not set(flatten(parents, list)).issubset(set(self.frame.columns)):
                 # one of the filters is the Index
-                self._filter_passthrough_available(parent, dependents)
-                name = self.operand("name")
-                if name is no_default:
+                name = self.operand("name") or self.frame._meta.index.name
+                if name is no_default and self.frame._meta.index.name is None:
                     name = "index"
+                elif self.frame._meta.index.name is not None:
+                    name = self.frame._meta.index.name
                 # replace the projection of the former index with the actual index
                 subs = Projection(self, name)
                 predicate = parent.predicate.substitute(subs, Index(self.frame))
@@ -2223,7 +2224,12 @@ class ResetIndex(Elemwise):
                 ):
                     return type(self)(self.frame, True, self.name)
                 return
-            return plain_column_projection(self, parent, dependents)
+            result = plain_column_projection(self, parent, dependents)
+            if result is not None and not set(result.columns) == set(
+                result.frame.columns
+            ):
+                result = result.substitute_parameters({"drop": True})
+            return result
 
 
 class AddPrefixSeries(Elemwise):
@@ -3704,6 +3710,9 @@ def plain_column_projection(expr, parent, dependents, additional_columns=None):
     )
     if isinstance(column_union, list):
         column_union = [col for col in expr.frame.columns if col in column_union]
+    elif column_union not in expr.frame.columns:
+        # we are accesing the index
+        column_union = []
     if column_union == expr.frame.columns:
         return
     result = type(expr)(expr.frame[column_union], *expr.operands[1:])
--- a/dask_expr/_expr.py
+++ b/dask_expr/_expr.py
@@ -2255,7 +2255,7 @@ class AddPrefix(Elemwise):
             f"{self.prefix}{c}"
             if not isinstance(c, tuple)
             else tuple(self.prefix + t for t in c)
-            for c in self.unique_partition_mapping_columns
+            for c in self.frame.unique_partition_mapping_columns
         }
 
     def _convert_columns(self, columns):
@@ -2286,7 +2286,7 @@ class AddSuffix(AddPrefix):
             f"{c}{self.suffix}"
             if not isinstance(c, tuple)
             else tuple(t + self.suffix for t in c)
-            for c in self.unique_partition_mapping_columns
+            for c in self.frame.unique_partition_mapping_columns
         }
 
     def _convert_columns(self, columns):
--- a/dask_expr/_expr.py
+++ b/dask_expr/_expr.py
@@ -1239,8 +1239,11 @@ class RenameFrame(Elemwise):
 
             columns = determine_column_projection(self, parent, dependents)
             columns = _convert_to_list(columns)
+            frame_columns = set(self.frame.columns)
             columns = [
-                reverse_mapping[col] if col in reverse_mapping else col
+                reverse_mapping[col]
+                if col in reverse_mapping and reverse_mapping[col] in frame_columns
+                else col
                 for col in columns
             ]
             columns = [col for col in self.frame.columns if col in columns]
@@ -2353,12 +2356,16 @@ class Head(Expr):
     def _simplify_down(self):
         if isinstance(self.frame, Elemwise):
             operands = [
-                Head(op, self.n, self.npartitions) if isinstance(op, Expr) else op
+                Head(op, self.n, self.operand("npartitions"))
+                if isinstance(op, Expr)
+                else op
                 for op in self.frame.operands
             ]
             return type(self.frame)(*operands)
         if isinstance(self.frame, Head):
-            return Head(self.frame.frame, min(self.n, self.frame.n), self.npartitions)
+            return Head(
+                self.frame.frame, min(self.n, self.frame.n), self.operand("npartitions")
+            )
 
     def _simplify_up(self, parent, dependents):
         from dask_expr import Repartition
--- a/dask_expr/_expr.py
+++ b/dask_expr/_expr.py
@@ -3607,7 +3607,7 @@ class Fused(Blockwise):
 
     @functools.cached_property
     def _name(self):
-        return f"{str(self)}-{_tokenize_deterministic(self.exprs)}"
+        return f"{str(self)}-{_tokenize_deterministic(*self.operands)}"
 
     def _divisions(self):
         return self.exprs[0]._divisions()
--- a/dask_expr/_expr.py
+++ b/dask_expr/_expr.py
@@ -1212,7 +1212,20 @@ class Elemwise(Blockwise):
 
 class RenameFrame(Elemwise):
     _parameters = ["frame", "columns"]
-    _preserves_partitioning_information = True
+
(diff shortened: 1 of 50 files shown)


dask_expr/_merge.py
```diff
@@ -152,6 +152,13 @@ def _get_original_predicate_columns(self, predicate):
     def __str__(self):
         return f"Merge({self._name[-7:]})"
 
+    @property
+    def unique_partition_mapping_columns(self):
+        return {
+            self.left_on,
+            self.right_on,
```
fjetter, line 159: this may not be correct. this PR is probably riddled with bugs

fjetter:
This can be extended to a much more useful property if we encode knowledge about this mapping itself. For instance, the unique mapping for a shuffle is a hash on a set of columns. Whenever two datasets are encountered that were using the same mapping this can even be used to have trivial merges.

This topic is being discussed over here https://github.com/dask/dask/issues/9425

phofl:
I updated your implementation to drop shuffles from merges as well and propagate the mapping through blockwise and similar ops. This is a rough implementation for now

We would need a lot more tests among other things if we want to do this now

dask_expr/io/parquet.py
```diff
@@ -773,6 +773,19 @@ def approx_statistics(self) -> dict:
...
+        # TODO This is a poor name
+        if self.calculate_divisions:
+            return {
+                col
+                for col in self.columns
+                if _divisions_from_statistics(self.aggregated_statistics, col)[1]
+                is not None
+            }
+        else:
+            return set()
```
fjetter, line 787: Note that this information is not sufficient to perform merges. It is sufficient to perform trivial groupby+apply operations but not to join two frames unless the statistics between the two frames are identical. 
It looks like your merge implementation doesn't distinguish this. We may need to propagate two distinct properties

dask_expr/_merge.py
```diff
@@ -314,6 +321,16 @@ def merge_indexed_right(self):
...
 
+    def _on_condition_alread_partitioned(self, expr, on):
+        if not isinstance(on, list):
+            result = (
+                on in expr.unique_partition_mapping_columns
+                or (on,) in expr.unique_partition_mapping_columns
+            )
+        else:
+            result = tuple(on) in expr.unique_partition_mapping_columns
+        return result and expr.npartitions == self.npartitions
```
hendrikmakait, line 332: IIUC, whether this works depends on the order of `on` and tuples in `unique_partition_mapping_columns`. We should make this order-independent.

hendrikmakait: See https://github.com/coiled/benchmarks/pull/1471#discussion_r1528086910

dask_expr/io/parquet.py
phofl: I removed this for now, since that's not relevant at the moment for benchmarks. But yeah I think we have to use a different property then

dask_expr/_merge.py
phofl: It is already order dependent, but I'll add a test to test for this

phofl: That would need a bit of care and goes a little into join reordering, so want to keep this out for now

phofl:
Update here: The correctness tpch tests are green on scale 100, so we *should* be good to go

dask_expr/_expr.py
```diff
@@ -407,6 +407,10 @@ def columns(self) -> list:
         except Exception:
             raise
 
+    @functools.cached_property
+    def unique_partition_mapping_columns(self):
+        return set()
```
fjetter, line 412: Very high level, I think the name doesn't make any sense any longer. I started this with this name encoding a different information.

I would also like to see a doc string here describing this property because this is highly nontrivial, I think.

What I initially planned is this

```python
    @functools.cached_property
    def unique_partition_mapping_columns(self):
        """A set of tuples of columns that identify a unique, bijective mapping from values to partition

        Example::

            unique_partition_mapping_columns = {
                ("A", "B"),
                "C",
            }

            Partition 1
            | A | B | C | D |
            |---|---|---|---|
            | 1 | 1 | 1 | 1 |
            | 1 | 2 | 1 | 2 |
            | 1 | 3 | 1 | 3 |

            Partition 2
            | A | B | C | D |
            |---|---|---|---|
            | 1 | 4 | 2 | 1 |
            | 1 | 5 | 3 | 2 |
            | 2 | 3 | 4 | 3 |
        """
        return set()
```

but this is not sufficient to remove unnecessary merges. I think there is some implicit information encoded in here that we should make explicit.

Is it...

- All columns that satisfy the above property and were created by a shuffle?

or is it something else? Because if it's the shuffle thing, I don't think this can be a set but there is always exactly one tuple (we know of) that satisfies this condition

dask_expr/_shuffle.py
```diff
@@ -91,6 +93,18 @@ def __str__(self):
     def _node_label_args(self):
         return [self.frame, self.partitioning_index]
 
+    @functools.cached_property
+    def _partitioning_index(self):
+        partitioning_index = self.partitioning_index
+        if isinstance(partitioning_index, (str, int)):
```
fjetter, line 99: I assume this is a check for whether this index is a column?
```suggestion
        if partitioning_index in self.frame.columns:
```

fjetter (commented):
The implementation as such looks OK as far as I can tell but we should really make sure that the semantics are clear for these columns

I wrote down a similar thing over in https://github.com/dask/dask/issues/9425#issuecomment-1396729930 before. 
When I started this PR I implemented case 2) by using the parquet statistics. Case 2) is **not** sufficient for merges.
I believe you started implementing case 1) with the assumption that it is always the same function/mapping that generated the partitioning, i.e. the shuffle. Is this correct?

dask_expr/_expr.py
phofl: > is it something else? Because if it's the shuffle thing, I don't think this can be a set but there is always exactly one tuple (we know of) that satisfies this condition

It is a shuffle, but also a merge based shuffle. And merge based shuffle have 2 tuples, the columns that were shuffled on on the left and the corresponding columns on the right. Plain shuffles will always be 1 tuple, that's correct.

dask_expr/_shuffle.py
phofl: partitioning_index could be an expression, and in with expressions doesn't work the way we need it here

phofl:
Yes I only implemented case 1 (and dropped case 2 after you pointed it out to me), I think case 1 is more helpful for us at the moment because we can use it by default (the statistics thing needs explicit action from the user).

> For case 1) one would typically also want to remember which function was used to generate this partitioning since that has to match as well for joins to be trivial. Maybe this is follow up work but we should at least make it explicit in docs/function names

I'll improve the docs either later today or tomorrow, but we now use the same partitioning function for all shuffles, so that should be fine from an implementation perspective, we need to document it though, that's correct.

dask_expr/_expr.py
fjetter: ok, that makes sense.

In this case I would name this property to something like `unique_column_mapping_from_shuffle`

fjetter:
> I think case 1 is more helpful 

Yes, of course. Case 2 can only do trivial single-df operations. Case 1 includes Case 2 trivially but we have to shuffle. The nice thing about my initial stab was that it could be just inferred from parquet data so `read_parquet("foo").groupby(...)` could work by just reading parquet metadata. I think this is also something we'll want to implement.

> I'll improve the docs either later today or tomorrow, but we now use the same partitioning function for all shuffles, so that should be fine from an implementation perspective, we need to document it though, that's correct.

Yes, I just want to make sure we understand this moving forward. Especially

1. There is a case (the parquet metadata) which looks very similar but isn't. We have to distinguish this.
2. While the shuffle currently is unique and used almost everywhere, there are other mechanisms to create a unique partitioning with the properties of Case 2.

dask_expr/_shuffle.py
```diff
@@ -804,6 +819,10 @@ class SetIndex(BaseSetIndexSortValues):
     }
     _filter_passthrough = True
 
+    @functools.cached_property
+    def unique_partition_mapping_columns(self):
+        return {tuple(self.other.columns)}
```
fjetter, line 824: Coming back to my argument about how this condition can be generated from different functions... I am not sure if this is actually correct.

IIUC set_index works by calculating the divisions and the sorting the partitions accordingly. This will create a different data distribution than if we were to shuffle on a couple of columns

phofl: Yep that's correct, set-index works differently (and is not included in this implementation fwiw).

We probably need a more general purpose propagation implementation. All methods that propagate the current mapping will also propagate the set_index or parquet mappings, but that can be a follow up when we add those

fjetter:
Test failures

```
FAILED dask_expr/tests/test_reductions.py::test_unique_numerical_columns[0] - TypeError: 'int' object is not iterable
FAILED dask_expr/tests/test_reductions.py::test_unique_numerical_columns[1] - TypeError: 'int' object is not iterable
FAILED dask_expr/tests/test_reductions.py::test_unique_numerical_columns[2] - TypeError: 'int' object is not iterable
```

```python-traceback
dask_expr/tests/test_reductions.py:530: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
dask_expr/tests/_util.py:29: in assert_eq
    if hasattr(obj, "dask"):
dask_expr/_collection.py:492: in dask
    return self.__dask_graph__()
dask_expr/_collection.py:473: in __dask_graph__
    out = out.lower_completely()
dask_expr/_core.py:436: in lower_completely
    new = expr.lower_once()
dask_expr/_core.py:393: in lower_once
    out = expr._lower()
dask_expr/_reductions.py:526: in _lower
    elif not self.need_to_shuffle:
../../../micromamba/envs/dask-expr/lib/python3.9/functools.py:993: in __get__
    val = self.func(instance)
dask_expr/_reductions.py:474: in need_to_shuffle
    for cols in self.frame.unique_partition_mapping_columns
../../../micromamba/envs/dask-expr/lib/python3.9/functools.py:993: in __get__
    val = self.func(instance)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = df[0]

    @functools.cached_property
    def unique_partition_mapping_columns(self):
>       columns = set(self.operand("columns"))
E       TypeError: 'int' object is not iterable
```

phofl:
Thx! I'll follow up with better naming and docs

PR #950, opened 2024-03-07
Implement string conversion for from_array

### LINKED ISSUE #691: Implement convert string option in some capacity
I think the best way forward in dask-expr is to enable the infer-string option from pandas to offload this stuff

phofl:
closes #691


milesgranger (approved):
A thing of beauty! Thanks, @phofl!

phofl:
dask/dask pr is here: https://github.com/dask/dask/pull/10988

PR #951, opened 2024-03-07
Add dtype and columns type check for shuffle

phofl:


dask_expr/tests/test_distributed.py
```diff
@@ -356,6 +357,19 @@ def test_func(n):
     pd.testing.assert_frame_equal(result, expected)
 
 
+@gen_cluster(client=True)
+async def test_future_in_map_partitions(c, s, a, b):
```
hendrikmakait, line 361: The test name seems wrong.

phofl: very wrong, thx

phofl:
dask/dask ci is expected

hendrikmakait (approved):
Thanks, @phofl!

PR #956, opened 2024-03-08
MINOR: Avoid confusion around shuffle method

hendrikmakait:
Maybe this helps someone avoid getting confused.

### DIFF
--- a/dask_expr/_collection.py
+++ b/dask_expr/_collection.py
@@ -783,7 +783,7 @@ Expr={expr}"""
         elif pd.api.types.is_list_like(on) and not is_dask_collection(on):
             on = list(on)
 
-        if shuffle_method or get_default_shuffle_method() == "p2p":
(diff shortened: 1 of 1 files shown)


dask_expr/_collection.py
```diff
@@ -783,8 +783,7 @@ def shuffle(
         elif pd.api.types.is_list_like(on) and not is_dask_collection(on):
             on = list(on)
 
-        method = shuffle_method or get_default_shuffle_method()
-        if method == "p2p":
+        if shuffle_method or get_default_shuffle_method() == "p2p":
```
phofl, line 786:
```suggestion
        if (shuffle_method or get_default_shuffle_method()) == "p2p":
```


This is incorrect actually, == should bind stronger than or

phofl (approved):

phofl:
thx

PR #961, opened 2024-03-08
Adjust split_out values to be consistent with other methods

phofl:
Running AB tests for this at the moment


phofl:
The new constant 10 over 20 seems to be an improvement

AB here: https://github.com/coiled/benchmarks/actions/runs/8206111172

Screenshot 2024-03-08 at 23 59 49

PR #962, opened 2024-03-08
Change split names in task based shuffle

phofl:
cc @crusaderky 

Potential fix, but I don't know task shuffles well enough to judge if this is a bad idea


crusaderky:
- Orthogonal to https://github.com/dask/distributed/pull/8563

phofl:
Wrong variable, sorry. Using the correct variable that is also used in dask/dask creates cycles. We are building the graph differently here. So closing this since it's not urgent.

phofl:
> Not sure this was the intent?

No that wasn't the intent, you are correct

PR #943, opened 2024-03-07
Add _repr_html_ and prettier __repr__ w/o graph materialization

milesgranger:
Originally part of #913, fixed by #930 which caused materialization of the graph, reverted by #941

This also updates `to_html` method which has the same issue before #930.

PR #963, opened 2024-03-11
Fix 'Empty' prefix to non-empty Series repr

milesgranger:
https://github.com/dask-contrib/dask-expr/pull/943#discussion_r1518975367


phofl (approved):
Fwiw a DataFrame without columns can be non empty too (strictly speaking), but that’s probably a corner case

phofl:
Thx

PR #964, opened 2024-03-11
Update README.md

fjetter:
Considering that we're switching the default, we should remove the overly defensive phrasing.


milesgranger (approved):
Wording seems good to me. :+1:

phofl (approved):
Lgtm as well, is GitHub actions down?

phofl:
Thx

PR #971, opened 2024-03-13
Fix unique with shuffle and strings

phofl:
closes https://github.com/dask/dask/issues/11001#issuecomment-1995244204


milesgranger (approved):
LGTM, thanks @phofl!

PR #973, opened 2024-03-14
Fixup predicate pushdown for query 19

phofl:
closes https://github.com/coiled/benchmarks/issues/1458

### DIFF
--- a/dask_expr/_merge.py
+++ b/dask_expr/_merge.py
@@ -107,7 +107,7 @@ class Merge(Expr):
                 return False
             return True
         elif isinstance(parent.predicate, And):
-            # If we can make that transformation than we should do it to further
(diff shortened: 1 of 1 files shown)


dask_expr/_merge.py
```diff
@@ -106,6 +106,13 @@ def _filter_passthrough_available(self, parent, dependents):
             elif len(predicate_columns) > 0:
                 return False
             return True
+        elif isinstance(parent.predicate, And):
+            # If we can make that transformation than we should do it to further
+            # align filters that sit on top of merges
```
milesgranger, line 111: nit

```suggestion
            # If we can make that transformation then we should do it to further
            # align filters that sit on top of merges
```

milesgranger (approved):
Very nice! Thanks, @phofl!

PR #981, opened 2024-03-15
Fix concat of series objects with column projection

phofl:
ref https://github.com/dask/dask-expr/issues/904


milesgranger (approved):
Thanks, @phofl!

PR #984, opened 2024-03-18
Ensure wrapping an array when comparing to Series works if columns are empty

fjetter:
Closes https://github.com/dask/dask/issues/11006


hendrikmakait (approved):
Thanks, @fjetter!

PR #987, opened 2024-03-19
Rename overloaded `to/from_dask_dataframe` API

rjzamora:
Closes https://github.com/dask/dask-expr/issues/905

### DIFF
--- a/dask_expr/_collection.py
+++ b/dask_expr/_collection.py
@@ -1261,9 +1261,9 @@ Expr={expr}"""
             FutureWarning,
             "`to_dask_dataframe` is deprecated, please use `to_legacy_collection`.",
         )
-        return self.to_legacy(*args, **kwargs)
+        return self.to_legacy_collection(*args, **kwargs)
 
-    def to_legacy(self, optimize: bool = True, **optimize_kwargs) -> _Frame:
+    def to_legacy_collection(self, optimize: bool = True, **optimize_kwargs) -> _Frame:
         """Convert to a legacy dask-dataframe collection
 
         Parameters
--- a/dask_expr/_collection.py
+++ b/dask_expr/_collection.py
@@ -1258,8 +1258,8 @@ Expr={expr}"""
             Key-word arguments to pass through to `optimize`.
         """
         warnings.warn(
-            FutureWarning,
             "`to_dask_dataframe` is deprecated, please use `to_legacy_collection`.",
+            FutureWarning,
         )
         return self.to_legacy_collection(*args, **kwargs)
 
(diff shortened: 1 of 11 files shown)


rjzamora:
@phofl - I've been finding the overloaded/conflicting meanings of `to/from_dask_dataframe` in `dask.array`/`dask.bag`/`dask-cudf`/etc to be a bit confusing. Hopefully you are still okay with renaming.

You mentioned earlier that we will also need to make changes in dask/dask. Does it make sense to leave the deprecation warning out until we are confident that dask/dask is using the "correct" name?

dask_expr/_collection.py
```diff
@@ -1245,8 +1245,26 @@ def repartition(
...
+        **optimize_kwargs
+            Key-word arguments to pass through to `optimize`.
+        """
+        warnings.warn(
+            "`to_dask_dataframe` is deprecated, please use `to_legacy_collection`.",
+            FutureWarning,
+        )
+        return self.to_legacy_collection(*args, **kwargs)
+
+    def to_legacy_collection(self, optimize: bool = True, **optimize_kwargs) -> _Frame:
```
phofl, line 1266: This is too internal imo, lets just use to_legacy_dataframe

dask_expr/_collection.py
```diff
@@ -4705,8 +4723,25 @@ def from_dict(
...
+        Whether to optimize the graph before conversion.
+    """
+    warnings.warn(
+        "`from_dask_dataframe` is deprecated, please use `from_legacy_collection`.",
+        FutureWarning,
+    )
+    return from_legacy_collection(*args, **kwargs)
+
+
+def from_legacy_collection(ddf: _Frame, optimize: bool = True) -> FrameBase:
```
phofl, line 4743: Same here

phofl:
Yeah leaving the warning out until the next dask/dask release happens would be good

dask_expr/_collection.py
rjzamora: Yeah, I've always been a bit bothered by the (lack of?) Series/Dataframe distinction, but I agree that collection is probably worse. will use `to_legacy_dataframe`

rjzamora:
Thanks for the review @phofl!  I addressed your comments.

phofl (approved):

phofl:
thx

rjzamora:
>Can you update the following things in dask/dask:
> - Add the new method to the api reference
> - replace usages of to_dask_dataframe/from_dask_dataframe with the new methods
>
>And create a PR here that adds the warning so that we can merge it before the next dask/dask release happens?

Sounds good. Thanks!

PR #990, opened 2024-03-21
Fix SettingWithCopyWarning in _merge.py

milesgranger:
Fix dask/distributed#8591

### DIFF
--- a/dask_expr/_merge.py
+++ b/dask_expr/_merge.py
@@ -771,7 +771,7 @@ def create_assign_index_merge_transfer():
         index_merge,
     ):
         if index_merge:
-            index = df[[]]
+            index = df[[]].copy()
             index["_index"] = df.index
         else:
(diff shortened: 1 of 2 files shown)


fjetter:
It might make sense to enable stricter filterwarnings in [pyproject.toml](https://github.com/dask/dask-expr/blob/b6696ee5d770cdae1824ce5a50c004e46ec21810/pyproject.toml#L69-L72). In dask and distributed we're raising on every warning unless the warning has been explicitly allowed. I think making this push for this PR is a little too much but adding this specific warning would make sense to me.

milesgranger:
Okay, so I ought to add such to distributed in order to fix https://github.com/dask/distributed/issues/8591?

fjetter:
no, distributed is already raising. I'm suggesting to add this to dask-expr in case there are other places where we're raising this warning but aren't noticing

PR #992, opened 2024-03-22
Ensure drop matches column names exactly

fjetter:
This makes me think that the core `Expr` might want to support something like argument normalization. Not pushing for this but if more cases like this pop up, that might help.


milesgranger (approved):
LGTM, thanks @fjetter!

PR #997, opened 2024-03-25
Set divisions with divisions already known

fjetter:
Closes https://github.com/dask/dask/issues/11020

The `compute_current_divisions` kwarg `set_divisions` appears to be unused, at least in dask-expr. From an API perspective I find this a little weird and we should consider deprecating this


phofl (approved):

phofl:
thx

PR #998, opened 2024-03-25
Nicer read_parquet prefix

fjetter:
https://github.com/dask/dask-expr/issues/989 discusses how this prefix is a little too verbose. For me as a developer it is quite important to know which of the two implementations is currently running but ordinary users will likely not care.

This would be a simple approach to just include the logical name.


mrocklin:
Thanks!

PR #1011, opened 2024-04-03
Make setattr work

### LINKED ISSUE #1010: Assigning a column by attribute does not work
Came across a subtle bug/change in dask-expr while debugging a dask-cudf test:

```python
import dask.dataframe as dd

df = dd.from_dict({"a": [0, 1]}, 1)
df.dtypes.a  # dtype('int64')

df.a = df.a.astype("float")
df.dtypes.a  # `meta` is NOT updated: dtype('int64')

df["a"] = df.a.astype("float")
df.dtypes.a  # `meta` IS updated: dtype('float64')
```

phofl:
closes #1010

### DIFF
--- a/dask_expr/_collection.py
+++ b/dask_expr/_collection.py
@@ -2894,6 +2894,24 @@ class DataFrame(FrameBase):
         out = self[columns]
         self._expr = out._expr
 
+    def __setattr__(self, key, value):
+        try:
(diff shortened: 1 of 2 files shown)


rjzamora (approved):
Thanks for the quick fix @phofl - LGTM

PR #1017, opened 2024-04-09
Fix unique with numeric columns

### LINKED ISSUE #1015: Incorrect ``unique`` result for column with numerical name
I'm getting an incorrect `unique` result after selecting a numerical column name from a `DataFrame` collection:

```python
import pandas as pd
import dask.dataframe as dd

pdf = pd.DataFrame({1: [0,3,0,1,2,0,3,1,1]})
df = dd.from_pandas(pdf, 3)

df[key].unique().compute()
```
```
0    0
1    1
2    3
0    3
1    2
2    0
3    1
Name: 1, dtype: int64
```

I haven't had time to dig into this yet, but it seems like things work fine when the column name is `0`. Also, changing the shuffle algorithm doesn't seem to resolve the issue.

phofl:
closes #1015

### DIFF
--- a/dask_expr/_reductions.py
+++ b/dask_expr/_reductions.py
@@ -173,9 +173,17 @@ class ShuffleReduce(Expr):
         from dask_expr._shuffle import RearrangeByColumn, SetIndexBlockwise, SortValues
 
         if is_index_like(self.frame._meta):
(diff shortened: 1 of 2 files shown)


rjzamora:
Thanks @phofl !

PR #1028, opened 2024-04-16
Fix backend dispatching for `read_csv`

rjzamora:
Backend dispatching doesn't currently work in all cases for `read_cudf`, because the `operation` property may be accessed outside a `config` context block. This PR resolves the issue by adding a `"dataframe_backend"` parameter.

### DIFF
--- a/dask_expr/_collection.py
+++ b/dask_expr/_collection.py
@@ -13,7 +13,7 @@ import dask.dataframe.methods as methods
 import numpy as np
 import pandas as pd
 import pyarrow as pa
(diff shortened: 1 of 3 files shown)


rjzamora:
Any objection here @phofl ?

Note that it's a bit simpler for rapids than marking the entire top-level `read_csv` function as "dispatchable" for now. However, that is exactly what we will need to do when the internal legacy-API fallback is removed.

phofl (approved):

phofl:
thx

PR #1029, opened 2024-04-16
Fuse more aggressively if parquet files are tiny

phofl:
This PR does a few things:

- Fuse tiny parquet files more aggressive up to 75MB per partition (in memory)
- set split_out in groupby also for length one grouping keys (this is more conservative and helps uf get rid of a few split_out=True optimisations in benchmarks)
- make isin work with npartitions=1 Dask Series objects

Here is an AB test for this (fuse tag)

https://github.com/coiled/benchmarks/actions/runs/8709180542

### DIFF
--- a/dask_expr/_groupby.py
+++ b/dask_expr/_groupby.py
@@ -86,7 +86,7 @@ def _as_dict(key, value):
 
 def _adjust_split_out_for_group_keys(npartitions, by):
     if len(by) == 1:
-        return math.ceil(npartitions / 10)
+        return math.ceil(npartitions / 15)
     return math.ceil(npartitions / (10 / (len(by) - 1)))
 
 
--- a/dask_expr/_groupby.py
+++ b/dask_expr/_groupby.py
@@ -85,6 +85,8 @@ def _as_dict(key, value):
 
 
 def _adjust_split_out_for_group_keys(npartitions, by):
+    if len(by) == 1:
+        return math.ceil(npartitions / 10)
     return math.ceil(npartitions / (10 / (len(by) - 1)))
 
 
@@ -222,7 +224,7 @@ class GroupByApplyConcatApply(ApplyConcatApply, GroupByBase):
         return self.frame.columns
 
     def _tune_down(self):
-        if len(self.by) > 1 and self.operand("split_out") is None:
+        if self.operand("split_out") is None:
             return self.substitute_parameters(
                 {
                     "split_out": functools.partial(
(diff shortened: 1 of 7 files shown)


dask_expr/_groupby.py
```diff
@@ -674,7 +676,7 @@ class GroupByReduction(Reduction, GroupByBase):
     _chunk_cls = GroupByChunk
 
     def _tune_down(self):
-        if len(self.by) > 1 and self.operand("split_out") is None:
+        if self.operand("split_out") is None:
```
fjetter, line 679: Wouldn't we always shuffle now?

dask_expr/io/parquet.py
```diff
@@ -1001,17 +1001,16 @@ def fragments_unsorted(self):
...
         total_uncompressed = 0
         after_projection = 0
-        col_op = self.operand("columns")
+        col_op = self.operand("columns") or self.columns
         for col in approx_stats["columns"]:
             total_uncompressed += col["total_uncompressed_size"]
             if col["path_in_schema"] in col_op:
                 after_projection += col["total_uncompressed_size"]
 
+        total_uncompressed = max(total_uncompressed, 75_000_000)
```
fjetter, line 1013: I suggest to expose this as a config value

dask_expr/_groupby.py
phofl: Yep

phofl: except if split_out=1 is set explicitly

fjetter: We just had a conversation about this and agreed that we'll go for this automatic behavior. This means that some group by operations will perform a bit worse since we are forcing a shuffle that is not strictly necessary.
For large output results the shuffle is a necessity and for tiny output results, the additional shuffle step only adds a marginal performance penalty in our testing.

It is a safer choice and most users will not want to or be able to dig in deep enough to set this parameter such that this is a good default choice.

dask_expr/io/parquet.py
phofl: https://github.com/dask/dask/pull/11052

phofl: I'll write docs for this tomorrow or later today after I finished the blog (we need release notes anyway for the trivial shuffles)

dask_expr/io/parquet.py
```diff
@@ -821,6 +821,8 @@ def sample_statistics(self, n=3):
         ixs = []
         for i in range(0, nfrags, stepsize):
             sort_ix = finfo_argsort[i]
+            # TODO: This is crude but the most conservative estimate
+            sort_ix = sort_ix if sort_ix < nfrags else 0
```
phofl, line 825: see

PR #1030, opened 2024-04-17
Add nr of columns to explain output for projection

phofl:
this was helpful for me during debugging

### DIFF
--- a/dask_expr/diagnostics/_explain.py
+++ b/dask_expr/diagnostics/_explain.py
@@ -86,7 +86,7 @@ def _explain_details(expr: Expr):
         details["path"] = expr.path
     elif isinstance(expr, Projection):
         columns = expr.operand("columns")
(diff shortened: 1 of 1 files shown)


dask_expr/diagnostics/_explain.py
```diff
@@ -84,6 +84,9 @@ def _explain_details(expr: Expr):
         details["how"] = expr.how
     elif isinstance(expr, ReadParquet):
         details["path"] = expr.path
+    elif isinstance(expr, Projection):
+        columns = expr.operand("columns")
+        details["nr_columns"] = len(columns) if isinstance(columns, list) else "Series"
```
hendrikmakait, line 89: nit:

```suggestion
        details["ncolumns"] = len(columns) if isinstance(columns, list) else "Series"
```

hendrikmakait (approved):
Thanks, @phofl!

PR #1033, opened 2024-04-19
Move IO docstrings over

phofl:
This is mostly copy-paste, we should especially update read-parquet after pycon


hendrikmakait (approved):
Copy+paste is fine for me as a start.

> we should especially update read-parquet after pycon
Could you add an issue for that?

PR #1041, opened 2024-04-25
Fix bug in ``Series`` reductions

rjzamora:
After reducing a `Series`, we need to convert back from the temporary `"__series__"` name. This is currently breaking several tests in `cugraph`.


phofl (approved):

phofl:
thx

PR #1044, opened 2024-04-26
Fix default name conversion in `ToFrame`

rjzamora:
Possible fix for a subtle optimization bug that shows up when an unnamed `Series` is shuffled and then converted to a `DataFrame` and merged. Definitely a bit of a "corner case", but does show up in `cugraph` CI.

### DIFF
--- a/dask_expr/_expr.py
+++ b/dask_expr/_expr.py
@@ -1622,6 +1622,7 @@ class ToFrame(Elemwise):
             return set(self.columns)
         return unique_mapping
 
+
 class ToFrameIndex(Elemwise):
     _parameters = ["frame", "index", "name"]
     _defaults = {"name": no_default, "index": True}
--- a/dask_expr/_expr.py
+++ b/dask_expr/_expr.py
@@ -1613,8 +1613,14 @@ class ToFrame(Elemwise):
     _keyword_only = ["name"]
     operation = M.to_frame
     _filter_passthrough = True
-    _preserves_partitioning_information = True
 
+    @functools.cached_property
+    def unique_partition_mapping_columns_from_shuffle(self):
+        unique_mapping = self.frame.unique_partition_mapping_columns_from_shuffle
+        if set(self.frame.columns) == unique_mapping:
+            # Account for default name conversion
+            return set(self.columns)
+        return unique_mapping
 
 class ToFrameIndex(Elemwise):
     _parameters = ["frame", "index", "name"]
--- a/dask_expr/tests/test_merge.py
+++ b/dask_expr/tests/test_merge.py
@@ -1011,3 +1011,19 @@ def test_merge_suffix_projections():
     expected = df.merge(df, on="a")
     expected = expected[expected["c_x"] == "A"]["c_y"]
     assert_eq(result, expected)
+
+
+def test_merge_after_rename():
+    pleft = pd.Series(range(10))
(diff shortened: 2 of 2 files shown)


dask_expr/tests/test_merge.py
```diff
@@ -1011,3 +1011,18 @@ def test_merge_suffix_projections():
...
+    pleft.columns = ["a"]
+
+    left = from_pandas(pd.Series(range(10)), npartitions=2)
+    left = left.drop_duplicates().to_frame()
+    left.columns = ["a"]
+
+    right = pd.DataFrame({"a": [1, 2] * 5})
+    expected = pleft.merge(right, how="inner")
+    result = left.merge(right, how="inner")
+    assert_eq(result, expected)
```
rjzamora, line 1028:
```suggestion
    assert_eq(result, expected, check_index=False)
```

dask_expr/_expr.py
```diff
@@ -1613,7 +1613,14 @@ class ToFrame(Elemwise):
     _keyword_only = ["name"]
     operation = M.to_frame
     _filter_passthrough = True
-    _preserves_partitioning_information = True
+
+    @functools.cached_property
+    def unique_partition_mapping_columns_from_shuffle(self):
+        unique_mapping = self.frame.unique_partition_mapping_columns_from_shuffle
```
phofl, line 1619: I don't see how this test covers the added function here? Could you elaborate?

rjzamora: Yeah, that's a fair question. This problem is still a bit confusing to me :)

When an un-named Series is shuffled, and then converted to a DataFrame, it's `unique_partition_mapping_columns_from_shuffle` result will be something like `{None}` instead of a set containing the **real** (default) column name (`{0}`). This results in a `KeyError` when `RenameFrame` tries to select the `None` key instead of `0`.

There seem to be several ways to avoid the error. However, my intuition was that the root problem is that `ToFrame` must properly account for the name of the column it creates.

phofl: Yes agree, but this check might not properly account for that

```
set(self.frame.columns) == unique_mapping
```

unique_mapping could be a tuple of one column, I think we have to be a bit more elaborate here

rjzamora: Okay, I could use your help here if the current solution is wrong/incomplete. I was thinking that the only case we need to catch here is when we are converting to a dataframe from an unnamed Index or Series, but I didn't dig into the Index case at all. Is that what you have in mind?

>unique_mapping could be a tuple of one column

Doesn't `unique_partition_mapping_columns_from_shuffle` always return a `set`?

rjzamora:
@phofl - Do you have a use case in mind where this still fails? I'd like to make sure this fix (or something better) is included in the next release.

phofl:
For future PRs: we need tests like the one I added if we change the partitioning implementation

rjzamora:
Oh cool - I didn't see `test_partitioning_knowledge.py` before. Thanks for the help here @phofl !

rjzamora:
Hmm - Seems like the new `test_merge_groupby_to_frame` test [is failing](https://github.com/dask/dask-expr/actions/runs/8925315421/job/24513709349?pr=1049) in https://github.com/dask/dask-expr/pull/1049 for 3.9

phofl:
good point, #1052 

That part of the test didn't make much sense

PR #1049, opened 2024-05-01
Add support for ``DataFrame.melt``

rjzamora:
Closes https://github.com/dask/dask-expr/issues/1002
dask/dask component: https://github.com/dask/dask/pull/11088

I'm not really sure if `melt` has been left out of the API for a specific reason. This PR adds basic support using `map_partitions`.

### DIFF
--- a/dask_expr/_collection.py
+++ b/dask_expr/_collection.py
@@ -3988,6 +3988,35 @@ class DataFrame(FrameBase):
             layers=maybe_pluralize(n_expr, "expression"),
         )
 
+    @derived_from(pd.DataFrame)
+    def melt(
+        self,
+        id_vars=None,
+        value_vars=None,
+        var_name=None,
+        value_name="value",
+        col_level=None,
+    ):
+        meta = make_meta(
+            meta_nonempty(self._meta).melt(
+                id_vars=id_vars,
+                value_vars=value_vars,
+                var_name=var_name,
+                value_name=value_name,
+                col_level=col_level,
+            )
+        )
+        return self.map_partitions(
+            M.melt,
+            clear_divisions=True,
+            meta=meta,
+            id_vars=id_vars,
+            value_vars=value_vars,
+            var_name=var_name,
+            value_name=value_name,
+            col_level=col_level,
+        )
+
     def _repr_data(self):
         meta = self._meta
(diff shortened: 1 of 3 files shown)


dask_expr/_collection.py
```diff
@@ -3988,6 +3988,35 @@ def to_html(self, max_rows=5):
...
+    ):
+        meta = make_meta(
+            meta_nonempty(self._meta).melt(
+                id_vars=id_vars,
+                value_vars=value_vars,
+                var_name=var_name,
+                value_name=value_name,
+                col_level=col_level,
+            )
+        )
```
rjzamora, line 4008: Probably don't need this.

dask_expr/tests/test_collection.py
```diff
@@ -2606,3 +2606,30 @@ def test_shape_integer(df):
...
+        {},
+        dict(id_vars="int"),
+        dict(value_vars="int"),
+        dict(value_vars=["obj", "int"], var_name="myvar"),
+        dict(id_vars="s1", value_vars=["obj", "int"], value_name="myval"),
+        dict(value_vars=["obj", "s1"]),
+        dict(value_vars=["s1", "s2"]),
+    ],
+)
+def test_melt(kwargs):
```
rjzamora, line 2623: Test copied from dask/dask for now.

dask_expr/_collection.py
phofl: Can you confirm this? Wouldn't want to merge it if we don't need it.

rjzamora:
Thanks for the review @phofl ! Sorry - Should have marked this as a draft. I was curious why `melt` was never added, and didn't want to bury too much time in a PR only to find out that there are deeper issues I didn't know about. If the answer is: "non of the guiding benchmarks/workflows seemed to need it", then I'll be happy to clean this up and make test coverage is thorough both here and in dask/dask.

dask_expr/_collection.py
rjzamora: Okay yeah. Default emulation works fine for the tests - Removing the unnecessary logic.

phofl (approved):
lgtm

can you add it to the api docs as well if not already done in your other PR?

phofl:
thx

PR #1050, opened 2024-05-02
reduce pickle size of parquet fragments

fjetter:
This is still early stage.

I wondered why our graph is always very large in bytes when using this parquet interface and dug a little. I found that the fragments are indeed quite large. What I found so far

- Something like the partition expression is a pyarrow.compute Expression. Even the vanilla True, i.e. "read all partitions" expression packs already 700KB and this is attached to every file.
- The parquetfileformat object is in a similar vicinity although we're rarely if ever specifiying anything and we could just as well use the default
- Most importantly, though, I discovered that most things we're unpacking manually in this Fragment are ad-hoc wrapped C objects, i.e. every time I invoke `Fragement.filesystem` I get a new python object even though the underlying C object is identical. This confuses pickle. Pickle typically deduplicates objects but with this it doesn't stand a chance.

We may be able to ditch this wrapper for newer pyarrow versions entirely but for old ones we may have to be clever at deduplicating things. I still have to run some tests but I suspect this will reduce graph size by orders of magnitude

### DIFF
--- a/dask_expr/io/parquet.py
+++ b/dask_expr/io/parquet.py
@@ -16,6 +16,7 @@ import dask
 import numpy as np
 import pandas as pd
 import pyarrow as pa
+import pyarrow.compute as pc
 import pyarrow.dataset as pa_ds
 import pyarrow.fs as pa_fs
 import pyarrow.parquet as pq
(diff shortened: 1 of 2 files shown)


dask_expr/io/tests/test_parquet.py
```diff
@@ -561,6 +561,15 @@ async def test_parquet_distriuted(c, s, a, b, tmpdir, filesystem):
     assert_eq(await c.gather(c.compute(df.optimize())), pdf)
 
 
+def test_pickle_size(tmpdir, filesystem):
+    pdf = pd.DataFrame({"x": [1, 4, 3, 2, 0, 5]})
+    [_make_file(tmpdir, df=pdf, filename=f"{x}.parquet") for x in range(10)]
+    df = read_parquet(tmpdir, filesystem=filesystem)
+    from distributed.protocol import dumps
+
+    assert len(b"".join(dumps(df.optimize().dask))) <= 8300
```
fjetter, line 570: this was at 19500 before so this is a reduction in size of about 60%

dask_expr/io/parquet.py
```diff
@@ -111,26 +112,42 @@ def _maybe_adjust_cpu_count():
 
 class FragmentWrapper:
```
fjetter, line 113: This class is a bit wild but it's all well encapsulated. Once this stabilizes we can upstream certain things and the wrapper can fade out. API wise it's still a drop in replacement and this is merely there to facilitate serialization

phofl (approved):
lgtm

phofl:
thx

dask_expr/io/tests/test_parquet.py
hendrikmakait: > this was at 19500 before so this is a reduction in size of about 60%

Did you mean ~60x?

phofl: I don't follow? 19.5k to 8.3k is roughly 60% less

hendrikmakait: Never mind, I typed too many zeros.

PR #1051, opened 2024-05-02
Add a bunch of docs

phofl:


dask_expr/io/parquet.py
```diff
@@ -330,6 +330,115 @@ def to_parquet(
...
+        If False (default), construct data-set from scratch. If True, add new
+        row-group(s) to an existing data-set. In the latter case, the data-set
+        must exist, and the schema must match the input data.
+    overwrite : bool, default False
+        Whether or not to remove the contents of `path` before writing the dataset.
+        The default is False.  If True, the specified path must correspond to
+        a directory (but not the current working directory).  This option cannot
+        be set to True if `append=True`.
+        NOTE: `overwrite=True` will remove the original data even if the current
+        write operation fails.  Use at your own risk.
```
hendrikmakait, line 361: Out of scope: At first glance, these two options appear to be the inverse of one another. That is not the case though. We may want to do a better job at explaining this. (I still don't understand it.)

hendrikmakait (approved):
Thanks, @phofl!

dask_expr/io/parquet.py
phofl: Yeah the parquet docs are ridiculously bad at actually explaining at what every option does

hendrikmakait: I addded https://github.com/dask/dask-expr/issues/1054 as a reminder.

PR #1052, opened 2024-05-02
Fixup failing test

phofl:
cc @rjzamora 

the tests didn't make sense since the indexes don't match after merge, removed the assert_eq now


rjzamora (approved):
Thanks!

PR #1053, opened 2024-05-02
Generalize ``get_dummies``

rjzamora:
- Generalizes `get_dummies`/`GetDummies` to work with `cudf`
- Fixes a minor bug (missing `"sparse"` parameter)

### DIFF
--- a/dask_expr/_dummies.py
+++ b/dask_expr/_dummies.py
@@ -156,6 +156,7 @@ class GetDummies(Blockwise):
         "prefix_sep",
         "dummy_na",
         "columns",
+        "sparse",
         "drop_first",
         "dtype",
     ]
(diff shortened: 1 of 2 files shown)


phofl:
I suppose we can't add a test here?

rjzamora:
>I suppose we can't add a test here?

This is a good question. I updated the test to work with both pandas and cudf.

Not sure if we want to add test coverage for the parameters?

phofl (approved):

phofl:
I think that's fine, it would be covered by the gpu ci, correct?

phofl:
thx

rjzamora:
>it would be covered by the gpu ci, correct?

I'm working on this :)

PR #1059, opened 2024-05-07
Add cache  argument to ``lower_once``

rjzamora:
Adds simple logic to `lower_once` so that we avoid re-lowering an expression (and its children) multiple times.

Addresses the `Expr.dask/compute` component of https://github.com/dask/dask-expr/issues/1055: `ddf.dask` is ~200 times faster for python-only reproducer.

### DIFF
--- a/dask_expr/_collection.py
+++ b/dask_expr/_collection.py
@@ -566,8 +566,8 @@ Expr={expr}"""
     def simplify(self):
         return new_collection(self.expr.simplify())
 
-    def lower_once(self):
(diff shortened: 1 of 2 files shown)


rjzamora:
Any objections to this kind of change @fjetter ?  We are already doing the same thing in `simplify_once`.

charlesbluca:
Opened https://github.com/dask-contrib/dask-sql/pull/1332 to point to this patch to test if this unblocks the hanging tests

fjetter (approved):

PR #1073, opened 2024-05-29
Add `how="leftanti"` support for cudf-backed merge

charlesbluca:
Looks like we should unblocked to support left anti joins when `dataframe.backend="cudf"`, similar to the case in legacy Dask dataframe:

https://github.com/dask/dask/blob/df4de6ea53054790b09006c8ea68ef8725d39025/dask/dataframe/multi.py#L565

Note that like the legacy code, we'll fail somewhere down in the comptutation stack if we try this on CPU - not sure if it makes sense to check the backend if `how="leftanti"` and eagerly raise a `NotImplementedError` if `dataframe.backend != "cudf"`.

cc @rjzamora

### DIFF
--- a/dask_expr/_collection.py
+++ b/dask_expr/_collection.py
@@ -2818,7 +2818,7 @@ class DataFrame(FrameBase):
         Parameters
         ----------
         right: dask.dataframe.DataFrame
-        how : {'left', 'right', 'outer', 'inner', 'leftsemi'}, default: 'inner'
+        how : {'left', 'right', 'outer', 'inner', 'leftsemi', 'leftanti'}, default: 'inner'
             How to handle the operation of the two objects:
 
             - left: use calling frame's index (or column if on is specified)
@@ -2830,6 +2830,9 @@ class DataFrame(FrameBase):
               on is specified) with other frame's index, preserving the order
               of the calling's one
             - leftsemi: Choose all rows in left where the join keys can be found
+              in right. Won't duplicate rows if the keys are duplicated in right.
+              Drops all columns from right. Only supported with 'cudf' backend.
+            - leftanti: Choose all rows in left where the join keys cannot be found
               in right. Won't duplicate rows if the keys are duplicated in right.
               Drops all columns from right.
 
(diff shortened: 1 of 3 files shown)


dask_expr/_collection.py
```diff
@@ -2830,6 +2830,9 @@ def merge(
               on is specified) with other frame's index, preserving the order
               of the calling's one
             - leftsemi: Choose all rows in left where the join keys can be found
+              in right. Won't duplicate rows if the keys are duplicated in right.
+              Drops all columns from right. Only supported with 'cudf' backend.
```
rjzamora, line 2834: I believe `"Only supported with 'cudf' backend."` only applies to "leftanti" in dask-expr.

dask_expr/tests/test_merge.py
```diff
@@ -996,6 +996,33 @@ def test_merge_leftsemi():
...
+    df2 = from_pandas(pdf2, npartitions=2)
+    assert_eq(
+        df1.merge(df2, how="leftanti"),
+        pdf1[~pdf1.aa.isin(pdf2.aa)],
+        check_index=False,
+    )
+    df2 = df2.rename(columns={"aa": "dd"})
+    assert_eq(
+        df1.merge(df2, how="leftanti", left_on="aa", right_on="dd"),
+        pdf1[~pdf1.aa.isin(pdf2.aa)],
```
rjzamora, line 1016: Seems like we could just do this in [merge_chunk](https://github.com/dask/dask/blame/3432f4601eb7edc9cd5c2d64fd463d23ab4c3b25/dask/dataframe/multi.py#L244) for pandas data to support `how="leftanti"` for cpu as well.

rjzamora (commented):
Thanks @charlesbluca - Seems reasonable to add "leftanti" support given that the necessary logic is pretty simple, and the legacy `dask.dataframe` API supports it.

dask_expr/_collection.py
charlesbluca: Good catch!

dask_expr/tests/test_merge.py
charlesbluca: Yeah good point, can look into this a bit more

charlesbluca: Pushed some commits to https://github.com/dask/dask/pull/11150 that, in conjunction with this PR, should unblock left anti/semi joins on CPU

PR #1104, opened 2024-07-12
Skip test if optional xarray cannot be imported

### LINKED ISSUE #1102: Please list test dependencies explicitly
The most recent release, `1.1.7`, introduced `xarray` as a dependency for running tests (3997b82). Yet, nowhere is that dependency mentioned, except in ci/environment.yml, which is not an obvious place. Nor is `xarray` a dependency of one of the packages `dask-expr` depends on.

Having test dependencies listed in either a requirements file or in a dedicated extra in `pyproject.toml`, allows our tooling to take care of it instead of manually having to intervene whenever dependencies change.

Alternatively, make tests depending on extra modules soft fail if the module is not available.

penguinpee:
Fix #1102

### DIFF
--- a/dask_expr/array/tests/test_array.py
+++ b/dask_expr/array/tests/test_array.py
@@ -131,6 +131,8 @@ def test_slicing_optimization_change_dimensionality():
 
 
 def test_xarray():
+    pytest.importorskip("xarray")
+
     import xarray as xr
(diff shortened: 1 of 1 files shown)


fjetter (approved):
Thanks!

PR #1107, opened 2024-07-17
Link fix in readme

benrutter:
Heya! I actually forked dask-expr initially to take a look at issue #11232 but the wild speed of @phofl beat me to it 😂

Spotted that the link in the readme doesn't work though, so here's the worlds smallest PR instead.


phofl (approved):

phofl:
Thx!

PR #1114, opened 2024-07-31
Register `read_parquet` and `read_csv` as "dispatchable"

rjzamora:
Registers  the `read_parquet` and `read_csv` definitions in dask-expr for the `"pandas"` backend.

**Some Notes**:
- I am doing this now, while RAPIDS is temporarily pinned to dask-2024.7.1 anyway
- I don't have specific plans to register anything "fancy" for the `"cudf"` backend right away. However, it probably makes sense to move away from the parquet `Engine` approach eventually (it probably makes more sense to lean into the `Expr` class structure to implement backend-specific logic).

### DIFF
--- a/dask_expr/_collection.py
+++ b/dask_expr/_collection.py
@@ -5106,7 +5106,6 @@ def read_csv(
     header="infer",
     dtype_backend=None,
     storage_options=None,
-    _legacy_dataframe_backend="pandas",
     **kwargs,
 ):
     from dask_expr.io.csv import ReadCSV
@@ -5120,7 +5119,7 @@ def read_csv(
             storage_options=storage_options,
             kwargs=kwargs,
             header=header,
-            dataframe_backend=_legacy_dataframe_backend,
+            dataframe_backend="pandas",
         )
     )
 
--- a/dask_expr/_collection.py
+++ b/dask_expr/_collection.py
@@ -14,7 +14,7 @@ import dask.dataframe.methods as methods
 import numpy as np
 import pandas as pd
 import pyarrow as pa
-from dask import compute, config, get_annotations
+from dask import compute, get_annotations
 from dask.array import Array
 from dask.base import DaskMethodsMixin, is_dask_collection, named_schedulers
 from dask.core import flatten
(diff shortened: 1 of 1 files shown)


dask_expr/_collection.py
```diff
@@ -5174,6 +5175,7 @@ def read_fwf(
     )
 
 
+@dataframe_creation_dispatch.register_inplace("pandas")
```
rjzamora, line 5178: For the cudf backend, dask-cudf can register a simple function that calls back into this function with `engine=CudfEngine` (already confirmed that this will work fine). The dask-cudf logic can also check the `filesystem` argument and do something like `read_parquet(..., filesystem="arrow").to_backend("cudf")` (since filesystem="arrow" is currently broken in dask-expr when the backend is `"cudf"`)

dask_expr/_collection.py
```diff
@@ -5099,17 +5099,18 @@ def from_dask_array(x, columns=None, index=None, meta=None):
...
 def read_csv(
     path,
     *args,
     header="infer",
     dtype_backend=None,
     storage_options=None,
+    _legacy_dataframe_backend="pandas",
```
rjzamora, line 5109: I'm open to other ideas. We can copy this entire function into dask-cudf if necessary, but it would be nice to have a simple hook like this while the `read_csv` logic is still so simple.

phofl: I'd prefer not having a top-level argument for uses like this

rjzamora, line 5109:
```suggestion
```

dask_expr/_collection.py
```diff
@@ -5119,7 +5120,7 @@ def read_csv(
             storage_options=storage_options,
             kwargs=kwargs,
             header=header,
-            dataframe_backend=dataframe_backend,
+            dataframe_backend=_legacy_dataframe_backend,
```
rjzamora, line 5123:
```suggestion
            dataframe_backend="pandas",
```

dask_expr/_collection.py
rjzamora: Fair enough. Removed it (Dask cuDF can just copy over the boilerplate code).

phofl (approved):

PR #1115, opened 2024-07-31
Introduce `ToBackend` expression

rjzamora:
Introduces a simple `ToBackend` (and `ToPandasBackend` implementation).

The current approach of using `map_partitions` in `PandasBackendEntrypoint` effectively blocks query-planning optimizations when data is moved between GPU and CPU. The obvious solution is to use a simple `Expr` class.

### DIFF
--- a/dask_expr/_backends.py
+++ b/dask_expr/_backends.py
@@ -5,7 +5,9 @@ import pandas as pd
 from dask.backends import CreationDispatch
 from dask.dataframe.backends import DataFrameBackendEntrypoint
 
+from dask_expr._collection import new_collection
 from dask_expr._dispatch import get_collection_type
+from dask_expr._expr import ToBackend
 
 try:
     import sparse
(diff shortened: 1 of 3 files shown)


rjzamora:
@phofl - This PR is (hopefully) much less controversial than the resource barrier idea in #1116 :)

dask_expr/_backends.py
```diff
@@ -32,25 +33,31 @@
 )
 
 
+class ToPandasBackend(ToBackend):
+    @staticmethod
+    def operation(df, options):
+        from dask.dataframe.dispatch import to_pandas_dispatch
```
phofl, line 39: any reason this is a local import? we import tons of stuff from dask.dataframe globally

dask_expr/_expr.py
```diff
@@ -1303,6 +1303,17 @@ def operation(df):
...
 
+class ToBackend(Elemwise):
+    _parameters = ["frame", "options"]
+    _projection_passthrough = True
+    _filter_passthrough = True
+    _preserves_partitioning_information = True
+
+    @staticmethod
+    def operation(df, options):
+        raise NotImplementedError()
```
phofl, line 1314: Can you set operation = None, this should raise during graph construction, not execution IIRC

dask_expr/_backends.py
rjzamora: Nope - I'll move it to the top level

dask_expr/_expr.py
rjzamora: Sounds good

rjzamora: Okay - Looks like `Elemwise` already has `operation = None` by default. So, I just removed the operation def altogether. Thanks!

phofl (approved):

phofl:
thx

PR #1119, opened 2024-08-12
Fix merging when index name in meta missmatches actual name

### LINKED ISSUE #1113: KeyError when joining two dask dataframes
KeyError: "Only a column name can be used for the key in a dtype mappings argument. '0' not found in columns."
```
import functools
import operator

import dask.array as da
import dask.dataframe as dd
import numpy as np
import pandas as pd
from dask import delayed


def test_func(x, array_location):
    index = np.arange(array_location[0] * array_location[1])
    return pd.DataFrame(index=index, data=dict(value=np.random.random(len(index))))


def _array_chunk_location(block_id, chunks):
    """Pixel coordinate of top left corner of the array chunk."""
    array_location = []
    for idx, chunk in zip(block_id, chunks):
        array_location.append(sum(chunk[:idx]))
    return tuple(array_location)


rng = da.random.default_rng()
array = rng.random((100, 100), chunks=(10, 10))
label_block_iter = zip(
    np.ndindex(*array.numblocks),
    map(
        functools.partial(operator.getitem, array),
        da.core.slices_from_chunks(array.chunks),
    ),
)

meta = dd.utils.make_meta(pd.DataFrame({
    "value": [1],
}, index=np.zeros(1, dtype=np.int64)))
results = []
for block_id, block in label_block_iter:
    array_location = _array_chunk_location(block_id, block)
    results.append(
        dd.from_delayed(
            delayed(test_func)(block, array_location),
            meta=meta,
            verify_meta=False,
        )
    )
ddf = dd.concat(results)
grouped = ddf.groupby(ddf.index, group_keys=False)
df1 = grouped.agg({"value": "sum"})


def partition_func(partition):
    return pd.DataFrame(index=partition.index.copy(), data={"value2": np.random.random(len(partition))})


df2 = df1.map_partitions(partition_func, meta=dd.utils.make_meta(pd.DataFrame({
    "value2": [1],
}, index=np.zeros(1, dtype=np.int64))))
df3 = df1.join(df2).compute()
```

phofl:
closes #1113

### DIFF
--- a/dask_expr/_shuffle.py
+++ b/dask_expr/_shuffle.py
@@ -731,7 +731,9 @@ class AssignPartitioningIndex(Blockwise):
                 if idx.ndim == 1:
                     idx = idx.to_frame()
             elif index_shuffle:
-                idx = obj.index.to_frame()
+                # set default index name, otherwise we will end up with 0
+                name = {"name": idx[0]} if idx == ["_index"] else {}
+                idx = obj.index.to_frame(**name)
             else:
                 idx = _select_columns_or_index(obj, idx)
             return idx
--- a/dask_expr/_shuffle.py
+++ b/dask_expr/_shuffle.py
@@ -732,7 +732,7 @@ class AssignPartitioningIndex(Blockwise):
                     idx = idx.to_frame()
             elif index_shuffle:
                 # set default index name, otherwise we will end up with 0
-                name = {"name": idx[0]} if idx == ["_index"] else {}
+                name = {"name": "_index"} if idx == ["_index"] else {}
                 idx = obj.index.to_frame(**name)
             else:
                 idx = _select_columns_or_index(obj, idx)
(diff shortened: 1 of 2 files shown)


dask_expr/tests/test_merge.py
```diff
@@ -1029,3 +1030,19 @@ def test_merge_after_rename(index):
     expected = pleft.merge(right, how="inner")
     result = left.merge(right, how="inner")
     assert_eq(result, expected, check_index=False)
+
+
+def test_missmatching_meta():
```
hendrikmakait, line 1035:
```suggestion
def test_mismatching_meta():
```

dask_expr/_shuffle.py
```diff
@@ -731,7 +731,9 @@ def _get_index(idx, obj):
                 if idx.ndim == 1:
                     idx = idx.to_frame()
             elif index_shuffle:
-                idx = obj.index.to_frame()
+                # set default index name, otherwise we will end up with 0
+                name = {"name": idx[0]} if idx == ["_index"] else {}
```
hendrikmakait, line 735: nit:
```suggestion
                name = {"name": "_index"} if idx == ["_index"] else {}
```

PR #1122, opened 2024-08-12
Avoid calling ``array`` attribute on ``cudf.Series``

rjzamora:
RAPIDS is currently pinned to 2024.7.1 until [Python-3.9 support is removed](https://github.com/rapidsai/build-planning/issues/88) (hopefully soon). When the pin is removed, `describe` will not work with the "cudf" backend unless we avoid calling `*.array` on a `cudf.Series`.

Note that the minor/breaking change was added in https://github.com/dask/dask-expr/pull/1111


rjzamora:
@phofl - Does this fix seem okay to you?

phofl (approved):

PR #1124, opened 2024-08-16
Make split_out for categorical default smarter

phofl:


dask_expr/_collection.py
```diff
@@ -4235,8 +4235,11 @@ def value_counts(
     ):
         if split_out is no_default:
             if isinstance(self.dtype, CategoricalDtype):
-                # unobserved categories are a pain
-                split_out = 1
+                # unobserved or huge categories will lead to oom errors
+                if self.cat.known:
+                    split_out = max(len(self.dtype.categories) // 100_000, 1)
```
hendrikmakait, line 4240:
```suggestion
                    split_out = len(self.dtype.categories) // 100_000 + 1
```
Otherwise, the value will remain 1 up to 199_999, only then increase every 100_000.

hendrikmakait: You'll also have to adjust the test then.

phofl: That's a nice solution! I though about this briefly but couldn't come up with anything nicer than the max thing

hendrikmakait (approved):
Thanks, @phofl!

PR #1132, opened 2024-09-02
Improve performance of `DelayedsExpr` through caching

hendrikmakait:
This PR fixes a (quadratic?) performance issue and reduces the runtime of 

```python
import dask.dataframe as dd
import dask
import pandas as pd

def construct(num):
    return pd.DataFrame({"x": [num] * 100})

delayed_frames = [dask.delayed(construct)(i) for i in range(4000)]
x = dd.from_delayed(delayed_frames).to_delayed()
```
from 47s to <400ms


phofl (approved):

phofl:
Thx

Issue #1136, opened 2024-09-19
Row-wise Mode Functionality (Dask Issue #11389)


thyripian:
I conducted a brief search to confirm if this has already been opened here.

Dask Issue [#11389 ](https://github.com/dask/dask/issues/11389) requested the addition of row-wise mode functionality for DataFrames. Through extensive research and testing, I found that the necessary modifications had to be made within dask_expr/_collection.py in the dask-expr repository.

I’m submitting this issue to formally document the need for these changes, and I plan to submit a PR shortly to support this. The code changes are already complete and have been thoroughly tested locally—just awaiting approval of the issue before submitting the PR.

PR #1137, opened 2024-09-19
Add Row-Wise Mode Functionality (axis=1) and Improve Metadata Handling in _collection.py

thyripian:
- Added functionality for row-wise mode calculation (axis=1) to support the Dask DataFrame API.
- The new implementation dynamically handles row-wise mode and ensures consistent metadata handling across partitions.
- Added validation for the axis parameter, with appropriate error handling for unsupported values.
- Ensured compatibility with existing column-wise (axis=0) mode functionality, preserving the original behavior for that case.

Resolves Dask Expressions issue #1136 and Dask issue #11389

### DIFF
--- a/dask_expr/_collection.py
+++ b/dask_expr/_collection.py
@@ -3650,13 +3650,47 @@ class DataFrame(FrameBase):
         return new_collection(Query(self, expr, kwargs))
 
     @derived_from(pd.DataFrame)
-    def mode(self, dropna=True, split_every=False, numeric_only=False):
-        modes = []
-        for _, col in self.items():
-            if numeric_only and not pd.api.types.is_numeric_dtype(col.dtype):
-                continue
-            modes.append(col.mode(dropna=dropna, split_every=split_every))
-        return concat(modes, axis=1)
+    # GH#11389 - Dask issue related to adding row-wise mode functionality
+    # GH#1136 - Dask-Expr specific implementation for row-wise mode functionality
+    # Contributor: @thyripian
+    def mode(self, axis=0, numeric_only=False, dropna=True, split_every=False):
+        if axis == 0:
+            # Existing logic for axis=0 (column-wise mode)
+            modes = []
+            for _, col in self.items():
+                if numeric_only and not pd.api.types.is_numeric_dtype(col.dtype):
+                    continue
+                modes.append(col.mode(dropna=dropna, split_every=split_every))
+            return concat(modes, axis=1)
+        elif axis == 1:
+            # Implement axis=1 (row-wise mode)
+            num_columns = len(self.columns)  # Maximum possible number of modes per row
+
+            def row_wise_mode(df):
+                result = df.mode(axis=1, numeric_only=numeric_only, dropna=dropna)
+                # Ensure consistent number of columns across all partitions
+                if result.shape[1] < num_columns:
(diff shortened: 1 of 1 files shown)


thyripian:
Unit tests for the added functionality have been submitted in the Dask repository’s PR due to the initial issue location and the entry point for the functionality being within dask/dataframe/core.py. Please note that this PR in Dask-expr must be approved before the Dask PR tests can successfully run, as the tests depend on the changes made here.

dask_expr/_collection.py
```diff
@@ -3650,13 +3650,46 @@ def query(self, expr, **kwargs):
...
+            for _, col in self.items():
+                if numeric_only and not pd.api.types.is_numeric_dtype(col.dtype):
+                    continue
+                modes.append(col.mode(dropna=dropna, split_every=split_every))
+            return concat(modes, axis=1)
+        elif axis == 1:
+            # Implement axis=1 (row-wise mode)
+            num_columns = len(self.columns)  # Maximum possible number of modes per row
+
+            def row_wise_mode(df):
```
phofl, line 3669: you don't have to do this, we generally expect that the columns in every partition are matching

dask_expr/_collection.py
```diff
@@ -3650,13 +3650,46 @@ def query(self, expr, **kwargs):
...
+                    result = result.iloc[:, :num_columns]
+                # Reindex columns to ensure consistent order
+                result = result.reindex(columns=range(num_columns))
+                # Set column data types to float64 to accommodate NaN values
+                result = result.astype("float64")
+                return result
+
+            # Create metadata with the correct number of columns and float64 dtype
+            meta = pd.DataFrame(
+                {i: pd.Series(dtype="float64") for i in range(num_columns)}
```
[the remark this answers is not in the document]
thyripian: Thank you for highlighting that forcing columns to float64 can lead to unintended type conversions. I initially used float64 to handle NaN values but agree it's better to preserve original data types. I'll update the implementation to dynamically determine data types based on the input, using pandas' nullable types like Int64 for integers and string for text data. This will handle missing values without unnecessary type changes, ensuring consistent data types across partitions while maintaining data integrity. Please let me know if you have further suggestions.

dask_expr/_collection.py
thyripian: Could you clarify which part isn't necessary? Are you referring to determining the number of columns for row-wise mode?

phofl: Everything that is in def row_wise_mode basically

dask_expr/_collection.py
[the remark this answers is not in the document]
phofl: Please don't do any explicit type casting, just use the nonempty meta of the input and call mode on it

dask_expr/_collection.py
thyripian: Understood. I'll make the necessary revisions this afternoon and push them to this PR once tested.

thyripian:
Apologies for the delay — I’ve had some commitments with my day job this past week. The current version has been simplified, with no explicit typecasting or nested functions. It has passed both my local bug script and the pending unit test in the Dask repo.

**Note:** I accidentally pushed my venv with GitHub Desktop earlier, but I’ve since re-pushed with the venv directory removed.

PR #1139, opened 2024-09-21
[WIP] Support ``blocksize`` and ``aggregate_files`` options in ``ReadParquetPyarrowFS``

rjzamora:
**Proposed Changes**

Adds optimization-time ("tune up") support for `blocksize` and `aggregate_files` in `ReadParquetPyarrowFS`.

I like how dask-expr currently "squashes" small files together at optimization time. This PR simply expands the functionality of `_tune_up` to:

1. Split (rather than fuse) oversized parquet files (when `blocksize` is provided; Default is `None` for now).
2. Enable/disable small-file fusion with the `aggregate_files` argument (default to `True` when `filesystem="arrow"`).

**Note on (1)**: This is definitely a **much** higher priority than (2).

**Note on (2)**: `ReadParquetPyarrowFS` already "aggregates" files by default to avoid small partitions. It seems pretty natural to extend this functionality to target a specific `blocksize` value when one is specified. I do think `aggregate_files` is  an intuitive way to enable/disable the "squashing" of small files at optimization time, but I'm okay with leaving this out in favor of the `"dataframe.parquet.minimum-partition-size"` config. With that said, I do think `blocksize` should take precedence over that config value.

**Background**

The legacy `read_parquet` infrastructure is obviously a mess, and I'd like to avoid the need to keep maintaining it (in both dask and rapids). The logic in `ReadParquetPyarrowFS` is slightly arrow-specific, but is already very close to what I was already planning to do in rapids. The only missing feature that is preventing it from supporting real-world use cases is the lack of support for splitting oversized files (a need we definitely run into a lot in the wild).

### DIFF
--- a/dask_expr/_collection.py
+++ b/dask_expr/_collection.py
@@ -5428,13 +5428,14 @@ def read_parquet(
             raise NotImplementedError(
                 "split_row_groups is not supported when using the pyarrow filesystem."
             )
-        if blocksize is not None and blocksize != "default":
+        if blocksize is not None and blocksize != "default" and calculate_divisions:
             raise NotImplementedError(
-                "blocksize is not supported when using the pyarrow filesystem."
+                "blocksize is not supported when using the pyarrow filesystem "
+                "if calculate_divisions is set to True."
             )
-        if aggregate_files is not None:
+        if aggregate_files not in (None, True, False):
             raise NotImplementedError(
-                "aggregate_files is not supported when using the pyarrow filesystem."
+                "aggregate_files must be bool or None when using the pyarrow filesystem."
             )
         if parquet_file_extension != (".parq", ".parquet", ".pq"):
             raise NotImplementedError(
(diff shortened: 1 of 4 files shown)


phofl:
I do have a pretty strong preference on **not** adding keywords that interact with other things counterintuitively. One of the main complexity drivers of the old read_parquet implementation is that there is a plethora of options that disable each other, so I don't want to repeat this here

rjzamora:
>I do have a pretty strong preference on not adding keywords that interact with other things counterintuitively.

Yeah, that makes sense. What do you think is the best way to deal with oversized files? I don't think the `aggregate_files` argument is necessary, but the lack of support for "splitting" large files is a pretty serious blocker right now. Would a `"dataframe.parquet.maximum-partition-size"` config be more palatable?

rjzamora:
**Update**: I moved away from using `blocksize` and `aggregate_files` in favor of `"dataframe.parquet.minimum-partition-size"` (already used) and `"dataframe.parquet.maximum-partition-size"` (new) configs.

PR #1142, opened 2024-10-04
Fix alignment issue with groupby index accessors

phofl:
closes https://github.com/dask/dask/issues/11412


hendrikmakait (approved):
Thanks, @phofl!

PR #1149, opened 2024-10-10
Raise exception when calculating divisons

phofl:
closes https://github.com/dask/dask/issues/11406
closes https://github.com/dask/dask/issues/11318

lets raise here, same as in dask/dask to be able to debug things better


hendrikmakait (approved):
Thanks, @phofl. A test would be helpful to prevent regressions but it appears that we don't test this in `dask/dask` either.

PR #1150, opened 2024-10-11
Fix binary operations with scalar on the left

phofl:
closes https://github.com/dask/dask/issues/11382


hendrikmakait (approved):
Thanks, @phofl!

PR #1151, opened 2024-10-11
Explicitly list setuptools as a build dependency in conda recipe

charlesbluca:
With Python 3.13, `setuptools` has been removed as a default dependency of `pip` on conda-forge, with the encouraged advice to explicitly list this dependency in the build/host section:

https://conda-forge.org/news/2024/08/21/sunsetting-pip-deps/

### DIFF
--- a/ci/recipe/meta.yaml
+++ b/ci/recipe/meta.yaml
@@ -24,7 +24,7 @@ requirements:
     - dask-core {{ dask_version }}
     - versioneer =0.28
     - setuptools >=62.6
-    - tomli
+    - tomli # [py<311]
   run:
     - python >=3.10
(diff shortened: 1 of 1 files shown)


ci/recipe/meta.yaml
```diff
@@ -23,6 +23,7 @@ requirements:
     - pip
     - dask-core {{ dask_version }}
     - versioneer =0.28
+    - setuptools >=62.6
     - tomli
```
charlesbluca, line 27: Small change to roughly match up with the other Dask recipes

```suggestion
    - tomli # [py<311]
```

phofl (approved):

phofl:
Thx

PR #1152, opened 2024-10-16
Fix ``Merge`` divisions after filtering partitions

rjzamora:
A Dask-cuDF user reported an error that suggested something was going wrong in the `Merge` graph:

```
kwargs:    {'how': 'inner', 'indicator': False, 'left_index': False, 'right_index': False, 'suffixes': ('_x', '_y'), 'result_meta': Empty DataFrame
Columns: [id_x, id_y, jaccard, id, uid]
Index: [], 'left_on': ['id_y'], 'right_on': ['id']}
Exception: 'AttributeError("\'tuple\' object has no attribute \'merge\'")'
```

It turns out that the problem happens when you perform a broadcast merge after filtering the partitions of the larger collection. In `main`, we use `_divisions()` to check the divisions of a (possibly-filtered) child expression when I'm pretty sure we need to be using the child's `divisions` property. This PR makes the necessary fix and adds test coverage.

### DIFF
--- a/dask_expr/tests/test_merge.py
+++ b/dask_expr/tests/test_merge.py
@@ -99,9 +99,8 @@ def test_broadcast_merge(how, npartitions):
 
     # Check result with/without fusion
     expect = pdf1.merge(pdf2, on="x", how=how)
-    # TODO: This is incorrect, but consistent with dask/dask
-    assert_eq(df3, expect, check_index=False, check_divisions=False)
-    assert_eq(df3.optimize(), expect, check_index=False, check_divisions=False)
+    assert_eq(df3, expect, check_index=False)
+    assert_eq(df3.optimize(), expect, check_index=False)
 
 
 def test_merge_column_projection():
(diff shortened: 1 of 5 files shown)


dask_expr/tests/test_merge.py
```diff
@@ -1083,3 +1083,17 @@ def test_merge_tuple_left_on():
...
+    b = from_dict({"xx": range(100), "yy": [1, 2] * 50}, npartitions=3)
+
+    result = a.partitions[:5].merge(
+        b, left_on=["y"], right_on=["yy"], how="inner", broadcast=True
+    )
+    expr = result.optimize(fuse=False).expr
+    assert not expr._filtered
+    assert expr.left._filtered
+    assert expr.divisions == expr._divisions()
+    assert len(expr.divisions) == 6
```
phofl, line 1099: can you check the result as well?

rjzamora: Good call.

While adding this check I actually noticed that divisions are often "wrong" after a merge (unrelated to the changes in this PR). More specifically, it seems like we are inheriting divisions after a broadcast join, even if we aren't merging on the index. If I understand correctly, we inherit the partition count in this case, but the divisions are likely to change.

phofl: Yeah that's correct, we probably should only inherit on index merges, pandas is not consistent with the index when you merge on columns :(

phofl (approved):

phofl:
thanks

PR #1159, opened 2024-11-05
Migrate Blockwise to use taskspec

fjetter:
This is migrating the blockwise expressions and connected ones to the new task spec.

I ran into interesting issues while ripping out the recursion in https://github.com/dask/dask-expr/pull/1158 and wanted to do this step here first.

### DIFF
--- a/dask_expr/_expr.py
+++ b/dask_expr/_expr.py
@@ -8,9 +8,9 @@ import warnings
 from collections import defaultdict
 from collections.abc import Callable, Mapping
 
-import dask
 import numpy as np
 import pandas as pd
+from dask._task_spec import Alias, DataNode, Task, TaskRef
 from dask.array import Array
 from dask.core import flatten
 from dask.dataframe import methods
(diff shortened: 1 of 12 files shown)


dask_expr/_expr.py
```diff
@@ -3752,34 +3765,20 @@ def _broadcast_dep(self, dep: Expr):
...
+            if self._broadcast_dep(_expr):
+                subname = (_expr._name, 0)
             else:
-                graph[(_expr._name, index)] = _expr._task(index)
+                subname = (_expr._name, index)
+            subgraphs[subname] = _expr._task(subname, subname[1])
 
         for i, dep in enumerate(self.dependencies()):
-            graph[self._blockwise_arg(dep, index)] = "_" + str(i)
+            subgraphs[self._blockwise_arg(dep, index)] = "_" + str(i)
```
fjetter, line 3778: I have to admit that I don't understand what's going on here but I just went ahead and translated the existing code

phofl (approved):

phofl (changes requested):
the failures look related

fjetter:
hmm, there are problems when running the dask/dask tests with `to_bag` n stuff. That's frustrating

fjetter:
Fix is https://github.com/dask/dask/pull/11495 

I will not change the CI config to test this. I can retrigger once dask/dask is merged

fjetter:
Final problem is https://github.com/dask/dask/pull/11496

the test asserts on the pickled graph size which is blowing up with this since deduplication is not working well. That function caching layer was introduced a little prematurely so let's rip it out and deal with this later

PR #1161, opened 2024-11-11
Fix fusion calling things multiple times

fjetter:
Closes https://github.com/dask/dask/issues/11507

### DIFF
--- a/dask_expr/_expr.py
+++ b/dask_expr/_expr.py
@@ -10,7 +10,7 @@ from collections.abc import Callable, Mapping
 
 import numpy as np
 import pandas as pd
-from dask._task_spec import Alias, DataNode, Task, TaskRef
+from dask._task_spec import Alias, DataNode, Task, TaskRef, execute_graph
 from dask.array import Array
 from dask.core import flatten
 from dask.dataframe import methods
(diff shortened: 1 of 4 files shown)


dask_expr/_expr.py
```diff
@@ -3766,19 +3766,44 @@ def _broadcast_dep(self, dep: Expr):
...
+            name,
+            Fused._execute_internal_graph,
+            # Wrap the actual subgraph as a data node such that the tasks are
+            # not erroneously parsed. The external task would otherwise carry
+            # the internal keys as dependencies which is satisfiable
+            DataNode(None, internal_tasks),
+            dependencies,
+            (self.exprs[0]._name, index),
+        )
+        return t
```
fjetter, line 3800: I implemented this (and similar patterns) now two or three times. I believe I'll factor this out to a classmethod like `Task.fuse_tasks`. **If** we end up rewriting the low level fusion this will obviously also come in handy. For now, we can keep this logic to get out a fix quickly

dask_expr/_expr.py
```diff
@@ -3766,19 +3766,44 @@ def _broadcast_dep(self, dep: Expr):
...
+            # also get the list of external dependencies when iterating over the
+            # subgraphs above.
+            for dep in external_deps
+        }
+        t = Task(
+            name,
+            Fused._execute_internal_graph,
+            # Wrap the actual subgraph as a data node such that the tasks are
+            # not erroneously parsed. The external task would otherwise carry
+            # the internal keys as dependencies which is satisfiable
```
hendrikmakait, line 3795: Do you mean

```suggestion
            # the internal keys as dependencies which is undesirable
```
? Otherwise I don't get it.

dask_expr/_expr.py
```diff
@@ -3766,19 +3766,44 @@ def _broadcast_dep(self, dep: Expr):
...
-            subgraphs[self._blockwise_arg(dep, index)] = "_" + str(i)
+            t = _expr._task(subname, subname[1])
+            assert t.key == subname
+            internal_tasks.append(t)
+            seen_keys.add(subname)
+            external_deps.update(t.dependencies)
+        external_deps -= seen_keys
+        dependencies = {
+            dep: TaskRef(dep)
+            # Note: the method dependencies isn't strictly needed here. We could
```
hendrikmakait, line 3785: What do you mean by "the method dependencies"?

hendrikmakait (approved):
Thanks, @fjetter. This looks generally good to me, two non-blocking questions about comments.

dask_expr/_expr.py
fjetter: *not*

PR #1162, opened 2024-11-11
Use Taskspec fuse implementation

fjetter:
This is using the `Task.fuse` (see https://github.com/dask/dask/pull/11509) method for the Fused task. In theory this should've reduced complexity quite a bit but I encountered problems since the subgraphs are including dead nodes, i.e. there are expressions in `_expr` that are no longer reachable if we walk the tasks. The `Task.fuse` raises in such a case since it requires a subgraph that is uniquely reducing to a single task (I put that condition in to catch stuff like this).

Adding an explicit culling step in advance fixes this so everything is OK. However, where is this redundant node coming from and can we get rid of this??

cc @phofl

### DIFF
--- a/dask_expr/_expr.py
+++ b/dask_expr/_expr.py
@@ -3776,11 +3776,15 @@ class Fused(Blockwise):
 
             assert t.key == subname
             internal_tasks[t.key] = t
+        # The above is ambiguous and we have to cull to get an unambiguous graph
         outkey = (self.exprs[0]._name, index)
         work = [outkey]
         internal_tasks_culled = []
         while work:
             tkey = work.pop()
+            if tkey not in internal_tasks:
+                # External dependency
+                continue
             t = internal_tasks[tkey]
             internal_tasks_culled.append(t)
             work.extend(t.dependencies)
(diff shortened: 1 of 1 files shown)


dask_expr/_expr.py
```diff
@@ -3766,32 +3766,29 @@ def _broadcast_dep(self, dep: Expr):
...
+        work = [outkey]
+        internal_tasks_culled = []
+        while work:
+            tkey = work.pop()
+            if tkey not in internal_tasks:
+                # External dependency
+                continue
+            t = internal_tasks[tkey]
+            internal_tasks_culled.append(t)
+            work.extend(t.dependencies)
```
fjetter, line 3790: All of this shouldn't be necessary if there weren't dead nodes

phofl: Do you have an example with dead nodes? I am curious to see what's going on there

phofl:
I'ver removed the culling, things should be good to go now

phofl (approved):

phofl:
thanks

PR #1166, opened 2024-11-14
Fix inifinite loop

andrii-uniq:
In case of absent `mode` paramater in the custom fsspec file system implementation removed rows caused infinite loop, as mentioned in comments several rows above

### DIFF
--- a/dask_expr/_core.py
+++ b/dask_expr/_core.py
@@ -507,8 +507,6 @@ class Expr:
             if key in _parameters:
                 idx = _parameters.index(key)
                 return self.operands[idx]
-            if is_dataframe_like(self._meta) and key in self._meta.columns:
-                return self[key]
 
             link = "https://github.com/dask-contrib/dask-expr/blob/main/README.md#api-coverage"
             raise AttributeError(


phofl (changes requested):
Please open an issue clearly describing the problem and then add a test for your changes here. The current change is definitely incorrect since it will break existing stuff

andrii-uniq:
```
    def __getattr__(self, key):
        try:
        except AttributeError as err:
            if key.startswith("_meta"):
                # Avoid a recursive loop if/when `self._meta*`
                # produces an `AttributeError`
                raise RuntimeError(
                    f"Failed to generate metadata for {self}. "
                    "This operation may not be supported by the current backend."
                )
           ....
            if is_dataframe_like(self._meta) and key in self._meta.columns:
                return self[key]
```
I removed the last two rows. I don't know full library guts, so I can not correctly understand the logic behind the intended behavior. But the code clearly shows contradiction. Also, I have a specific example where these rows caused an infinite loop, but this is a rather specific case. If it is suitable, I can add a test for this case.

phofl:
the current logic definitely behaves as intended for what it was added for, there is no contradiction.

Yes, please add an example that is failing / causing an infinite loop

andrii-uniq:
I see:
` # Avoid a recursive loop if/when self._meta*`
`if is_dataframe_like(self._meta)`
Okay, I'll prepare a minimal example and will add it.

PR #1168, opened 2024-11-15
Deprecated and remove from_legacy_dataframe usage

phofl:
goes with https://github.com/dask/dask/pull/11529

### DIFF
--- a/.github/workflows/dask_test.yaml
+++ b/.github/workflows/dask_test.yaml
@@ -47,7 +47,7 @@ jobs:
           cache-environment-key: environment-${{ steps.date.outputs.date }}-0
 
       - name: Install current main versions of dask
-        run: python -m pip install git+https://github.com/dask/dask
+        run: python -m pip install git+https://github.com/phofl/dask@legacy-dataframe
 
       - name: Install current main versions of distributed
         run: python -m pip install git+https://github.com/dask/distributed
--- a/.github/workflows/dask_test.yaml
+++ b/.github/workflows/dask_test.yaml
@@ -47,7 +47,7 @@ jobs:
           cache-environment-key: environment-${{ steps.date.outputs.date }}-0
 
       - name: Install current main versions of dask
-        run: python -m pip install git+https://github.com/phofl/dask@legacy-dataframe
+        run: python -m pip install git+https://github.com/dask/dask
 
       - name: Install current main versions of distributed
         run: python -m pip install git+https://github.com/dask/distributed
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -47,7 +47,7 @@ jobs:
           cache-environment-key: environment-${{ steps.date.outputs.date }}-1
 
       - name: Install current main versions of dask
-        run: python -m pip install git+https://github.com/phofl/dask@legacy-dataframe
+        run: python -m pip install git+https://github.com/dask/dask
         if: ${{ matrix.environment-file == 'ci/environment.yml' }}
 
       - name: Install current main versions of distributed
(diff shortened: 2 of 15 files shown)


.github/workflows/dask_test.yaml
```diff
@@ -47,7 +47,7 @@ jobs:
           cache-environment-key: environment-${{ steps.date.outputs.date }}-0
 
       - name: Install current main versions of dask
-        run: python -m pip install git+https://github.com/dask/dask
+        run: python -m pip install git+https://github.com/phofl/dask@legacy-dataframe
```
jrbourbeau, line 50: TODO revert before merging

```suggestion
        run: python -m pip install git+https://github.com/dask/dask
```

.github/workflows/test.yaml
```diff
@@ -47,7 +47,7 @@ jobs:
           cache-environment-key: environment-${{ steps.date.outputs.date }}-1
 
       - name: Install current main versions of dask
-        run: python -m pip install git+https://github.com/dask/dask
+        run: python -m pip install git+https://github.com/phofl/dask@legacy-dataframe
```
jrbourbeau, line 50: TODO revert before merging

```suggestion
        run: python -m pip install git+https://github.com/dask/dask
```

dask_expr/io/json.py
```diff
@@ -114,7 +113,6 @@ def read_json(
         path_converter=path_converter,
         **kwargs,
     )
-    return from_legacy_dataframe(df)
```
jrbourbeau, line 117: Nice. It looks like we can do this now since the changes over in https://github.com/dask/dask/pull/11529 make some methods (e.g. `dd.read_json`) more `dask-expr` aware

dask_expr/io/json.py
```diff
@@ -172,7 +170,7 @@ def to_json(
     from dask.dataframe.io.json import to_json
 
     return to_json(
-        df.to_legacy_dataframe(),
+        df,
```
jrbourbeau, line 173: Over in `orc.py` this is `df.optimize()` -- should we do something similar here? Or is the `.optimize()` maybe not needed over there?

dask_expr/_collection.py
```diff
@@ -1430,9 +1417,18 @@ def to_dask_array(
...
+
+        arr = self.values
+
+        chunks = self._validate_chunks(arr, lengths)
+        arr._chunks = chunks
+
+        if meta is not None:
+            arr._meta = meta
+
+        return arr
```
jrbourbeau, line 1431: Just noting that this is the same implementation as over in `dask/dask`

jrbourbeau (commented):
Thanks @phofl

dask_expr/tests/test_reductions.py
```diff
@@ -146,10 +146,10 @@ def test_dataframe_split_every(pdf, df, split_every, expect_tasks, reduction):
 
 
 @pytest.mark.parametrize(
-    "split_every,expect_tasks", [(False, 53), (None, 57), (5, 57), (2, 73)]
+    "split_every, expect_tasks", [(False, 53), (None, 57), (5, 57), (2, 73)]
 )
 def test_dataframe_mode_split_every(pdf, df, split_every, expect_tasks):
-    assert_eq(df.to_legacy_dataframe().mode(split_every=split_every), pdf.mode())
```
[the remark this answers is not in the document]
phofl: No clue, I was as confused as you are about this. Probably a mistake when adding the test at some point

dask_expr/io/json.py
phofl: Yep

dask_expr/io/json.py
phofl: This calls to_delayed, which triggers the optimization. I'd rather collect them in a single place

PR #1176, opened 2024-12-16
Fix assign optimization when overwriting columns

phofl:
closes https://github.com/dask/dask/issues/11596


hendrikmakait (approved):
Thanks, @phofl

PR #1177, opened 2024-12-16
Remove legacy conversion functions

phofl:
Removing the legacy implementation, so lets rip out the converter functions


hendrikmakait (approved):
Thanks, @phofl

